PR #2, opened 2023-11-08
Feature: Add configuration to add an initial pin which can be centered on

dopenguin:
## Summary

Coordinates to set an initial pin can now be configured with the possibility to be centered on.

* Uncomment `initial` in `clients/meldemichel/src/mapConfiguration.ts:115-118`
  * When starting the client, the map is centered on a pin set to Billstraße 82 in Hamburg
  * If `centerOn` is set to `false` / is not set, the map will still zoom to the pin. This is caused by the ReverseGeocoder having `'plugin/pins/transformedCoordinate'` set as `coordinateSource`
* If `movable` is set to `false`, the style of the cursor is set to `not-allowed`

\- / -


packages/clients/meldemichel/src/index.html
```diff
@@ -34,6 +36,20 @@
...
+              centerOn: true,
+              coordinates,
+            },
+            movable: movablePin,
+            style: {
+              fill: movablePin ? '#0392cf' : '#ff0019',
+            },
+          },
+        },
+        */
```
warm-coolguy, line 52: I figured [this](https://github.com/Dataport/polar/pull/2/commits/a3cc780e45b891bffd45cc20bef974aea97f2422) would be a good idea. Do you agree?

packages/plugins/Pins/src/store/index.ts
```diff
@@ -28,6 +29,21 @@ const getInitialState = (): PinsState => ({
...
+  originalEvent,
+}: MapBrowserEvent<PointerEvent>) {
+  if (!dragging && pinsLayer) {
+    pinsLayer
+      .getFeatures(map.getEventPixel(originalEvent))
+      .then(
+        (features) =>
+          (document.body.style.cursor = features.length ? 'not-allowed' : '')
+      )
+  }
```
warm-coolguy, line 44: This also changes the cursor when not directly hovering the pin, i.e. hovering its hole or lower left svg box corner. I understand that we're completely rendering in OL's canvas and therefore can't make use of the usual CSS approaches. However, in the [current implementation](https://static.hamburg.de/kartenclient/prod/), we do have highlighting on the features, and it does disappear in the lower left svg box corner; would you please look up that solution? It may be applicable here, too.

packages/plugins/Pins/README.md
```diff
@@ -14,22 +14,30 @@ The usage of `displayComponent` has no influence on the creation of Pins on the
...
+| coordinateSource | string?        | The pins plugin may react to changes in other plugins. This specifies the path to such store positions. The position must, when subscribed to, return a GeoJSON feature.                                                                                                                            |
+| initial          | number[]?      | Configuration options for setting an initial pin.                                                                                                                                                                                                                                                   |
+| movable          | boolean?       | Whether user may drag pin with mouse. Defaults to false.                                                                                                                                                                                                                                            |
```
warm-coolguy, line 24: While this is technically correct (and, therefore, the best kind of correct), we also need to implement that the cursor resists being moved by *clicking* another position. Currently, it can be moved by clicking.

warm-coolguy (changes requested):

packages/plugins/Pins/README.md
dopenguin: I'm thinking about changing `movable` to accept the strings `drag`, `click` and `none`, where

- `drag` accepts dragging and clicking
- `click` only accepts clicking
- `none` won't budge

when trying to move the pin. `movable` would still accept a boolean for now, but that configuration would become deprecated. Would that suffice?

packages/plugins/Pins/src/store/index.ts
dopenguin: Implemented in 5eee84cd

dopenguin:
🏓 @warm-coolguy

packages/plugins/Pins/README.md
warm-coolguy: Sounds great. 👍

warm-coolguy:
@dopenguin 🏓

warm-coolguy (changes requested):
Please update the files `single.immovable.html` and `single.movable.html` regarding this newly developed feature.

packages/plugins/Pins/README.md
dopenguin: Done f7f5fa6f

dopenguin:
Regarding: https://github.com/Dataport/polar/pull/2#pullrequestreview-1724121969:

What did you have in mind?

warm-coolguy (approved):
https://github.com/Dataport/polar/pull/2/commits/15ae31d02f704e99134a1ebdc8b53807c39ef820 is what I had in mind. Also added https://github.com/Dataport/polar/pull/2/commits/76661944c5d423b00a7783e25d3b702e15bd01c5 since we had lost the grab/grabbing cursors on draggable pins.

Approve from my end. You may want to check my changes before merging.

packages/plugins/Pins/src/store/index.ts
```diff
@@ -85,27 +105,57 @@ const storeModule: PolarModule<PinsState, PinsState> = {
...
         )
       }
+      if (!movable || movable === 'none') {
+        rootGetters.map.addInteraction(move)
+        move.on(
+          'select',
+          ({ selected }) =>
+            (document.body.style.cursor = selected.length
+              ? 'not-allowed'
+              : 'inherit')
```
dopenguin, line 115:
```suggestion
              : '')
```
This change from 76661944c5d423b00a7783e25d3b702e15bd01c5 has no influence on the behaviour if one wants the pin movable, as this interaction is only added to the map if the pin should not be usable.  

Looking at `main`, the issue was already present beforehand and may have something to do with the `Translate` interaction in `makeMarkerDraggable` not being updated properly. Would an issue be the way for handling this?

dopenguin (commented):
🏓 @warm-coolguy

packages/plugins/Pins/src/store/index.ts
warm-coolguy: https://github.com/Dataport/polar/issues/15

PR #4, opened 2023-11-09
Remove buildAttributions and add core feature to setAttributions

dopenguin:
## Summary

Removed redundantly complex code in favour of a simple approach of what `setAttributions` is supposed to do: Just get the `layerAttributions` from the configuration.

## Instructions for local reproduction and review#

Test any client whether they still show the correct amount of attributions.  
In contrast to the current `main`-version, an empty line is now hidden in `meldemichel`.

\- / -


warm-coolguy (approved):
I have extended the AfM client example configuration to showcase a situation where this doesn't quite work; when zooming out, the attribution is still displayed, despite the layer not being shown.

However, bevore finishing my review, I realised it's the same on `main`. Maybe it's supposed to work this way? 🤔 It's not an issue of this MR anyway, so here's the Approval.

PR #6, opened 2023-11-09
Add workflow generating a sbom on push to main

dopenguin:
## Summary

Add workflow generating a sbom on push to main. Removed old @cyclonedx-package-based variant.

## Instructions for local reproduction and review

Gotta merge to see the workflow in action.

## Relevant tickets, issues, et cetera

\- / -

PR #7, opened 2023-11-09
Add staticAttributions linking to the copyright information

dopenguin:
## Summary

Add staticAttributions linking to the copyright information and sbom.

- `npm run meldemichel:dev` and `npm run meldemichel:build && npm run meldemichel:build:example`
  - [x] Complete: Layer Attributions and Copyright Information of Hamburg City are shown
  - [x] Report and Single: Only Layer Attributions are shown

#6

### DIFF
--- a/packages/clients/meldemichel/src/mapConfigurations.ts
+++ b/packages/clients/meldemichel/src/mapConfigurations.ts
@@ -74,6 +74,10 @@ const commonLayers: LayerConfiguration[] = [
   },
 ]
 
+// TODO: Update link with reference to sbom
+const sbomAttributionAnchor =
+  '<a href="https://github.com/Dataport/polar/network/dependencies" target="_blank">Software Bill of Materials</a>'
+
 const commonAttributions: Partial<AttributionsConfiguration> = {
   initiallyOpen: false,
   layerAttributions: [
@@ -152,6 +156,16 @@ const mapConfigurations = {
             title: 'meldemichel.attributions.reports',
           },
         ],
+        staticAttributions: [
+          `<ul style="display: flex; column-gap: 8px; list-style-type: none; padding: 0">
+            <li>
+              <a href="https://www.hamburg.de/impressum/" target="_blank">Impressum</a>
+            </li>
+            <li>
+              ${sbomAttributionAnchor}
+            </li>
+          </ul>`,
+        ],
       },
       geoLocation,
       gfi: {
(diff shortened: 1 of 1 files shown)


packages/clients/meldemichel/src/mapConfigurations.ts
```diff
@@ -74,6 +74,10 @@ const commonLayers: LayerConfiguration[] = [
   },
 ]
 
+// TODO: Update link with reference to sbom
+const sbomAttributionAnchor =
+  '<a href="https://github.com/Dataport/polar/network/dependencies" target="_blank">Software Bill of Materials</a>'
```
warm-coolguy, line 79: This is insufficient due to linking to the current state, not the software state. We may very well have live versions including packages that are no longer used in later (unpublished) states; the live environment's dependencies would be effectively missing.

I'd propose we make a copy of the GitHub SBOM on each production build that's included in the NPM package and then link this neighbouring file. But I'm not sure if we're out of luck here: Is this GitHub Dependency graph up to date *right after commiting* and *before the actions run*? If not, this may require using the old way to generate the SBOM again.

This also requires generating the SBOM *before* publishing and adding it to the build in question, so I assume the workflow files have to be merged to get these actions in order. However, I suppose that at the same time we can stop having the SBOM file in the project's root (or anywhere in the project) altogether. After all, the project itself does not contain its dependencies directly; only the builds do. This'd allow us to skip commiting the generated SBOM.

warm-coolguy (changes requested):

packages/clients/meldemichel/src/mapConfigurations.ts
dopenguin: My guess would be that we would have to generate `package-lock.json`-files for every client that we are publishing to be able to properly document the dependencies.

However, I'm not quite sure whether this would suffice as we may have to link to the devDependencies included in the root `package.json`.

Thinking about this problem made be realize though, that other Web-GIS do not link to the used dependencies as part of an SBOM; e.g. [Geo-Online HH](https://geoportal-hamburg.de/geo-online/) or [OpenStreetMap](https://www.openstreetmap.org/).  
Is it actually required?

warm-coolguy: From my understanding it is required as part of the [EU Cyber Resilience Act](https://eur-lex.europa.eu/legal-content/EN/TXT/HTML/?uri=CELEX:52022PC0454), which is currently a proposal. Upon entry into force, [we have 12 months to adapt](https://digital-strategy.ec.europa.eu/en/policies/cyber-resilience-act).

The relevant information regarding this is in the first document, ANNEX I 2(a).

Lockfiles seem to match these criteria. 👍 However, it seems we may forward this task to our future selves; it may e.g. resolve itself upon the planned changes to the build structure.

dopenguin (commented):
🏓 @warm-coolguy

packages/clients/meldemichel/src/mapConfigurations.ts
dopenguin: Adding SBOM has been postponed for now

warm-coolguy (approved):

PR #8, opened 2023-11-10
Prepare Meldemichel Alpha

warm-coolguy:
## Summary

- Adding Meldemichel documentation to GitHub Page
- Adding Meldemichel test clients to GitHub Page
- Implementing AfM data interface
  - Done by also extending core state (now includes map center)

- Run the `buildPages` script and check the results
  - Is the documentation comprehensible/complete?
  - Do the clients for alpha (REPORT/SINGLE) work appropriately?
- REPORT mode can be used to check the data interface.

## TODO

- [x] Add documentation regarding PINS if https://github.com/Dataport/polar/pull/2 is merged first


warm-coolguy:
@dopenguin 🏓 TODO resolved.

packages/clients/meldemichel/API.md
```diff
@@ -1,20 +1,24 @@
...
-| reportServiceId | string?                              | COMPLETE mode only. ID of the report layer to display.                                                                                                                                                                                                                                  |
-| configOverride  | object?                              | You will probably not need this object. It is included to override any internal configuration, should the need for quick reconfiguration arise. Please refer to the general documentation on what can be done here. If you need it, please communicate back with us what the issue was. |
+| mode            | enum["REPORT", "SINGLE", "COMPLETE"] | See chapters below for an overview of the modes.                                                                                                                                                                                                                                        |
```
dopenguin, line 18: Why the reorder of the `enum`?

packages/clients/meldemichel/API.md
```diff
@@ -72,101 +54,91 @@ A document rendering the map client could e.g. look like this:
...
+## Rendering COMPLETE mode
 
-```js
-import meldemichelMapClient from '@polara/client-meldemichel'
+The `index.html` included in the package's `dist` folder has been prepared for this mode and must merely be hosted.
 
-const mapInstance = await meldemichelMapClient.createMap({
-  // ... see `createMap` description
-})
+// TODO update this block after implementation (TO BE IMPLEMENTED)
```
dopenguin, line 126: Shouldn't the paragraphs below already translated now?

packages/types/custom/core.ts
```diff
@@ -531,11 +531,13 @@ export interface PolarError {
 
 export interface CoreState {
   map: number
+  center: [number, number] | null
   clientHeight: number
   clientWidth: number
   components: number
   configuration: MapConfig
   errors: PolarError[]
+  plugin: object
```
dopenguin, line 540: I can't seem to find where this resides. I can find `modules.plugin`, but that is not what is typed here

packages/clients/meldemichel/src/store/module.ts
```diff
@@ -4,12 +4,21 @@
...
-  mapBaseLayer: number | string
-  vendor_maps_position: string // resembling 'number,number'
-  vendor_maps_address_str: string
-  vendor_maps_address_hnr: number | string
+  mapCenter?: string // resembling 'number,number'
+  mapZoomLevel?: number | string
+  mapBaseLayer?: number | string
+  vendor_maps_position?: string // resembling 'number,number'
+  vendor_maps_address_str?: string
+  vendor_maps_address_hnr?: number | string
```
dopenguin, line 12: The types don't correctly represent the actual values. All the values are always present in the object, but may return `undefined`, so an intersection with `undefined` would be the way to go

packages/clients/meldemichel/src/mapConfigurations.ts
```diff
@@ -16,8 +16,6 @@ const stadtplan = '453'
 const luftbilder = '452'
 export const hamburgBorder = '6074' // boundary layer for pins / geolocalization
 
-// TODO decide whether to use static hhborder.geojson (TODO update file - is there a difference since last time?)
```
dopenguin, line 19: The `hhborder.geojson` is still present and should be deleted, if the todo is resolved

packages/clients/meldemichel/example/report.html
```diff
@@ -53,12 +53,12 @@
...
-      <p>Coordinates: <span id="test-display-coordinate">null</span></p>
-      <p>Address: <span id="test-display-address">null</span></p>
-      <p>Address data: <span id="test-display-address-data">null</span></p>
+      <h2>Subscribed data</h2>
+      <ul id="subscribed-data-list"></ul>
+      <h2>Setting data</h2>
+      <p>This uses test data. See report.html contents.</p>
       <button onclick="_polar.setMapState()">
-        Set parameters after rendering
+        Set test parameters after rendering (only works initially)
```
dopenguin, line 61: Clicking the button doesn't center on the respective coordinate which is irritating.  
Was that the case before as well?

dopenguin (changes requested):
🏓 @warm-coolguy

packages/clients/meldemichel/API.md
warm-coolguy: Seemed appropriate to me to list the options relevant to users first. "COMPLETE" will probably only be used by us.

packages/clients/meldemichel/API.md
warm-coolguy: No, this is the "COMPLETE" region of the document. We're not done with the implementation here.

packages/types/custom/core.ts
warm-coolguy: It's e.g. used in `packages/clients/meldemichel/src/store/module.ts`:
`let address = rootState?.plugin?.addressSearch?.chosenAddress`

packages/clients/meldemichel/src/store/module.ts
warm-coolguy: The actual object referred to in `SetMapStatePayload` is not defined by us, but by the people installing our client. They may very well leave out any of the information provided.

The extending `GetMapState` indeed does always have all fields; however, directly typing `undefined` as any key's value further below (as done for `vendor_maps_address_hnr`) does not yield an error. Hence, TS accepts that `?` also allows for setting the value as `undefined`.

packages/clients/meldemichel/src/mapConfigurations.ts
warm-coolguy: The unused file has been deleted. The lib function `passesBoundaryCheck` takes care of the underlying problem.

packages/clients/meldemichel/example/report.html
warm-coolguy: It centers on the `mapCenter`, not the `vendor_maps_position`, which is the pin coordinate. This is intended: This function aims to recreate a previous user view.

warm-coolguy:
🏓 @dopenguin

packages/types/custom/core.ts
dopenguin: Weird, that I can't seem to find the place in the core 🎩

packages/clients/meldemichel/src/store/module.ts
dopenguin: I'd still press for typing the values differently but won't open a barrel if you won't 🎩

dopenguin (approved):

PR #9, opened 2023-11-13
fix icon menu content placement

warm-coolguy:
## Summary

- Replaced hack with absolute positioning.
- Fixed bug that appeared when Zoom was sorted prior to Attributions, but decided to not render itself due to tiny client window.
- Remove spacing on IconMenu's ul that seems to have leaked in unnoticed.

Run arbitrary clients in arbitrary environments to check if something is broken. Most notable change should occur in Meldemichel example clients for `'SINGLE'` mode, where the attributions were previously displayed above the map.

### DIFF
--- a/packages/plugins/IconMenu/src/components/IconMenu.vue
+++ b/packages/plugins/IconMenu/src/components/IconMenu.vue
@@ -45,7 +45,7 @@
               : 'icon-menu-list-item-content',
             'icon-menu-list-item-content-scrollable-y',
           ]"
-          :style="getContentStyle(Number(index))"
+          :style="`max-height: ${maxHeight};`"
         />
       </template>
     </component>
@@ -84,20 +84,6 @@ export default Vue.extend({
   },
   methods: {
     ...mapMutations('plugin/iconMenu', ['setOpen']),
-    getContentStyle(index: number) {
-      // HACK: Zoom is currently the only plugin that adds two buttons instead of one.
-      const zoomIndex = this.menus.findIndex(({ id }) => id === 'zoom')
-      const isAfterZoom = zoomIndex < index
-      return `max-height: ${this.maxHeight}; ${
-        this.isHorizontal
-          ? `right: calc(${
-              this.menus.length - (isAfterZoom ? index + 1 : index + 2)
-            } * (-100% - 0.5rem));`
-          : `top: calc(${
-              isAfterZoom && zoomIndex !== -1 ? index + 1 : index
-            } * (-100% - 0.5rem));`
-      }`
-    },
     toggle(index) {
       const { open } = this
       if (open === index) {
@@ -112,17 +98,17 @@ export default Vue.extend({
 
 <style lang="scss" scoped>
 .icon-menu-list {
+  position: relative;
   list-style: none;
+  padding: 0;
 }
 
 .icon-menu-list-item-horizontal {
-  position: relative;
   float: left;
   margin-left: 0.5rem;
 }
 
 .icon-menu-list-item {
-  position: relative;
   margin-bottom: 0.5rem;
   z-index: 1;
 }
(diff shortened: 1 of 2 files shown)


packages/plugins/IconMenu/src/components/IconMenu.vue
```diff
@@ -112,31 +98,33 @@ export default Vue.extend({
...
+  top: 0;
+  right: calc(100% + 0.5em);
 }
 
 .icon-menu-list-item-content-horizontal {
   position: absolute;
   white-space: nowrap;
-  top: 3em;
+  top: calc(100% + 0.5em);
+  right: -0.5em;
```
dopenguin, line 127:
```suggestion
  right: 0;
```
Otherwise it hugs the side of the display

dopenguin (changes requested):
Content like the one from the attribution plugin is still overflowing outside of the client.  
Should this be handled as a separate task or would you like to tackle this here?

🏓@warm-coolguy

packages/plugins/IconMenu/src/components/IconMenu.vue
warm-coolguy: I figured it was more important to have the box align to the buttons than to have the sidebar not hugging the side of the viewport.

spacing
spacing-with-sidebar

warm-coolguy:
> Content like the one from the attribution plugin is still overflowing outside of client. Should this be handled as a separate task or would you like to tackle this here?
> 
> 🏓@warm-coolguy

I think it's acceptable for IconMenu contents to overflow the client. After all, such windows can be closed (and hence don't prevent using other overlapped contents), and especially on narrow use cases (i.e. the sidebar scenario) there's hardly enough space inside the client for comfortable/usable menus.

However, I agree that the contents should always be reachable. On writing an issue, I failed to create a scenario where the contents left the reachable area. Are you able to create such a scenario? If so, please do make an issue of it.

🏓 @dopenguin

packages/plugins/IconMenu/src/components/IconMenu.vue
dopenguin: When testing with `right: 0;` on a phone with a viewport of `412px x 919px`, I received the following results in Chrome:

att_0
layerchooser_0

With `right: -0.5em;` the results were the following:

att_-0 5em
layerchooser_-0 5em

Meaning, that the alignment with the Buttons was only present when using `right: 0;`.

Could you please elaborate in which use cases the above situations would occur?

dopenguin:
> > Content like the one from the attribution plugin is still overflowing outside of client. Should this be handled as a separate task or would you like to tackle this here?
> > 🏓@warm-coolguy
> 
> I think it's acceptable for IconMenu contents to overflow the client. After all, such windows can be closed (and hence don't prevent using other overlapped contents), and especially on narrow use cases (i.e. the sidebar scenario) there's hardly enough space inside the client for comfortable/usable menus.
> 
> However, I agree that the contents should always be reachable. On writing an issue, I failed to create a scenario where the contents left the reachable area. Are you able to create such a scenario? If so, please do make an issue of it.
> 
> 🏓 @dopenguin

Tackled the issue in #17

packages/plugins/IconMenu/src/components/IconMenu.vue
warm-coolguy: `right: 0;` visibly doesn't produce alignment between the buttons and the contents in your screenshots.

My screenshots have been provided from desktop browsers. I see, the difference is that `scrollbar-gutter` produces the right distance in desktop browsers, but not on mobile devices, where scrollbars are usually these thin lines atop the inside of boxes.

For the time being, I suggest we leave it at that. It looks good on desktop and acceptable on mobile devices. Let's say it's intentional to make better usage of the available space. With `right: 0;`, both are slightly off.

dopenguin: We'll probably get back to this, but fine by me for the time being

dopenguin (approved):
🚀 @warm-coolguy

PR #10, opened 2023-11-13
deduplicate bound layer types

warm-coolguy:
## Summary

A common type for plugins bound by layer boundaries has been introduced.

Running pipeline is sufficient, no direct effects on code.

### DIFF
--- a/packages/lib/passesBoundaryCheck/index.ts
+++ b/packages/lib/passesBoundaryCheck/index.ts
@@ -2,6 +2,7 @@ import { Map } from 'ol'
 import { Coordinate } from 'ol/coordinate'
 import VectorLayer from 'ol/layer/Vector'
 import VectorSource from 'ol/source/Vector'
+import { LayerBoundPluginOptions } from '@polar/lib-custom-types'
 
 // arbitrarily give up after 10s of stalling
 let readinessChecks = 0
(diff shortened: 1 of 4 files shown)


packages/types/custom/core.ts
```diff
@@ -228,7 +228,21 @@ export interface GfiLayerConfiguration {
   format?: string
 }
 
-export interface GeoLocationConfiguration extends PluginOptions {
+export interface LayerBoundPluginOptions extends PluginOptions {
```
dopenguin, line 231: A changelog entry for these changes is missing

packages/lib/passesBoundaryCheck/index.ts
```diff
@@ -33,7 +34,7 @@ const isReady = async (source: VectorSource): Promise<boolean> => {
  */
 export const passesBoundaryCheck = async (
   map: Map,
-  boundaryLayerId: string | undefined,
+  boundaryLayerId: LayerBoundPluginOptions['boundaryLayerId'],
```
dopenguin, line 37: Depending on what is defined in `@polar/lib-custom-types` it may change the behaviour in the future. Would that reason justify a changelog entry in this package or would that be just overkill?

dopenguin (changes requested):

dopenguin:
🏓 @warm-coolguy

packages/types/custom/core.ts
warm-coolguy: https://github.com/Dataport/polar/pull/10/commits/ce793407c2aa1a7c9550c59919c7128127d3442b

packages/lib/passesBoundaryCheck/index.ts
warm-coolguy: For now it's the same, and in the future changes are supposed to trickle through. I don't think this is a change worth announcing, but an internal detail.

warm-coolguy:
> 🏓 @warm-coolguy

🏓 @dopenguin

dopenguin (approved):
After the conflicts have been resolved, this can be merged

PR #11, opened 2023-11-13
refactor pins module code for readability

warm-coolguy:
## Summary

Implement refactoring for pins module to deduplicate code and make it easier to read. Suggestion from Pascal, hence commited in his authorship.

Test whether behaviour stayed the same for clicking/dragging outside of Hamburg boundaries in Meldemichel dev client.


dopenguin (approved):
g2g 🏓 @warm-coolguy

PR #13, opened 2023-11-14
Feature: Extend addressSearch accessibility

dopenguin:
## Summary

Various accessbility changes have been added to improve the overall usability of the addressSearch.

The meldemichel complete client can be run as an example.
- Information is read on input to a user using a screen reader regarding the minimum amount of characters required
- The result list can now only be navigated via arrow keys
- The first result is focused after a successful search

\- / -


packages/plugins/AddressSearch/src/components/Results.vue
```diff
@@ -117,6 +154,53 @@ export default Vue.extend({
...
+        nextIndex = index - 1
+        usedInnerDex = this.featureListsWithCategory[nextIndex].length - 1
+      } else if (nextInnerDex === featureListLength) {
+        nextIndex =
+          index + 1 === this.featureListsWithCategory.length ? 0 : index + 1
+        usedInnerDex = 0
+      } else {
+        nextIndex = index
+        usedInnerDex = nextInnerDex
+      }
```
warm-coolguy, line 178: This doesn't always work in the DISH client. Disable CORS in your browser to test the client appropriately.

Test cases:

- Enter "Rathaus" into the search. Only the first search result category is reachable by the arrow keys.
- Enter "Rathausm" into the search. The second category can be reached, but there is no way back up with the arrow keys.

In both scenarios, the open/close buttons for further results are not reachable by the arrow keys, and neither are the results reachable from the open/close buttons with arrow keys. Especially the latter case seems unintuitive since we're still "in" the thing we're navigating.

packages/plugins/AddressSearch/src/components/Results.vue
```diff
@@ -100,6 +125,18 @@ export default Vue.extend({
...
+        const firstElement =
+          // @ts-expect-error | Type conversion is fine here as the querySelector method is monkeyPatched in core/createMap
+          (document.querySelector('[data-app]') as ShadowRoot).getElementById(
+            'polar-plugin-address-search-results-feature-0-0'
+          )
+        if (firstElement) {
+          firstElement.focus()
+        }
+      })
+    },
```
warm-coolguy, line 139: Due to the search running automagically on entering x characters, this may be a very annoying feature to many users with a lower typing speed or fast services. Realized this on the DISH client, where the first search is for keywords stored in the client and therefore almost instantaneous – this doesn't allow my to finish writing.

packages/plugins/AddressSearch/src/components/Results.vue
```diff
@@ -26,14 +28,37 @@
...
+        tag="li"
+        tabindex="-1"
         :class="{
           'polar-plugin-address-search-hidden-result':
             innerDex >=
             (openCategories.includes(category)
               ? Number.MAX_SAFE_INTEGER
               : limitResults),
         }"
+        @keydown.down.prevent.stop="
```
warm-coolguy, line 46: This stops working when NVDA is turned on.

packages/plugins/AddressSearch/src/components/Results.vue
```diff
@@ -100,6 +125,18 @@ export default Vue.extend({
...
+    featuresAvailable(): void {
+      this.$nextTick(() => {
+        const firstElement =
+          // @ts-expect-error | Type conversion is fine here as the querySelector method is monkeyPatched in core/createMap
+          (document.querySelector('[data-app]') as ShadowRoot).getElementById(
+            'polar-plugin-address-search-results-feature-0-0'
+          )
+        if (firstElement) {
+          firstElement.focus()
+        }
```
warm-coolguy, line 137: Duplicate implementation to `focusFirstResult`.

warm-coolguy (changes requested):
This feature is missing a changelog entry.

🏓 @dopenguin

packages/plugins/AddressSearch/src/components/Results.vue
```diff
@@ -5,12 +5,14 @@
...
     class="mx-auto overflow-y-auto rounded-b-xl polar-plugin-address-search-toolbar-results"
     :max-height="maxHeight"
+    tabindex="-1"
   >
     <v-list
       v-for="(
         { features, category, categoryId }, index
       ) in featureListsWithCategory"
       :key="['results-list', index].join('-')"
+      tag="ul"
```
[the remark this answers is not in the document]
dopenguin: Removed b0a168de

packages/plugins/AddressSearch/src/components/Results.vue
dopenguin: Could this be changed to be a configurable feature?  
IIRC it was a requirement given for Meldemichel

packages/plugins/AddressSearch/src/components/Results.vue
dopenguin: Deduplicated 2e873110

packages/plugins/AddressSearch/src/components/Results.vue
dopenguin: The problems regarding the results occur because the full list is being rendered (the list items are in the DOM) meaning that the open / close buttons don't properly hide them.  
If all the results are being shown for the full list, one can navigate between the two lists.

However, I fully understand that the behaviour is weird regarding the open / close buttons so I'll take a look into them being also only navigable via arrow keys.

dopenguin: This should now work as intended in both Meldemichel and DISH.

If you have any suggestions to clean up the way the id of the next element is retrieved, I am open to 🏌️⛳

packages/plugins/AddressSearch/src/components/Results.vue
dopenguin: Can now be configured and defaults to `false` c0338281

packages/plugins/AddressSearch/src/components/Results.vue
dopenguin: It still works, the visual focus is lost though.

While looking into this problem, I didn't manage to find a solution where the visible arrow-key-focus exists while using NVDA.  
Do you have a suggestion in which direction a solution may lie or do you think that this is negligible as a ScreenReader user most likely will focus (HA) on what the ScreenReader is reading?

dopenguin:
🏓 @warm-coolguy

packages/plugins/AddressSearch/src/components/Results.vue
warm-coolguy: I think both visual and speech information are required. However, I now consider this resolved:

After some research I figured out it's a skill issue on our end. ScreenReader users seem to be used to change the arrow key mode by pressing "Insert+Space". I tested that and it then works as intended.

🎩

packages/plugins/AddressSearch/src/components/Results.vue
warm-coolguy: - There a small issue left in DISH where arrow navigation doesn't work for singular result in later categories. E.g. search for "2-20/Rathausstraße". The arrow key won't properly work in this situation.
- On pressing space on a selected list item, the list item shows an effect, but nothing happens. It was like that before, but would you include a quick fix while we're here?
- "If you have any suggestions to clean up the way the id of the next element is retrieved, I am open to 🏌️⛳" – I assume it would be shorter to write functions `getNextElement` and `getPreviousElement` that get the current element's ID and just navigate the DOM with `.nextSibling` and `.previousSibling`. Since we have one level of nesting (`<ul>`) to consider, we'd also need `parentElement` and `children` for switching lists and some skipping on at least `div` and `hr`. 🤔

warm-coolguy:
> 🏓 @warm-coolguy

🏓 @dopenguin

packages/plugins/AddressSearch/src/components/Results.vue
dopenguin: - I'll take a look into that
- If it is a rather easy fix, sure. Otherwise I will write an issue
- That doesn't really seem easier tbh :D We can ☎️ later

warm-coolguy: - Confirmed.
- It still ripples on my machine.
- Created. https://github.com/Dataport/polar/issues/26

warm-coolguy:
> 🏓 @warm-coolguy

🏓 @dopenguin

packages/plugins/AddressSearch/src/components/Results.vue
dopenguin: The ripple should be dealt with in all situations fbfb8eba

warm-coolguy (approved):
🚀 @dopenguin

PR #20, opened 2023-11-15
Fix: Type access to core configuration

dopenguin:
## Summary

The parameter `configuration` of the `core` store is always defined so optional chaining is not necessary.

If the pipeline succeeds, we are good to go

\- / -


warm-coolguy (approved):
:rocket: @dopenguin

PR #23, opened 2023-11-17
Feature/meldemichel afm button

warm-coolguy:
## Summary

Add AfmButton that allows users to switch from map-only view to AfM reporting process.

Check in COMPLETE mode. Toasts when no address is chosen, and opens AfM reporting process when address is chosen. The AfM reporting process shows the address selected in the map, and going "Back" in the process produces the previous map view (regarding its center, size of client is different).

### DIFF
--- a/packages/clients/meldemichel/src/mapConfigurations.ts
+++ b/packages/clients/meldemichel/src/mapConfigurations.ts
@@ -124,7 +124,7 @@ const geoLocation: Partial<GeoLocationConfiguration> = {
 }
 
 const mapConfigurations: Record<
-  string,
+  keyof typeof MODE,,
   (reportServiceId: string, afmUrl: string) => object
 > = {
   [MODE.COMPLETE]: (reportServiceId: string, afmUrl: string) => {
--- a/packages/clients/meldemichel/src/mapConfigurations.ts
+++ b/packages/clients/meldemichel/src/mapConfigurations.ts
@@ -124,7 +124,7 @@ const geoLocation: Partial<GeoLocationConfiguration> = {
 }
 
 const mapConfigurations: Record<
-  keyof typeof MODE,,
+  keyof typeof MODE,
   (reportServiceId: string, afmUrl: string) => object
 > = {
   [MODE.COMPLETE]: (reportServiceId: string, afmUrl: string) => {
@@ -171,9 +171,11 @@ const mapConfigurations: Record<
       },
       pins: commonPins,
       reverseGeocoder,
-      meldemichelAfmButton: {
-        displayComponent: true,
-        afmUrl,
+      meldemichel: {
+        afmButton: {
+          displayComponent: true,
+          afmUrl,
+        },
       },
     }
   },
--- a/packages/clients/meldemichel/src/mapConfigurations.ts
+++ b/packages/clients/meldemichel/src/mapConfigurations.ts
@@ -172,10 +172,7 @@ const mapConfigurations: Record<
       pins: commonPins,
       reverseGeocoder,
       meldemichel: {
-        afmButton: {
-          displayComponent: true,
-          afmUrl,
-        },
+        afmButton: { afmUrl },
       },
     }
   },
--- a/packages/clients/meldemichel/src/mapConfigurations.ts
+++ b/packages/clients/meldemichel/src/mapConfigurations.ts
@@ -212,7 +212,7 @@ export const getMapConfiguration = ({
     )
   }
   return {
-    // @ts-expect-error | reportServiceId might be undefined, but that's catched above for relevant cases
+    // @ts-expect-error | reportServiceId might be undefined, but that's caught above for relevant cases
     ...mapConfigurations[mode](reportServiceId, afmUrl),
   }
 }
--- a/packages/clients/meldemichel/src/mapConfigurations.ts
+++ b/packages/clients/meldemichel/src/mapConfigurations.ts
@@ -123,8 +123,11 @@ const geoLocation: Partial<GeoLocationConfiguration> = {
   showTooltip: true,
 }
 
-const mapConfigurations = {
-  [MODE.COMPLETE]: (reportServiceId: string) => {
+const mapConfigurations: Record<
+  string,
+  (reportServiceId: string, afmUrl: string) => object
+> = {
+  [MODE.COMPLETE]: (reportServiceId: string, afmUrl: string) => {
     return {
       ...commonMapConfiguration,
       addressSearch,
@@ -161,15 +164,17 @@ const mapConfigurations = {
             // TODO doesn't work atm; no coordinate source
             geometry: false,
             window: true,
-            properties: {
-              filename: 'Name of file',
-            },
+            properties: { filename: 'Name of file' },
           },
         },
         coordinateSources: [], // to be set in addPlugins.ts
       },
       pins: commonPins,
       reverseGeocoder,
+      meldemichelAfmButton: {
+        displayComponent: true,
+        afmUrl,
+      },
     }
   },
   [MODE.REPORT]: () => ({
@@ -195,8 +200,6 @@ const mapConfigurations = {
 
 export const getMapConfiguration = ({
   mode,
-  // NOTE temporarily commenting out, needed later in NeuesAnliegen-Plugin
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   afmUrl,
   reportServiceId,
 }: Pick<
@@ -208,13 +211,8 @@ export const getMapConfiguration = ({
       'POLAR Meldemichel Client: Missing reportServiceId configuration in mode COMPLETE.'
     )
   }
-  // @ts-expect-error | known issue; type definitions fit after merge in core
   return {
-    // @ts-expect-error | methods don't have a common interface
-    ...mapConfigurations[mode](reportServiceId),
-    // NOTE local plugin does not exist yet
-    // meldemichelNeuesAnliegen: {
-    //   afmUrl
-    // }
+    // @ts-expect-error | reportServiceId might be undefined, but that's catched above for relevant cases
+    ...mapConfigurations[mode](reportServiceId, afmUrl),
   }
 }
--- a/packages/clients/meldemichel/src/plugins/AfmButton/AfmButton.vue
+++ b/packages/clients/meldemichel/src/plugins/AfmButton/AfmButton.vue
@@ -10,7 +10,7 @@
     @keydown.space="click"
   >
     <v-icon small>fa-map-location</v-icon>
-    {{ $t('common:plugins.meldemichelAfmButton.buttonText') }}
+    {{ $t('common:plugins.meldemichel.afmButton.buttonText') }}
   </v-btn>
 </template>
 
@@ -29,10 +29,10 @@ export default Vue.extend({
       )
     },
     afmUrl() {
-      const url = this.configuration.meldemichelAfmButton.afmUrl
+      const url = this.configuration.meldemichel.afmButton.afmUrl
       if (!url) {
         console.error(
-          'Missing afmUrl parameter in `configuration.meldemichelAfmButton.afmUrl`.'
+          'Missing afmUrl parameter in `configuration.meldemichel.afmButton.afmUrl`.'
         )
         return ''
       }
@@ -46,7 +46,7 @@ export default Vue.extend({
       if (!this.mapStateReady) {
         this.$store.dispatch('plugin/toast/addToast', {
           type: 'info',
-          text: 'plugins.meldemichelAfmButton.missingAddress',
+          text: 'plugins.meldemichel.afmButton.missingAddress',
           timeout: 10000,
         })
       }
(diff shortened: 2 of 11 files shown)


packages/clients/meldemichel/src/mapConfigurations.ts
```diff
@@ -123,8 +123,11 @@ const geoLocation: Partial<GeoLocationConfiguration> = {
   showTooltip: true,
 }
 
-const mapConfigurations = {
-  [MODE.COMPLETE]: (reportServiceId: string) => {
+const mapConfigurations: Record<
+  string,
```
dopenguin, line 127:
```suggestion
  keyof typeof MODE,,
```
Also, `[MODE.COMPLETE]` can be simpliefied to `COMPLETE`

packages/clients/meldemichel/src/mapConfigurations.ts
```diff
@@ -123,8 +123,11 @@ const geoLocation: Partial<GeoLocationConfiguration> = {
...
+  (reportServiceId: string, afmUrl: string) => object
```
dopenguin, line 128: Correct would be to return `MapConfig` here but that wouldn't immediately work because of the overlap of configuring a plugin during `addPlugins` or during `createMap` via this config.

An idea would be to have all the completely required parameters, meaning a plugin would not work if not present, to be present during initialization in `addPlugins` and make all configurations in `MapConfig` a `Partial`, so that parameters be be overwritten.

If you want, you can add that change here. Otherwise, feel free to create an issue and I will tackle that.

packages/clients/meldemichel/src/mapConfigurations.ts
```diff
@@ -161,15 +164,17 @@ const mapConfigurations = {
...
-              filename: 'Name of file',
-            },
+            properties: { filename: 'Name of file' },
           },
         },
         coordinateSources: [], // to be set in addPlugins.ts
       },
       pins: commonPins,
       reverseGeocoder,
+      meldemichelAfmButton: {
```
dopenguin, line 174: Why not just `afmButton`?

Otherwise I'd suggest adding an overall directive `meldemichel` in which `afmButton` is a parameter - this is probably not the last local plugin we are writing.

packages/clients/meldemichel/src/mapConfigurations.ts
```diff
@@ -161,15 +164,17 @@ const mapConfigurations = {
...
+        displayComponent: true,
+        afmUrl,
```
dopenguin, line 176: Wouldn't just `url` suffice?

packages/clients/meldemichel/src/mapConfigurations.ts
```diff
@@ -208,13 +211,8 @@ export const getMapConfiguration = ({
...
   }
-  // @ts-expect-error | known issue; type definitions fit after merge in core
   return {
-    // @ts-expect-error | methods don't have a common interface
-    ...mapConfigurations[mode](reportServiceId),
-    // NOTE local plugin does not exist yet
-    // meldemichelNeuesAnliegen: {
-    //   afmUrl
-    // }
+    // @ts-expect-error | reportServiceId might be undefined, but that's catched above for relevant cases
```
dopenguin, line 215:
```suggestion
    // @ts-expect-error | reportServiceId might be undefined, but that's caught above for relevant cases
```

packages/clients/meldemichel/src/mapConfigurations.ts
```diff
@@ -161,15 +164,17 @@ const mapConfigurations = {
...
           },
         },
         coordinateSources: [], // to be set in addPlugins.ts
       },
       pins: commonPins,
       reverseGeocoder,
+      meldemichelAfmButton: {
+        displayComponent: true,
+        afmUrl,
+      },
```
dopenguin, line 177:
```suggestion
      meldemichelAfmButton: { afmUrl },
```
`displayComponent` already defaults to `true` as set in `packages/clients/meldemichel/src/polar-client.ts`

packages/clients/meldemichel/src/plugins/AfmButton/index.ts
```diff
@@ -0,0 +1,12 @@
+import Vue from 'vue'
+import { PluginOptions } from '@polar/lib-custom-types'
+import AfmButton from './AfmButton.vue'
+import language from './language'
+
+export default (options: PluginOptions) => (instance: Vue) =>
+  instance.$store.dispatch('addComponent', {
+    name: 'scale',
```
dopenguin, line 8:
```suggestion
    name: 'afmButton',
```
Or `meldemichelAfmButton`, if you insist :D

packages/clients/meldemichel/src/plugins/AfmButton/AfmButton.vue
```diff
@@ -0,0 +1,71 @@
...
+    {{ $t('common:plugins.meldemichelAfmButton.buttonText') }}
+  </v-btn>
+</template>
+
+<script lang="ts">
+import Vue from 'vue'
+import { mapGetters } from 'vuex'
+
+export default Vue.extend({
+  name: 'MeldemichelAfmButton',
```
dopenguin, line 22: Especially here it is clear, that the name should simply be `AfmButton`.

I was also thinking about naming the file `index.vue`, as the path `.../AfmButton/AfmButton.vue` looks weird

packages/clients/meldemichel/src/plugins/AfmButton/AfmButton.vue
```diff
@@ -0,0 +1,71 @@
+<template>
+  <v-btn
```
dopenguin, line 2: The tooltip including `Einen neuen Schaden melden` is missing

packages/clients/meldemichel/src/plugins/AfmButton/AfmButton.vue
```diff
@@ -0,0 +1,71 @@
...
+          text: 'plugins.meldemichelAfmButton.missingAddress',
+          timeout: 10000,
+        })
+      }
+    },
+  },
+})
+</script>
+
+<style lang="scss">
```
dopenguin, line 58:
```suggestion
<style lang="scss" scoped>
```

packages/clients/meldemichel/src/plugins/AfmButton/AfmButton.vue
```diff
@@ -0,0 +1,71 @@
+<template>
+  <v-btn
+    v-if="afmUrl"
+    :target="mapStateReady ? '_blank' : ''"
+    color="primary"
+    class="meldemichel-afm-button"
+    :href="mapStateReady ? afmUrl : '#'"
+    large
```
dopenguin, line 8:
```suggestion
```
Looks better without it in my opninion

packages/clients/meldemichel/src/plugins/AfmButton/AfmButton.vue
```diff
@@ -0,0 +1,71 @@
...
+    @click="click"
+    @keydown.space="click"
```
dopenguin, line 10: As `href` and `target` are added to `v-btn`, it is being rendered as an anchor `<a>`. Are you sure, that the intended behaviour applies for anchors as well?

I also find it a bit misleading having the anchor look like a button, but that may just be me 🤷

dopenguin (changes requested):
🏓 @warm-coolguy

packages/clients/meldemichel/src/mapConfigurations.ts
warm-coolguy: >Also, `[MODE.COMPLETE]` can be simpliefied to `COMPLETE`

It can be, but I figured relating to the ENUM indicates more obviously what's going on.

packages/clients/meldemichel/src/mapConfigurations.ts
warm-coolguy: https://github.com/Dataport/polar/issues/25

packages/clients/meldemichel/src/mapConfigurations.ts
warm-coolguy: Sounds good. https://github.com/Dataport/polar/pull/23/commits/7199419d10018cb5dfef40b80006fe450bf6e451

packages/clients/meldemichel/src/mapConfigurations.ts
warm-coolguy: It's settable from the outer `createMaps` and called `afmUrl` there. I figured the prefix would make more clear what it is. Just naming it `url` in `createMap` could be misleading, and I don't think we should have multiple names for one thing.

packages/clients/meldemichel/src/plugins/AfmButton/AfmButton.vue
warm-coolguy: Each plugin has a vue name that matches its folder name. Sometimes with a `src` in between, which makes it less obvious. Since we're in Meldemichel/AfmButton, `MeldemichelAfmButton` makes perfect sense as a name.

packages/clients/meldemichel/src/plugins/AfmButton/AfmButton.vue
warm-coolguy: https://github.com/Dataport/polar/pull/23/commits/e6cfedf6a6eefe86b6b85cd5420cef31d96c9aba

packages/clients/meldemichel/src/plugins/AfmButton/AfmButton.vue
warm-coolguy: With this, the effect on `.v-btn__content` would not work. However, the CSS is practically scoped since it's in a wrapping class name. There should be no leakage concerns.

packages/clients/meldemichel/src/plugins/AfmButton/AfmButton.vue
warm-coolguy: Our opinions differ in this decision, albeit I did not consider which version is more aesthetic. I fathomed the main interaction button (that allows proceeding in the reporting process) demands a lottle more weight and visibility. After all, it also helps with the procedure by toasting what is to be done in the client.

packages/clients/meldemichel/src/plugins/AfmButton/AfmButton.vue
warm-coolguy: While it's an anchor, it looks like a button, so both interactions might be expected. Allowing space for the interaction should not hurt since it can't be used for scrolling in the `COMPLETE` mode client anyway.

In which way is it misleading? Since it does not just hardlink static content, but continues a process, I deem this correct and in line with the following AfM process.

packages/clients/meldemichel/src/mapConfigurations.ts
warm-coolguy: https://github.com/Dataport/polar/pull/23/commits/d29bd8fb7d771552f4b08fa652f8f8ea539d5274

packages/clients/meldemichel/src/plugins/AfmButton/index.ts
warm-coolguy: https://github.com/Dataport/polar/pull/23/commits/3d3a27cac7ae413f174e29ad0b28413efe4e6f35

warm-coolguy:
> 🏓 @warm-coolguy

🏓 @dopenguin

packages/clients/meldemichel/src/mapConfigurations.ts
dopenguin: By changing the type to `keyof typeof MODE` that is still the case, even if the enum is not used for all the modes.

Though, if you insist on keeping it that way, I won't press you 🎩

packages/clients/meldemichel/src/plugins/AfmButton/AfmButton.vue
dopenguin: If one is not using `large`, it still seems reasonably important. With `large` it seems almost obstructive to me. As we don't have a requirement though -> 🎩

packages/clients/meldemichel/src/mapConfigurations.ts
dopenguin: In `core` (regarding layer configuration), in `addressSearch` (regarding search configuration) and in `reverseGeocoder` (regarding WPS configuration) it is always simply named `url` - why should this be different?

The name in `createMap` should stay this way, because the argument for clear usage can be made there.  
I'd say it is clear enough that `afmUrl` changes the value of `meldemichel.afmButton.url`; the lastest, when one looks in the README.md.

dopenguin (changes requested):
🏓 @warm-coolguy

packages/clients/meldemichel/src/mapConfigurations.ts
warm-coolguy: Renaming `url: afmUrl` considered harmful.

warm-coolguy:
> 🏓 @warm-coolguy
🏓 @dopenguin 

We could have a quote tower, you know?

packages/clients/meldemichel/src/mapConfigurations.ts
dopenguin: Talked this through ☎️

dopenguin (approved):
🚀 @warm-coolguy

PR #24, opened 2023-11-17
fix linebreak on ratio scale on narrow devices

warm-coolguy:
## Summary

There were linebreaks here. They're gone now.

Test Meldemichel clients on local GitHub pages with mobile device, if need be.


dopenguin (approved):
🚀 @warm-coolguy

Issue #27, opened 2023-11-20
Attributions are not being translated when the window is open


dopenguin:
**Describe the bug**
When switching the language, the attributions are not being translated if the window is open.  
If the window is closed and reopened, the values are translated.

**Expected behavior**
If a locale string is given for an attributions, they should always be in the currently selected language.

**Steps to reproduce**
1. `npm run snowbox`
2. Open Attributions
3. Switch language from `English` to `German`
4. The attributions stay in English

**Additional context**
In contrast to other information displayed to the user, the attributions are rendered via `v-html`. The issue probably occurs because the language switch is not actively watched.

PR #28, opened 2023-11-20
Add link to repository and license as required by EUPL v1.2

dopenguin:
## Summary

As required by the EUPL license, a link as well as the license are always linked to when the attributions plugin is used.

Start any client and open the attributions. When using `snowbox`, the locale switch can be tested with #27 in mind.


warm-coolguy (approved):
Bene. @dopenguin

PR #29, opened 2023-11-21
Chore/package json links

warm-coolguy:
## Summary

Adding repository links to all package.json files. This way NPM will refer back to our GitHub repository.

No CHANGELOG entries have been made since I don't think we need patch releases for this. The links will be there eventually.


dopenguin (approved):
🚀 @warm-coolguy

PR #31, opened 2023-11-22
Extend accessibility of IconMenu, Features rendered in it and the AfmButton

dopenguin:
## Summary

- Update all the `v-tooltips` to have `primary` as their background colour and `primaryContrast` as border colour.
- Add new `core` state variable `hasSmallDisplay` which is updated on `resize`.
- Hide the `v-tooltip`s by disabling them on small devices.

- [ ] Tooltips of `AfmButton`, `Export`, `Fullscreen`, `GeoLocation`, all plugins rendered in `IconMenu`, `LayerChooser` and `Zoom` are now shown in the described colour-scheme
- [ ] Tooltips of `AfmButton`, `Export`, `Fullscreen`, `GeoLocation`, all plugins rendered in `IconMenu`, `LayerChooser` and `Zoom` are now hidden if the widht of the window is smaller than `768px` or the height of the window is smaller than `480px`. This change should also work if the window is resized.


dopenguin:
I was also thinking about adding a wrapper component of `v-tooltip` to `core` removing the need to add `:disabled="hasSmallDisplay"` as well as the related `vuex` parts. 
I'm not sure if the added ease of usage would exceed the needed implementation time. What is your opinion on this, @warm-coolguy?

warm-coolguy:
> I was also thinking about adding a wrapper component of `v-tooltip` to `core` removing the need to add `:disabled="hasSmallDisplay"` as well as the related `vuex` parts. I'm not sure if the added ease of usage would exceed the needed implementation time. What is your opinion on this, @warm-coolguy?

Wouldn't it be more correct to nest it in `components`? Accessing the core's vuex store from there seems legitimate since it's a POLAR component and the core is *always* supposed to exist.

I'm not sure about development time. However, a wrapper component may reduce diverging implementations in the future. I propose trying and dropping this centralization should it not work within reasonable time.

warm-coolguy (changes requested):
1. I (personally) do not like the `primary` styling of tooltips, and it is (technically) adverse to what `primary` is supposed to do, which is directing the user towards major actionable contents. With these colours, tooltips lose their tooltippiness. I've checked the contrast of the original design against (worst case) white background (https://colourcontrast.cc/7e7e7e/ffffff) and agree that action needs to be taken. I propose making the text bolder and every-so-slightly darken the background color until AAA is met at around #767676. (If applied, please re-check on white background since tooltips are a little transparent, which I deem a nice effect.)
2. Removing tooltips on mobile devices seems off at first, since mobile users may need further context on usage, too. However, I agree actions needs to be taken due to their obstructiveness. I considered [close delay](https://vuetifyjs.com/en/api/v-tooltip/#props-close-delay) for a bit, but that may stress some users and still look wonky on overlapping contents? My suggestion: We may fill the informational gaps with additional `h` to the window contents. When glancing over the plugins, only some require this. The `LayerChooser` is pretty obvious as-is, but `Attributions` look like random garbage at first glance. An additional headline (that is probably identical to the tooltip) should fill this gap. Non-content buttons can't be hovered on mobile anyway without clicking, and on clicking their use becomes obvious immediately ...

warm-coolguy:
🏓 @dopenguin

dopenguin:
> 1. I (personally) do not like the `primary` styling of tooltips, and it is (technically) adverse to what `primary` is supposed to do, which is directing the user towards major actionable contents. With these colours, tooltips lose their tooltippiness. I've checked the contrast of the original design against (worst case) white background (https://colourcontrast.cc/7e7e7e/ffffff) and agree that action needs to be taken. I propose making the text bolder and every-so-slightly darken the background color until AAA is met at around #767676. (If applied, please re-check on white background since tooltips are a little transparent, which I deem a nice effect.)
> 
> 2. Removing tooltips on mobile devices seems off at first, since mobile users may need further context on usage, too. However, I agree actions needs to be taken due to their obstructiveness. I considered [close delay](https://vuetifyjs.com/en/api/v-tooltip/#props-close-delay) for a bit, but that may stress some users and still look wonky on overlapping contents? My suggestion: We may fill the informational gaps with additional `h` to the window contents. When glancing over the plugins, only some require this. The `LayerChooser` is pretty obvious as-is, but `Attributions` look like random garbage at first glance. An additional headline (that is probably identical to the tooltip) should fill this gap. Non-content buttons can't be hovered on mobile anyway without clicking, and on clicking their use becomes obvious immediately ...

Regarding 1:  
I took a look into test report regarding the accessibility of MELMI from a few years ago because we already had the tooltips in primary colour in the previous iteration. It point did not have a remark regarding its colour so I'll change it to your suggestion!

Regarding 2:  
I like the idea of moving the tooltip content to be displayed as part of the content of the plugin so I'll move it there, if required.

dopenguin:
🏓 @warm-coolguy

packages/clients/meldemichel/src/plugins/AfmButton/AfmButton.vue
```diff
@@ -1,5 +1,9 @@
 <template>
-  <v-tooltip :left="!hasSmallWidth" :top="hasSmallWidth">
+  <v-tooltip
+    :left="!hasSmallWidth"
+    :top="hasSmallWidth"
+    :disabled="hasSmallDisplay"
+  >
```
warm-coolguy, line 6:
```suggestion
  <v-tooltip :left="!hasSmallWidth" :disabled="hasSmallDisplay">
```
`top` and `disabled` are true in the same situation, right? In that case, top could be omitted. Or is there something configurable (planned) so that this should stay in?

warm-coolguy (changes requested):

warm-coolguy:
> 🏓 @warm-coolguy

🏓 @dopenguin 

that_was_easy.mp4

packages/clients/meldemichel/src/plugins/AfmButton/AfmButton.vue
dopenguin: `hasSmallWidth` and `hasSmallDisplay` are not the same variables:
- `hasSmallWidth` if the width of the map is smaller than `SMALL_DISPLAY_WIDTH`
- `hasSmallDisplay` is true, if either the height or the widht of the window` is smaller than `SMALL_DISPLAY_WIDTH`

warm-coolguy: Oh, I forgot about map width and client width. Thanks for clarification. 🎩

warm-coolguy (approved):

PR #32, opened 2023-11-22
Fix map overlay being visible on desktop if map is not in focus

dopenguin:
## Summary

Whenever the user focused on any element that was not the map, the map overlay was still being shown. This was especially annoying when scrolling the legend.

- `npm run snowbox`
- Open legend
- Scroll
- The overlay is not visible

One can now scroll on any part of the page other than the map and the overlay is only shown, once the cursor is actually on the `div` of the map.


warm-coolguy (approved):
🚀 @dopenguin

Issue #19, opened 2023-11-15
Toast Errors should not be dismissed automatically


warm-coolguy:
Toast errors disappear after some time. Errors should not disappear by themselves as decided here: https://github.com/Dataport/polar/pull/14#discussion_r1393242239

The client may not work correctly, and the error might contain relevant information to the user.

**TODO**

- [x] Remove timeouts for error toasts everywhere
- [x] Either ignore timeouts for errors (hard decision) or at least warn about timeouts on errors (softer)
- [x] Write an ADR regarding this decision

PR #36, opened 2023-11-29
No timeout for error toasts

raschju:
## Summary

* remove timeouts from error toasts
* ignore timeouts when set for an error toast
* add warning that timeout is ignored

*run client and check if error toasts close on their own
*check ADR in wiki

#19

### DIFF
--- a/packages/plugins/GeoLocation/src/store/actions.ts
+++ b/packages/plugins/GeoLocation/src/store/actions.ts
@@ -243,7 +243,7 @@ const actions: PolarActionTree<GeoLocationState, GeoLocationGetters> = {
       {
         type: 'error',
         text: 'common:plugins.geoLocation.button.tooltip.locationAccessDenied',
-        timeout: 5000,
+        timeout: 0,
       },
       { root: true }
     )
--- a/packages/plugins/GeoLocation/src/store/actions.ts
+++ b/packages/plugins/GeoLocation/src/store/actions.ts
@@ -162,7 +162,6 @@ const actions: PolarActionTree<GeoLocationState, GeoLocationGetters> = {
         ? {
             type: 'error',
             text: 'plugins.geoLocation.toast.boundaryError',
-            timeout: 0,
           }
         : {
             type: 'info',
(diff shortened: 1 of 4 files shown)


dopenguin:
The `'plugin/toast/addToast'` in `actions.onError` should be changed to `getters.toastAction` with an additional check if it is defined, just like in `actions.printPositioningFailed`

packages/plugins/GeoLocation/src/store/actions.ts
```diff
@@ -243,7 +243,7 @@ const actions: PolarActionTree<GeoLocationState, GeoLocationGetters> = {
       {
         type: 'error',
         text: 'common:plugins.geoLocation.button.tooltip.locationAccessDenied',
-        timeout: 5000,
+        timeout: 0,
```
dopenguin, line 246:
```suggestion
```
`timeout` not needed

packages/plugins/Toast/src/store/index.ts
```diff
@@ -48,9 +48,15 @@ const storeModule: PolarModule<ToastState, ToastState> = {
...
-        setTimeout(() => {
-          commit('removeToast', smearedToast)
-        }, toast.timeout)
+        if (toast.type !== 'error') {
+          setTimeout(() => {
+            commit('removeToast', smearedToast)
+          }, toast.timeout)
+        } else {
+          console.warn(
+            'Timeouts for error messages are disabled, they can only be dismissed manually.'
```
dopenguin, line 57:
```suggestion
            '@polar/plugin.toast: Timeouts for error messages are disabled, they can only be dismissed manually.'
```
I'd also be open for just `Toast`, but this is probably a clearer indication

dopenguin:
The `timeout` on the error toasts in `GeoLocation/src/store/action:printPositioningFailed` and `Pins/src/store/index:isCoordinateInBoundaryLayer` should be removed as well

dopenguin (changes requested):
Also mind https://github.com/Dataport/polar/pull/36#issuecomment-1831695375 and https://github.com/Dataport/polar/pull/36#issuecomment-1831714774

🏓 @raschju

dopenguin (approved):
🚀 @raschju

Issue #37, opened 2023-12-01
Documentation error regarding getter subscribe


warm-coolguy:
The documentation is wrong about how getters are subscribed to.

For example, the LoadingIndicator's README notes:

```js
map.subscribe('plugin/loadingIndicator/showLoader', (showLoader) => {
  /* This code is called on showLoader updates. */
})
```

This does nothing, as only state variables can be *subscribed* to. Getters must be watched. Example:

```js
mapInstance.$store.watch(
    (_, getters) => getters['plugin/loadingIndicator/showLoader'],
    (active) => {
        // effect
    }
)
```

With this issue, all README's should be checked, and code for getter examples be fixed.

PR #39, opened 2023-12-04
fix listening documentation

warm-coolguy:
There has been a mixup about state and getters in the plugin documentation. Since state and getters have to be listened to with deviating methods, a fix was in order.

Resolves: #37


dopenguin (approved):
🚀 @warm-coolguy

PR #40, opened 2023-12-04
refactor address search keyboard navigation

### LINKED ISSUE #26: AddressSearch keyboard navigation refactor
>If you have any suggestions to clean up the way the id of the next element is retrieved, I am open to 🏌️⛳

>I assume it would be shorter to write functions `getNextElement` and `getPreviousElement` that get the current element's ID and just navigate the DOM with `.nextSibling` and `.previousSibling`. Since we have one level of nesting (`<ul>`) to consider, we'd also need `parentElement` and `children` for switching lists and some skipping on at least `div` and `hr`. 🤔

_Originally posted in https://github.com/Dataport/polar/pull/13#discussion_r1396815624_

We may be able to write this solution more concise, but that's fully optional.

warm-coolguy:
## Summary

Prior implementation has been replaced with DOM navigation that is easier to reason about.

Check in DISH and Meldemichel. Relevant search strings to DISH are "Rathaus", "Rathausm", and "2-20/Rathausstraße".

Resolves #26

### DIFF
--- a/packages/plugins/AddressSearch/src/components/Results.vue
+++ b/packages/plugins/AddressSearch/src/components/Results.vue
@@ -158,11 +158,11 @@ export default Vue.extend({
       const focus = ['BUTTON', 'LI']
       const sibling = down ? 'nextElementSibling' : 'previousElementSibling'
 
-      let searchBase: Element | undefined = originalTarget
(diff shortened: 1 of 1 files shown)


dopenguin (approved):
🚀 @warm-coolguy

PR #42, opened 2023-12-05
implement clustering and cluster styles

warm-coolguy:
## Summary

- Vector layers can now be changed to cluster features
- Clusters of more than one feature are represented by a multi-marker
- Clicking a multi-marker zooms into the bounding box of its child elements
- Clicking a multi-marker on maximum zoom instead selects the multi-marker
- The WFS list correctly lists not clusters, but their children with the actual information
- The tooltips work correctly on single features (which are 1-element clusters, technically)
- The Filter works appropriately (i.e. filters children, from which clusters are built)

Some optimization took place since clusters tended to overly redraw, resulting in a frozen browser on the multiple thousand reports in the map.

Run Meldemichel to see if clusters behave as expected and interact correctly with the Filter and (as far as implemented) WFS list.


packages/core/src/utils/markers.ts
```diff
@@ -0,0 +1,108 @@
...
+export interface MarkerStyle {
+  strokeWidth?: string | number
+  stroke?: string
+  fill?: string
+}
+
+export type GetMarkerFunction = (style: MarkerStyle, multi: boolean) => Style
+
+export const imgSize = [23, 32]
+export const imgSizeMulti = [35, 32]
```
dopenguin, line 13: These seem like awfully specific values - please add a comment on why we got the magic here

packages/core/src/utils/markers.ts
```diff
@@ -0,0 +1,108 @@
+import Style from 'ol/style/Style'
+import Icon from 'ol/style/Icon'
+
+export interface MarkerStyle {
+  strokeWidth?: string | number
+  stroke?: string
+  fill?: string
+}
```
dopenguin, line 8: The type also exists in the types package - mind deleting this and changing the paths?

packages/core/src/utils/markers.ts
```diff
@@ -0,0 +1,108 @@
...
+
+export type GetMarkerFunction = (style: MarkerStyle, multi: boolean) => Style
```
dopenguin, line 10:
```suggestion
type GetMarkerFunction = (style: MarkerStyle, multi: boolean) => Style
```
Only used locally. If it should be used in any other place, please place it in the `types.ts`

packages/core/src/utils/addClusterStyle.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export const addClusterStyle = (mapConfiguration: MapConfig): MapConfig => {
+  mapConfiguration.layerConf.forEach((layerConfiguration) => {
+    if (layerConfiguration.clusterDistance !== undefined) {
+      const style =
+        mapConfiguration.extendedMasterportalapiMarkers?.defaultStyle ||
+        defaultStyle
+      // @masterportal/masterportalapi hook
+      layerConfiguration.style = (feature: Feature) => {
+        const visibleFeaturesCount = (feature.get('features') || []).filter(
```
dopenguin, line 17: `feature.get('features')` doesn't have a type attached thus it is `any`, making `visibleFeaturesCount` `any` as well.

Easiest fix would be to just add `: number` behind `visibleFeaturesCount`

packages/core/src/components/MapContainer.vue
```diff
@@ -75,7 +76,9 @@ export default Vue.extend({
     const map = api.map.createMap(
       {
         target: this.$refs['polar-map-container'],
-        ...this.mapConfiguration,
+        ...(this.mapConfiguration.extendedMasterportalapiMarkers
+          ? addClusterStyle
+          : (x) => x)(this.mapConfiguration),
```
dopenguin, line 81:
```ts
const config = this.mapConfiguration.extendedMasterportalapiMarkers
  ? addClusterStyle(this.mapConfiguration)
  : this.mapConfiguration
```
🏌️ ⛳

packages/core/src/utils/markers.ts
```diff
@@ -0,0 +1,108 @@
...
+  <g stroke="${stroke}" stroke-width="${strokeWidth}" fill="${fill}" fill-rule="nonzero">
+    <path d="M22.584 1C15.101 1 9 7.101 9 14.584c0 8.103 7.865 20.448 11.63 25.93a2.36 2.36 0 0 0 3.908 0c3.766-5.482 11.63-17.922 11.63-25.93C36.168 7.054 30.068 1 22.584 1z"/>
+    <path d="M18.584 1C11.101 1 5 7.101 5 14.584c0 8.103 7.865 20.448 11.63 25.93a2.36 2.36 0 0 0 3.908 0c3.766-5.482 11.63-17.922 11.63-25.93C32.168 7.054 26.068 1 18.584 1z"/>
+    <path d="M14.584 1C7.101 1 1 7.101 1 14.584c0 8.103 7.865 20.448 11.63 25.93a2.36 2.36 0 0 0 3.908 0c3.766-5.482 11.63-17.922 11.63-25.93C28.168 7.054 22.068 1 14.584 1z"/>
+  </g>
+</svg>
+`)}`
+
+// center bottom of marker 📍 is intended to show the spot
+export const anchor = [0.5, 1]
```
dopenguin, line 62: All of these exported `const` are never used outside of this file.

Is there an intended use for them that is not clear yet?

packages/core/src/utils/resolveClusterClick.ts
```diff
@@ -0,0 +1,16 @@
+import { Feature, Map } from 'ol'
+import { createEmpty, extend } from 'ol/extent'
+
+export const resolveClusterClick = (map: Map, feature: Feature) => {
+  const features = feature.get('features')
```
dopenguin, line 5: `feature.get('features')` resolves to any, thus the iteration with `forEach` is typed with `any` as well.

Please add an appropriate type

packages/core/src/utils/markers.ts
```diff
@@ -0,0 +1,108 @@
...
+    if (cache.size > 1000) {
+      console.warn(
+        `@polar/core/src/utils/markers.ts: 1000+ styles have been created. This is possibly a memory leak. Please mind that the methods exported by this module are memoized. You *may* be calling the methods with constantly newly generated objects, or maybe there's just a lot of styles.`
+      )
+    }
+    return markerStyle
+  }
+}
+
+const getGetStyle =
```
dopenguin, line 87: Would `getStyleFunction` work as well or is it just another evil?

packages/core/src/utils/markers.ts
```diff
@@ -0,0 +1,108 @@
...
+
+export const getDefaultStyle = memoizeStyle(getGetStyle(defaultFill))
+
+export const getHoveredStyle = memoizeStyle(
+  memoizeStyle(getGetStyle(defaultHoverFill))
+)
+
+export const getSelectedStyle = memoizeStyle(
+  memoizeStyle(getGetStyle(defaultSelectionFill))
+)
```
dopenguin, line 108: Why does `memoizeStyle` receive itself here?

packages/core/src/utils/markers.ts
```diff
@@ -0,0 +1,108 @@
...
+      console.warn(
+        `@polar/core/src/utils/markers.ts: 1000+ styles have been created. This is possibly a memory leak. Please mind that the methods exported by this module are memoized. You *may* be calling the methods with constantly newly generated objects, or maybe there's just a lot of styles.`
+      )
+    }
+    return markerStyle
+  }
+}
+
+const getGetStyle =
+  (fallbackFill: string): GetMarkerFunction =>
```
dopenguin, line 88: The fill is never fallen back to, as it is the only one being used in this scope - `fill` suffices

packages/plugins/Gfi/src/store/getters.ts
```diff
@@ -158,29 +159,19 @@ const getters: PolarGetterTree<GfiState, GfiGetters> = {
...
     // trigger getter on those who indicate feature change possibility
-    noop(
-      clientHeight,
-      clientWidth,
-      center,
-      zoomLevel,
-      visibilityChangeIndicator
-    )
+    noop(clientHeight, clientWidth, center, zoomLevel)
+    noop(visibilityChangeIndicator)
```
dopenguin, line 163: Why in two calls of doing ✨ nothing ✨ ?

packages/plugins/Gfi/src/store/getters.ts
```diff
@@ -191,7 +182,7 @@ const getters: PolarGetterTree<GfiState, GfiGetters> = {
         )
           .filter(isVisible)
           .map((feature) => {
-            feature.set('_gfiLayerId', layer.get('id'))
+            feature.set('_gfiLayerId', layer.get('id'), true)
```
dopenguin, line 185: Why should the update be silent? Or rather: Why wasn't it done before and only now became relevant?

packages/clients/meldemichel/package.json
```diff
@@ -38,5 +39,8 @@
     "@polar/plugin-scale": "^1.0.0",
     "@polar/plugin-toast": "^1.0.0",
     "@polar/plugin-zoom": "^1.0.0"
+  },
+  "peerDependencies": {
+    "@masterportal/masterportalapi": "2.8.0"
```
dopenguin, line 44: Why is it needed as a peerDependency?

dopenguin (changes requested):
🏓 @warm-coolguy

packages/clients/meldemichel/src/index.html
```diff
@@ -24,7 +24,7 @@
       // //  SIMULATE MODES DURING DEVELOPMENT  // //
       // // // // // // // // // // // // // // // //
       const mode = 'COMPLETE' // COMPLETE, REPORT, SINGLE
-      const stage = true // true, false
+      const stage = false // true, false
```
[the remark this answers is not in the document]
warm-coolguy: It's not necessary, but I deem it advantageous and intentionally left it in. The stage data is not representative of what's currently on the services, neither in quality nor quantity, and there's some report types no longer active still on the stage server, which is confusing.

The stage systems are still required for actual deployments to test the whole workflow, but I'd say during development we should get as close to the final data as possible.

packages/core/CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## unpublished
 
-- Feature: Add hovered and selected feature to vuex store. This is an optional functionality that has to be explicitly enabled and works with the `@masterportal/masterportalapi` default marker design. See configuration parameter `extendedMasterportalapiMarkers`.
+- Feature: Add hovered and selected feature to vuex store that supports clustering. This is an optional functionality that has to be explicitly enabled and works with the `@masterportal/masterportalapi` default marker design. See configuration parameter `extendedMasterportalapiMarkers`.
```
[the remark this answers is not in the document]
warm-coolguy: https://github.com/Dataport/polar/pull/42/commits/038eb4aec1af9f63e42b6cb3936b24c10a24f8b1

packages/plugins/Gfi/src/store/actions.ts
```diff
@@ -132,7 +132,7 @@ const actions: PolarActionTree<GfiState, GfiGetters> = {
           layer
             // @ts-expect-error | layers reaching this have a source
             .getSource()
-            .on('changefeature', debouncedVisibilityChangeIndicator)
+            .on('addfeature', debouncedVisibilityChangeIndicator)
```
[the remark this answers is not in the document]
warm-coolguy: The purpose of this action (`debouncedVisibilityChangeIndicator`) is to update a computed value in another position. Namely, when not moving the map at all, changing the filter and then checking the WFS list would previously result in an old view. That view is now linked to this indicator.

It is not `'addfeature'` anymore since the mechanism to change visibility has been modified. Previously, all features were *modified* one after another. Now, the whole set is copied and re-added after cleaning the source. While counter-intuitive at first, this results in far less follow-up computations by the cluster layer (that would, else, on any `'changefeature'` completely re-calculate clustering and block the browser up to 30s).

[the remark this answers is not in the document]
warm-coolguy: https://github.com/Dataport/polar/pull/42/commits/833e86f8155886dfb12ab5a37e93b58e52210ea4

Writing about this made me do another change. Now, features are no longer copied, which is possible by "unlinking" them from the source before all changes take place. Since no copies are made anymore, less unexpected behaviour should take place when referring to features by reference in the future.

packages/core/src/utils/markers.ts
warm-coolguy: https://github.com/Dataport/polar/pull/42/commits/3b5b80e5b16a7094a6819ca05fabe1a176adc5c8

I don't remember why they're separate to the rest of the svg strings. Maybe they were intended to be configurable at some point. Changing the height values actually scales the markers. We should keep this in mind, should a need arise.

packages/core/src/utils/markers.ts
warm-coolguy: Don't mind if I do. https://github.com/Dataport/polar/pull/42/commits/cdb594111def412933b2b2fcab5b4a1944afcd8e

packages/core/src/utils/addClusterStyle.ts
warm-coolguy: https://github.com/Dataport/polar/pull/42/commits/16783312ff0efe6785e7ef1553077eb6668c5d6a Easiest fix applied.

packages/core/src/components/MapContainer.vue
warm-coolguy: https://github.com/Dataport/polar/pull/42/commits/c0fd2b7a0d249bad3c0c5589baf23e672bd85f3b

packages/core/src/utils/markers.ts
warm-coolguy: I figured that having these values available could help with re-use of the marker designs later by allowing to use lines like `import { makeMarker } from '@polar/core/src/utils/markers'` in a client.

packages/core/src/utils/resolveClusterClick.ts
warm-coolguy: https://github.com/Dataport/polar/pull/42/commits/0e33f6c9d4f21e5e8e8e8fa62d8e48ed4b78a908

packages/core/src/utils/markers.ts
warm-coolguy: That's less fun, but a good idea. https://github.com/Dataport/polar/pull/42/commits/aea03ce9e924242ae75ea5a72199044c873bfa58

packages/core/src/utils/markers.ts
warm-coolguy: Better safe than sorry. https://github.com/Dataport/polar/pull/42/commits/da369b3c11f4b0f5e0de1421eb86de443b019f62

packages/core/src/utils/markers.ts
warm-coolguy: Oh, it is fallen back to. It's intentionally set first so that any `fill` parameter in the `style` (coming from configuration) spread later may override it.

packages/plugins/Gfi/src/store/getters.ts
warm-coolguy: Doing it in one line makes the auto-linter spread it to seven lines length, breaking the 50-lines-per-function linting rule. I am keeping my poker face up while writing this.

packages/plugins/Gfi/src/store/getters.ts
warm-coolguy: Since changing a feature's attributes may result in effects on the style and clustering, doing this loud results in a cluster recomputation and rerender. It became relevant on trying to use the productive service.

packages/clients/meldemichel/package.json
warm-coolguy: https://github.com/Dataport/polar/pull/42/commits/7a19d98125c12b11cbbc1785b5db3403f4dd5ea9

I guess I moved the code requiring it to the core later, but don't remember.

warm-coolguy:
> 🏓 @warm-coolguy

🏓 @dopenguin

packages/core/src/utils/markers.ts
dopenguin: Hm, I think we never really decided on it, but wouldn't that fit more into a lib-package then?

packages/core/src/utils/markers.ts
dopenguin: This information would fit great as a comment!

packages/plugins/Gfi/src/store/getters.ts
dopenguin: This seems like information I would gladly stumble upon if we ever have to refactor this!

dopenguin:
> > 🏓 @warm-coolguy
> 
> 🏓 @dopenguin

🏓 @warm-coolguy

packages/core/src/utils/markers.ts
dopenguin: Just reread the code 🎩

packages/core/src/utils/markers.ts
warm-coolguy: https://github.com/Dataport/polar/pull/42/commits/c799e39e59471fda0baf3a2357cc45cc10aa09af

Narrowed it for now.

packages/plugins/Gfi/src/store/getters.ts
warm-coolguy: https://github.com/Dataport/polar/pull/42/commits/dd558a4ef2a6219dc82e6ebb51b511f84547d843 De-arcaned.

warm-coolguy:
> > > 🏓 @warm-coolguy
> > 
> > 
> > 🏓 @dopenguin
> 
> 🏓 @warm-coolguy

🏓 @dopenguin

dopenguin (approved):
> > > > 🏓 @warm-coolguy
> > > 
> > > 
> > > 🏓 @dopenguin
> > 
> > 
> > 🏓 @warm-coolguy
> 
> 🏓 @dopenguin

🚀 @warm-coolguy

PR #43, opened 2023-12-07
Feature/wfs list and markers interwoven

warm-coolguy:
## Summary

- Hovering GFI list features highlights their marker counterparts in the map.
- Hovering markers in the map highlights their counterparts in the GFI list (if visible).
- Clicking a GFI list feature zooms into the feature and selects it, showing its details; if it's a non-resolvable cluster feature at maximum zoom, its contents can be switched between, and the correct page of the clicked entry is initially open.
- Clicking a feature in the map selects it in the GFI list; if it's a non-resolvable cluster feature, its contents can be switched between. If the GFI was closed previously, it is now open.

Check interactivity between Feature List and map markers by clicking/hovering the elements. An unresolvable cluster feature ad max zoom level is at the time of writing located in "Annenstraße 9".

## TODO

- [x] This PR should be reviewed and must be merged **after** https://github.com/Dataport/polar/pull/42.
- [x] Pull file `packages/core/src/vuePlugins/actions/useExtendedMasterportalapiMarkers/index.ts` apart for better readability.
- [x] Add missing implementation for hover effects.


packages/clients/meldemichel/src/plugins/Gfi/Feature.vue
dopenguin, line 1: When I click a feature in the list, it zooms to the feature on prod, if it is in a cluster in two iterations.  

With the current changes, the map only centers on the feature and doesn't zoom on it. Features in a cluster stay in a cluster.

no_full_zoom

packages/core/src/vuePlugins/actions/useExtendedMasterportalapiMarkers/index.ts
```diff
@@ -7,28 +7,47 @@ import {
...
+      duration: 400,
+      easing: easeOut,
+    })
+  }
+}
+
 // eslint-disable-next-line max-lines-per-function
 export function useExtendedMasterportalapiMarkers(
-  { getters, commit }: PolarActionContext<CoreState, CoreGetters>,
+  this: Store<CoreState>,
```
dopenguin, line 37: Generally, `PolarActionHandler` would be the type to use for the whole function but as this seems a bit weird to type, adding `PolarStore` would be better

packages/core/src/vuePlugins/actions/useExtendedMasterportalapiMarkers/index.ts
```diff
@@ -44,6 +63,70 @@ export function useExtendedMasterportalapiMarkers(
...
+
+    let selectedCluster = feature.get('features') ? feature : null
+
+    if (selectedCluster === null) {
+      // @ts-expect-error | Really, it's there!
+      const candidates = map
+        .getLayers()
+        .getArray()
+        .find((layer) => layer.get('id') === feature.get('_gfiLayerId'))
+        // @ts-expect-error | Really, it's there!
```
dopenguin, line 81:
```suggestion
        // @ts-expect-error | The gfi layer has a source defined if this action is called
```

packages/core/src/vuePlugins/actions/useExtendedMasterportalapiMarkers/index.ts
```diff
@@ -44,6 +63,70 @@ export function useExtendedMasterportalapiMarkers(
...
+          map.getView().calculateExtent(map.getSize()),
+          map.getView().getProjection()
+        )
+
+      selectedCluster = candidates.find((candidate) =>
+        candidate.get('features').includes(feature)
+      )
+    }
+
+    selectedCluster?.setStyle(
```
dopenguin, line 93: Writing something like 

```ts
 const selectedCluster: Feature<Geometry> = feature.get('features')
      ? feature
      : // @ts-expect-error | Really, it's there!
        map
          .getLayers()
          .getArray()
          .find((layer) => layer.get('id') === feature.get('_gfiLayerId'))
          // @ts-expect-error | Really, it's there!
          .getSource()
          .getFeaturesInExtent(
            map.getView().calculateExtent(map.getSize()),
            map.getView().getProjection()
          )
          .find((candidate) => candidate.get('features').includes(feature))
```
Seems more fitting here, as the `selectedCluster` is always defined.

The addition of `: Feature<Geometry>` would be needed for the same reason as `candidates`.

dopenguin (changes requested):
Started the review but haven't gone all the way. Some breadcrumbs to start

🏓 @warm-coolguy

packages/clients/meldemichel/src/plugins/Gfi/Feature.vue
warm-coolguy: What do you think about it? Do you like this behaviour?

While it wasn't an intentional decision to leave it at that, I've since started to feel like this is a more pleasent behaviour. More control is left at the user's discretion.

At this point, we may decide to go either route. It shouldn't be difficult to switch, but delivering in this state might make for an interesting proposal.

Or am I just foolishly falling in love with a bug?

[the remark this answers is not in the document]
warm-coolguy: I've found that, too, and feel like this is actually a bug. At this zoom level and in that context, the singular marker hardly makes a difference to the grouped marker regarding understanding where something is. Also, it even stays after getting filtered; seems like it's a copy kept while the selection is on it.

I'd prefer to not replicate this decision.

packages/core/src/vuePlugins/actions/useExtendedMasterportalapiMarkers/index.ts
warm-coolguy: https://github.com/Dataport/polar/pull/43/commits/d0a21113aef5e373990d9fc9efc3be512bea4a08 Agreed.

packages/core/src/vuePlugins/actions/useExtendedMasterportalapiMarkers/index.ts
```diff
@@ -44,6 +63,70 @@ export function useExtendedMasterportalapiMarkers(
...
+      selected?.setStyle(undefined)
+      selected = null
+      return
+    }
+
+    let selectedCluster = feature.get('features') ? feature : null
+
+    if (selectedCluster === null) {
+      // @ts-expect-error | Really, it's there!
+      const candidates = map
```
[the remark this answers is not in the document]
warm-coolguy: https://github.com/Dataport/polar/pull/43/commits/22b33f6db450c89955bfca005e5f0f49235220c3

packages/core/src/vuePlugins/actions/useExtendedMasterportalapiMarkers/index.ts
warm-coolguy: https://github.com/Dataport/polar/pull/43/commits/5fb1d59dbf3230b41c342572ba74871519b2bb6a

Oh, nice.

warm-coolguy:
> Started the review but haven't gone all the way. Some breadcrumbs to start
> 
> 🏓 @warm-coolguy

**NOM NOM NOM NOM**

🏓 @dopenguin

packages/clients/meldemichel/src/plugins/Gfi/Feature.vue
dopenguin: I took a look at how google maps handles this and the new behaviour is the same as with google maps. Thus, we can leave it as is 🎩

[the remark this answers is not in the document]
dopenguin: ☎️ this was not an intended feature from the get-go. We leave it, as it is implemented here

dopenguin (changes requested):
Made some small changes myself, looks good altogether!  
After the last few threads are resolved, we are g2g.

dopenguin:
> > Started the review but haven't gone all the way. Some breadcrumbs to start
> > 🏓 @warm-coolguy
> 
> **NOM NOM NOM NOM**
> 
> 🏓 @dopenguin

🏓 @warm-coolguy

packages/clients/meldemichel/src/plugins/Gfi/Feature.vue
```diff
@@ -89,9 +80,22 @@ export default Vue.extend({
...
+    currentProperties(): GeoJsonProperties {
+      return { ...this.windowFeatures[this.visibleWindowFeatureIndex] }
+    },
+  },
+  mounted() {
+    this.setActionButton({
+      component: ActionButtons,
+      props: {},
+    })
+  },
```
dopenguin, line 92: This is not properly added, as it happens too late. It is only visible, if the feature is clicked a second time.

Check with the workaround in DISH, how that is handled there.

packages/plugins/Gfi/src/components/List.vue
```diff
@@ -46,7 +46,15 @@
...
           :three-line="listText.length === 3"
+          :class="{
+            'gfi-feature-list-item-hovered': isFeatureHovered(feature),
+          }"
           @click="setOlFeatureInformation(feature)"
+          @keydown.space.prevent="setOlFeatureInformation(feature)"
+          @mouseleave="unhover"
+          @mouseover="hover(feature)"
+          @focusout="unhover"
+          @focus="hover(feature)"
```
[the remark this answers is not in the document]
warm-coolguy: https://github.com/Dataport/polar/issues/52

packages/core/src/vuePlugins/actions/useExtendedMasterportalapiMarkers/index.ts
```diff
@@ -0,0 +1,189 @@
...
+
+  map
+    .getLayers()
+    .getArray()
+    .filter(layerFilter)
+    .forEach((layer) => {
+      // only vector layers reach this
+      const source = (layer as VectorLayer<VectorSource>).getSource()
+      if (source !== null) {
+        // @ts-expect-error | Undocumented hook.
```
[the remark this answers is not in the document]
warm-coolguy: We could remember and reuse the keyword "Undocumented". 🤔 Writing a separate update guide is probably a different task altogether and I fear it'd become an outdated/reality-diverging document quickly anyway.

packages/plugins/Gfi/README.md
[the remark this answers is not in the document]
warm-coolguy: The FeatureList is described in the GFI's README in chapter `gfi.featureList`. There's an entry `bindWithCoreHoverSelect` describing that

>the hover/select fields in the core's state will be listened to and interacted with. This will result in a bilateral hovering and selecting of features with the core.

I feel like this is sufficient; or where do you miss the annotation? I'd prefer not to mention it in the core since the core should be as plugin-agnostic as possible.

packages/clients/meldemichel/src/plugins/Gfi/Feature.vue
warm-coolguy: I don't see a workaround (Meldemichel GFI does the same as DISH GFI?) and I'm currently confused as of why there's a need for one anyway. Maybe we can fix the need for one eventually. But for today, what's the arcana here?

warm-coolguy:
> > > Started the review but haven't gone all the way. Some breadcrumbs to start
> > > 🏓 @warm-coolguy
> > 
> > 
> > **NOM NOM NOM NOM**
> > 🏓 @dopenguin
> 
> 🏓 @warm-coolguy

🏓 @dopenguin

packages/plugins/Gfi/README.md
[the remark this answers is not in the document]
dopenguin: There is no entry in the changelog

packages/core/src/vuePlugins/actions/useExtendedMasterportalapiMarkers/index.ts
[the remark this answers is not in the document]
dopenguin: ☎️ Opened an internal ticket

packages/clients/meldemichel/src/plugins/Gfi/Feature.vue
dopenguin: #55

warm-coolguy: https://github.com/Dataport/polar/issues/55

dopenguin (approved):
> > > > Started the review but haven't gone all the way. Some breadcrumbs to start
> > > > 🏓 @warm-coolguy
> > > > 
> > > > 
> > > **NOM NOM NOM NOM**
> > > 🏓 @dopenguin
> > 
> > 🏓 @warm-coolguy
> 
> 🏓 @dopenguin 

🚀 @warm-coolguy

PR #46, opened 2023-12-11
hide fullscreen on iphone (not supported)

warm-coolguy:
## Summary

The fullscreen button is hidden on iPhone devices (and other devices that do not support any of the used fullscreen methods).

This is due to Apple simply not supporting it on iPhone.

https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API#browser_compatibility

Run the client in dev mode and check with any device.

### DIFF
--- a/packages/plugins/Fullscreen/CHANGELOG.md
+++ b/packages/plugins/Fullscreen/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Feature: Hide obstructive tooltip on small devices.
 - Fix: Documentation error regarding plugin state.
+- Fix: Since iPhones do not support fullscreen mode, the fullscreen button is hidden on such devices.
 
 ## 1.0.0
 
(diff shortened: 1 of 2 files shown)


packages/plugins/Fullscreen/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Feature: Hide obstructive tooltip on small devices.
 - Fix: Documentation error regarding plugin state.
+- Fix: Since iPhones do not support fullscreen mode, the fullscreen button is hidden on such devices.
```
dopenguin, line 7:
```suggestion
- Fix: Since iPhones do not support fullscreen mode, the fullscreen button is hidden on such devices. See [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API#browser_compatibility) for more information.
```

warm-coolguy:
> 🏓 @warm-coolguy

🏓 @dopenguin

dopenguin (approved):
🚀 @warm-coolguy

PR #48, opened 2023-12-11
Update MoveHandle to only be rendered once

dopenguin:
## Summary

- Update the usage of the MoveHandle to be a singleton
- Opening any plugin in the IconMenu as well as a GFI now opens in the MoveHandle on mobile and closes a previously open one
- Fixed an issue with images in GFI constantly triggering onload

## TODO

- [x] Changelogs are still misssing, review is still ready
- [x] Update Filter to propagate its change in height when something is opened / closed

DISH and Meldemichel should be checked on Desktop and Mobile for functionality of IconMenu and GFI.


warm-coolguy (changes requested):
Unbenannt

After `npm install`, there's this issue on running any (tried DISH and Meldemichel) client.

packages/clients/dish/src/plugins/Gfi/Content.vue
```diff
@@ -186,12 +173,24 @@ export default Vue.extend({
...
     },
   },
+  beforeCreate() {
+    if (this.hasWindowSize && this.hasSmallWidth) {
+      this.setActionButton({
+        component: ActionButton,
+        props: { exportProperty: this.exportProperty },
+      })
+    }
+  },
```
warm-coolguy, line 183: While the `v-if` above updates, this does not. I propose using two watchers (on hasWindowSize/hasSmallWidth) that both fire the same method which sets as action button to the GFI plugin either the object in lines 178-181 or `null`, depending on the two variables' new combinatory state (that may, while we're at it, also be a computed value that's a boolean – that way we prevent mistakes on later changes in the logic by funneling all truth through a single value position).

packages/clients/meldemichel/src/plugins/Gfi/Feature.vue
```diff
@@ -1,17 +1,11 @@
...
-      fa-grip-lines
-    </v-icon>
-    <v-card-actions>
+    <v-card-actions v-if="!hasWindowSize || !hasSmallWidth">
       <!-- TODO implement when vector clusters are done
       <v-btn>Prev</v-btn>
       <v-btn>Next</v-btn>
       -->
-      <v-spacer></v-spacer>
+      <v-spacer />
```
warm-coolguy, line 8: Both ways to model this are currently in the code base. If it matters to you, there's probably a linting rule that would autofix this globally for us. You e.g. added it in `packages/components/MoveHandle.vue` in HTML style, but seem to prefer self-closing?

🎩 since it doesn't really matter.

dopenguin:
> Unbenannt
> 
> After `npm install`, there's this issue on running any (tried DISH and Meldemichel) client. This is not the case on the main branch.

Fixed.

warm-coolguy (changes requested):
On resizing a narrow page, the GFI window jumps back down. This can easily be reconstructed on desktop with a narrow browser. This behaviour is not present on production (https://efi2.schleswig-holstein.de/dish/dish_client/index.html), where the GFI window instead keeps it position. I imagine this may be annoying on switching between landscape and portrait and would prefer it to stay the same *as close as possible*.

packages/plugins/LayerChooser/src/components/LayerChooser.vue
warm-coolguy, line 1: height_issue

I can produce this view by having the options open and closing them. The MoveHandle should snap to bottom. In the same fashion, when I pulled the LayerChooser up, and then switch to the options, I cannot pull them further up. This results in unreachable options on narrow devices.

warm-coolguy (changes requested):

packages/components/MoveHandle.vue
warm-coolguy, line 1: On resizing a narrow page, the GFI window jumps back down. This can easily be reconstructed on desktop with a narrow browser. This behaviour is not present on production (https://efi2.schleswig-holstein.de/dish/dish_client/index.html), where the GFI window instead keeps it position. I imagine this may be annoying on switching between landscape and portrait and would prefer it to stay the same as close as possible.

warm-coolguy (changes requested):

packages/plugins/Gfi/src/components/List.vue
warm-coolguy, line 1: whoopsie

The last page may be shorter and is allowed to be scrolled further up.

packages/components/MoveHandle.vue
warm-coolguy, line 1: The Filter element may change its height, but the MoveHandle does not recognize it.

warm-coolguy (changes requested):

warm-coolguy:
🏓 @dopenguin

packages/components/MoveHandle.vue
warm-coolguy: Nvm, this was noted as a TODO initially.

packages/plugins/LayerChooser/src/components/LayerChooser.vue
dopenguin: 2a04354f5b269dd579dacd6e6dd4df52dbe67a23 fixed

packages/clients/dish/src/plugins/Gfi/Content.vue
```diff
@@ -1,23 +1,7 @@
 <template>
   <v-card class="dish-gfi-content">
```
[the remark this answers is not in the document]
dopenguin: `box-shadow` removed if content is displayed in MoveHandle a8de054b

Export button stuff under the other thread

packages/components/MoveHandle.vue
dopenguin: Added the feature not only for gfi but generally 2f7c1bcd

packages/clients/dish/src/plugins/Gfi/Content.vue
dopenguin: Fixed 08acc17d

dopenguin:
> 🏓 @dopenguin

🏓 @warm-coolguy

packages/components/MoveHandle.vue
```diff
@@ -71,16 +113,26 @@ export default Vue.extend({
...
+    handleElement.style.width = '100%'
+    handleElement.style['z-index'] = 1
+    handleElement.style.left = '0'
+    handleElement.style.top = `${calculateTop(
+      top,
+      this.$root.$el.clientHeight,
+      this.maxHeight
+    )}px`
+    this.resizeObserver = new ResizeObserver(this.updateMaxHeight)
+    this.resizeObserver.observe(handleElement)
```
warm-coolguy, line 127: I feel like we ought to unobserve the handleElement beforeDestroy, albeit I can't quite put my finger on how it could lead to erroneous behaviour not to do it. 🎩 at your discretion.

warm-coolguy (changes requested):

packages/components/MoveHandle.vue
warm-coolguy, line 1: 1. Open browser in wide mode
2. Open Filter
3. Make browser narrower
4. Content is invisible and only becomes visible upon closing and re-opening the filter

While this should be uncommon and is not a dealbreaker, I strongly feel like some additional (and maybe conditional) call in the MoveHandle's `mounted` should suffice to resolve this issue. If it's easy, let's go peasy lemon squeezy. If it's not, 🎩 at your discretion.

warm-coolguy (changes requested):

warm-coolguy:
> > 🏓 @dopenguin
> 
> 🏓 @warm-coolguy

🏓 @dopenguin

packages/components/MoveHandle.vue
dopenguin: Done 3e6fdce8

packages/components/MoveHandle.vue
dopenguin: ☎️

dopenguin:
> > > 🏓 @dopenguin
> > 
> > 
> > 🏓 @warm-coolguy
> 
> 🏓 @dopenguin

🏓 @warm-coolguy

warm-coolguy (approved):
🚀 @dopenguin

PR #49, opened 2023-12-12
melmi colour scheme proposal

warm-coolguy:
## Summary

Pin colour changes.

Run the client with a browser plugin to simulate colour blindness and check how the pins look.

colour_proposal

Left to right: Untouched, Hovered, Pin Marker, Selected

This colour scheme accounts for most kinds of colour blindness. While red is used twice, the pin shape helps with sorting this out.

I am still somewhat unhappy about how it looks with simulated protanomaly, which affects about 1% of men. However, it is still a huge improvement over the previous colour scheme.

Should this not suffice, we should think about patterns/hatches in the long run.


dopenguin (approved):
🚀 @warm-coolguy

PR #51, opened 2023-12-12
add source maps to builds

warm-coolguy:
## Summary

Client builds now contain source maps.

This seems helpful since, if made available, it eases debugging productive issues. Also, we have no code secrets since we're OSS. Do you agree with this decision?

Try building to check.


dopenguin (approved):
🚀 @warm-coolguy

PR #56, opened 2023-12-13
Update icons of layerChooser to be clearer for the content behind it

dopenguin:
## Summary

After some feedback it became clear, that the previously used atlas icon needed to be replaced with a different icon.  
The decision fell on `layer-group` as it most closely resembles the usage one sees in e.g. Google Maps.

As all clients have been updated, all clients should be run and checked.


warm-coolguy (approved):
🚀 @dopenguin

PR #58, opened 2023-12-13
Add the possibility to update the icon of the close-button of the MoveHandle

dopenguin:
## Summary

Add the possibility to update the icon of the close-button of the MoveHandle and update the value for Meldemichel.

- Meldemichel: Check the view of a single feature with its new closeButton whether it is clear enough that it leads back to the FeatureList.
- Check that no changes to the close button have been made.


packages/clients/meldemichel/src/plugins/Gfi/Feature.vue
```diff
@@ -9,7 +9,7 @@
         :aria-label="$t('common:plugins.gfi.header.close')"
         @click="close"
       >
-        <v-icon small>fa-xmark</v-icon>
+        <v-icon small>fa-angles-right</v-icon>
```
warm-coolguy, line 12: >Meldemichel: Check the view of a single feature with its new closeButton whether it is clear enough that it leads back to the FeatureList.

On a random civilian surprise test, this icon was not interpreted as "back", but as "close". I assume that it's difficult to expect a feature list when the feature list is not known and the visible GFI contents have not yet been mentally linked to the IconMenu.

I still deem this an acceptable result, even though a little learning is involved. It's way better than having the X icon for both close and "back" (although it's not *strictly* back when this was the first view) functionality.

A suggestion was made that both the XMARK and ANGLES-RIGHT buttons may be present. In contrast, it's clear that ANGLES-RIGHT must be something different than close, and closing the MoveHandle is one step easier.

Whether you see this as a fitting solution and an effort in this MR or a separate discussion shall be up to you. 🎩

warm-coolguy (changes requested):
🏓 @dopenguin

packages/clients/meldemichel/src/plugins/Gfi/Feature.vue
dopenguin: I'd say we can talk about this again if the need arises. Otherwise, this solution should suffice *enough*.

warm-coolguy (approved):
🚀 @dopenguin

PR #59, opened 2023-12-13
Set minHeight back to 0.2 to have content changes always be visible

dopenguin:
## Summary

Set minHeight back to 0.2 to have content changes always be visible.

Check whether the height is sufficient to have all different kinds of content changes be visible on narrow devices.


packages/components/MoveHandle.vue
warm-coolguy, line 1: >Check whether the height is sufficient to have all different kinds of content changes be visible on narrow devices.

This criterion is not met. I can easily have invisible changes in landscape mode. However, I do not think that we should force portrait mode or increase the MoveHandle minHeight further. It is always possible to have another tool that has its changes out of this 0.2 region.

Have you considered using [v-ripple](https://v2.vuetifyjs.com/en/directives/ripple/#usage) as an indicator on the element? There seems to be no way intended to actually trigger the effect without [some hacking](https://stackoverflow.com/questions/56824904/how-can-i-programmatically-trigger-ripple-effect-on-a-vuetify-js-component) on content changes (to be detected by the MoveHandle), and then we'd have to check if the focus is moved – or do we even want to move the focus? Maybe it doesn't matter as much since we're talking touch devices anyway.

Should we use this effect, we can go back to a minHeight of 0.1 to allow more flexibility for the user.

warm-coolguy (changes requested):

warm-coolguy (changes requested):
🏓 @dopenguin

packages/components/MoveHandle.vue
dopenguin: Now, instead of changing the `minHeight`, the MoveHandle jiggles a little bit if it is opened / the content is changed 515e822d

warm-coolguy: I liked this solution in theory, but absolutely dislike seeing it. It feels old, when the update information is already visible/recognizable it's confusing/annoying (by temporarily reducing readability and view calmness), and I somehow expect to see Clippy pop up and offer me to help with making a report.

I've quickly mocked how ripple may look to get a feeling for whether I'd have issues with this too

```
const blinker = this.$el.children[0]
blinker.style.transition = 'background-color ease 0.125s'
blinker.style['background-color'] = 'rgba(180, 180, 180, 1.0)'
setTimeout(() => (blinker.style['background-color'] = ''), 125)
```

and prefer it by a lot.

I suggest we go with the ripple effect, either by making use of the vuetify v-ripple directive (if possible here) or by writing our own small ripple effect. It's how Material Design indicates interaction, after all, and this is the direct result of an interaction and "part of the click" in my mind.

packages/components/MoveHandle.vue
```diff
@@ -126,9 +126,19 @@ export default Vue.extend({
       this.$root.$el.clientHeight,
       this.maxHeight
     )}px`
+    handleElement.style.transition = 'margin-top ease 0.25s'
```
warm-coolguy, line 129: This should be solved in the style tag as it's non-temporary.

warm-coolguy (changes requested):

warm-coolguy (approved):
🚀 @dopenguin

PR #64, opened 2023-12-18
Fix/#57 remove useless plugin attributes

raschju:
## Summary

Attribute windowLength does not exist and therefore has been removed.


packages/clients/afm/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## 1.0.1
 
 - Fix: The included example files have been updated to the new syntax and work again.
+- Fix: Remove configurable attribute windowLength which does not exist anymore.
```
dopenguin, line 6: Changelog entry is not necessary, as the respective customer does not recognize that parameter.

packages/clients/dish/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 1.0.1
+
+- Fix: Remove configurable attribute windowLength which does not exist anymore.
+
```
dopenguin, line 6: Changelog entry is not necessary, as the respective customer does not recognize that parameter.

dopenguin (changes requested):
🏓 @raschju @KlampMe

packages/clients/dish/CHANGELOG.md
KlampMe: thanks for the note 🏓

dopenguin (approved):
🚀 @KlampMe @raschju

Issue #72, opened 2024-01-03
GFI window navigation


warm-coolguy:
In the GFI window shown in e.g. the Meldemichel client, the arrows clickable to browse a multi-marker appear too small and grey.

The elements are supposed to be in the client's primary color, and to have the font-size of 24px, analoguous to how the close button appears.

PR #74, opened 2024-01-04
fix gfi window navigation style to match neighbouring elements

warm-coolguy:
## Summary

In mobile mode, navigational arrows now appear in the same fashion as their neighbouring elements do.

Visual check on mobile/desktop.

Resolves: #72


dopenguin (approved):
🚀 @warm-coolguy

PR #76, opened 2024-01-04
improve focus highlighting on WFS List

### LINKED ISSUE #67: Remove default vuetify focus & hover style on WFS List items in @polar/plugin-gfi
`vuetify` adds a `:before`-pseudo-element changing the colour of the background to `black` with an opacity of `0.12`.  
As the `List.vue` adds its own background-colour, possibly on multiple elements, this leads to a weird visual effect.

Task of this issue is that only the background-colour defined in `List.vue` should be visible on the focused `<v-list-item>`.

wfs_list

### LINKED ISSUE #68: Only set border on currently hovered / focused v-list-item and not related ones in WFS List of @polar/gfi
If a feature is clustered, all the related features are highlighted in the WFS-List as well. This means, that all of them have a background-colour of `#dff0dd` and a border of `inset 2px #3fa535`.

However, the user might not understand which feature he is currently selecting if he is navigating the list via keyboard or looses track of his mouse.

Task of this issue is to only set the border on the actively focused / hovered feature; the background-colour should be left as is.

warm-coolguy:
## Summary

The focused feature is now highlighted with a solid border, the hovered with a dashed border, and map/neighbour focusing is indicated with a greenish background to the features.

This way, all states of highlighting should be easier to decipher for end users.

Visual check including mixed usage with keyboard and mouse on list and map.

Resolves #67
Resolves #68

### DIFF
--- a/packages/plugins/Gfi/CHANGELOG.md
+++ b/packages/plugins/Gfi/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## unpublished
 
+- Feature: Improve WFS list highlighting with focus/hover styles that are easier to decipher for end users.
 - Feature: Add the possibility to update the close-button to e.g. indicate movement to the vector layer feature list.
 
 ## 1.1.0
--- a/packages/plugins/Gfi/src/components/List.vue
+++ b/packages/plugins/Gfi/src/components/List.vue
@@ -180,12 +180,23 @@ export default Vue.extend({
 }
 
 .v-list-item {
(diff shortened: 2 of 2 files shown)


dopenguin (changes requested):
A changelog entry is missing.

Otherwise, this looks perfect!

🏓 @warm-coolguy

dopenguin (approved):
🚀 @warm-coolguy

Issue #35, opened 2023-11-24
Document required plugin order in `addPlugins`


warm-coolguy:
Plugins do sometimes, by configuration, rely upon other plugins. The plugins relied upon require to be initialized before the relying plugins, which means *being put in the `addPlugins` method first*.

More technically: If a plugin relied upon by configuration is not ready at setup time, the relying plugin's `setupPlugin`-Action may register itself to changes on unknowns, and will never receive updates, as if the relations were not configured. *This is only relevant to `setupPlugin`*. Other relation configurations without listeners will work fine.

This feels off. We may write a centralized lib function that asynchronously tries again to register itself for `setupPlugin`-Actions; the target should be available after the call stack is through, or on, metaphorically, `$nextTick`. 😄 Should that not work out quickly, we should at least document where plugin relations lead to initialization order requirements.

PR #78, opened 2024-01-05
document plugin order requirements

warm-coolguy:
## Summary

It has been documented which plugin fields add initialization order requirements. While this isn't the optimal solution, I don't think we need an enhanced mechanism for this as of now, and should address this more deeply if the requirement arises.

Resolves: #35


dopenguin (approved):
🚀 @warm-coolguy

Issue #54, opened 2023-12-13
Hatchable markers


warm-coolguy:
Since markers support multiple states (normal, hovered, selected) and pins look a little like markers, additional distinguishing criterions would be helpful. Currently, we use colours. However, it is difficult to have four different colours that are distinguishable despite all existing kinds of colour blindness.

Enter [masterportalAPI hatching](https://bitbucket.org/geowerkstatt-hamburg/masterportalapi/src/master/src/vectorStyle/styles/polygon/polygonStyleHatch.js).

We could use these methods to make the available hatch styles available to configuration. The challenge is to make these work in a plausible manner in the rather small markers.

PR #81, opened 2024-01-08
implement hatchable markers

warm-coolguy:
## Summary

Map markers can now be styled with masterportalapi hatching features.

Test by setting in packages/clients/meldemichel/src/mapConfiguration.ts as fill in any extenderMasterportalapiMarkers style e.g. the following:

```js
          {
            pattern: 'diagonal-right',
            size: 12,
            lineWidth: 4,
            backgroundColor: [255, 255, 255, 1],
            patternColor: [225, 0, 25, 1],
          }
```

This will slow down the Meldemichel client a lot when used as `defaultStyle` due to the abundance of markers. However, it's very usable as hover or selection style.

It is currently not actually used since the color differantiation currently suffices.

Resolves: #54


packages/core/src/utils/markers/index.ts
```diff
@@ -26,13 +27,28 @@ const makeMarker = ({
   fill = defaultFill,
 }: MarkerStyle) =>
   `${prefix}${encodeURIComponent(`
-<svg width="${imgSize[0]}" height="${imgSize[1]}" viewBox="0 0 30 43" xmlns="http://www.w3.org/2000/svg">
+<svg width="${imgSize[0]}" height="${
+    imgSize[1]
+  }" viewBox="0 0 30 43" xmlns="http://www.w3.org/2000/svg">
```
dopenguin, line 32: I can't seem to see the difference here on why the linter previously didn't add a linebreak.  
Nothing has changed, right? 🤨

packages/core/src/utils/markers/index.ts
```diff
@@ -26,13 +27,28 @@ const makeMarker = ({
...
+      ? ''
+      : `<defs>
+          <pattern id="img" patternUnits="userSpaceOnUse" x="0" y="0" width="${
+            fill.size || 30
+          }" height="${fill.size || 30}">
+            <image href="${getPolygonFillHatch(fill).toDataURL()}" width="${
+          fill.size || 30
+        }" height="${fill.size || 30}"/>
+          </pattern>
+        </defs>`
```
dopenguin, line 45: This code segment is added here twice.  
I suggest adding a function that receives `fill` as a parameter and returns this segment

packages/core/src/utils/markers/hatches.ts
dopenguin, line 1: I broadly scanned this file but as it is just a simple copy, for the most part, I won't look at it in-depth 🎩

dopenguin (changes requested):
I really like this feature! Looking forward to actually see it in use :D  

Made some small comments, looks good overall! 🏓 @warm-coolguy

packages/core/src/utils/markers/index.ts
warm-coolguy: Yeah, that's weird. Maybe I changed something forth and back and this was the result that's *also* fine. I remember that I fiddled a little with height and width. But I'm not sure.

packages/core/src/utils/markers/index.ts
warm-coolguy: Good catch.

https://github.com/Dataport/polar/pull/81/commits/c7b6a7c31c4952c6d325854765758674f1de07b3

dopenguin (approved):
🚀 @warm-coolguy

Issue #52, opened 2023-12-13
Mobile hover effects


warm-coolguy:
**Situation**

Hovering on the WFS list and pins is not intended on mobile devices, since mobile devices don't *really* support hovering. However, it is *possible* when panning the map with a finger on a pin. It is now up to debate whether to also support "hovering" on the WFS list in the same fashion, which is with touch interactions.

If we plan to support it, the following thread has useful information and links.

https://github.com/Dataport/polar/pull/43#discussion_r1424340939

Beside the above thread, the tooltip displayed for the feature should be removed for mobile, too.

**Resolution**

- Do we want to support "hovering" on mobile devices? ~~Yes~~ / No
- Do we need to remove "hovering" on pins on mobile devices? Yes / ~~No~~

dopenguin:
> [#82 (comment)](https://github.com/Dataport/polar/pull/82#issuecomment-1882537105) Comment open for discussion here. @dopenguin

I'd suggest that we re-evaluate the importance of removing the hovering for mobile devices.  
If it seems important enough, we should make an estimation on how long it would take someone to implement this. If the time needed exceeds the estimation, this would then simply become WONTFIX.

dopenguin:
> > [#82 (comment)](https://github.com/Dataport/polar/pull/82#issuecomment-1882537105) Comment open for discussion here. @dopenguin
> 
> I'd suggest that we re-evaluate the importance of removing the hovering for mobile devices. If it seems important enough, we should make an estimation on how long it would take someone to implement this. If the time needed exceeds the estimation, this would then simply become WONTFIX. @warm-coolguy

After short discussion, we came to the solution to keep this behaviour for now.

PR #82, opened 2024-01-08
remove tooltips on touch and pen events

warm-coolguy:
## Summary

See commit history.

Mind that the touch/pen interactions on the user's GeoLocation has been left in intentionally. The tooltip is, in that situation, not redundant information.

Testable in Meldemichel client.

Relates: #52


dopenguin (changes requested):
This pull request resolves only one issue that is described in #52, namely the visible tooltips.  
It is still possible to re-enact a hovering on a feature in the map or the list and having a style change in both.

The following screenshots are from the hovering on the feature in the map. Hovering the list element looks slightly different but is not desired as well.

iPad Pro with Safari:
hovering

OnePlus 9 Pro with Chrome:
hovering_small

🏓 @warm-coolguy

warm-coolguy:
> It is still possible to re-enact a hovering on a feature in the map or the list and having a style change in both.

Ah, missed that. I've tried and researched, but there seems no easy way to distinguish what kind of device produced the mouseover event. All solutions use a `isMobile` flag that we don't have. I'd rather not touch this as to not break anything, since this hovering effect is not as obstructive/confusing as the tooltip is.

I propose we merge this and file away the rest as WONTFIX for now.

Issue #33, opened 2023-11-23
Refactor overly long functions and files


warm-coolguy:
**Describe the bug**
On linting, various warnings appear.

**Expected behavior**
On linting, no warnings appear.

**Steps to reproduce**
Lint.

**Screenshots**
E.g. visible here: https://github.com/Dataport/polar/actions/runs/6965878816

**Desktop (please complete the following information):**
Node@18.13.0

**Smartphone (please complete the following information):**
Never linted on a smartphone, but it probably still happens.

**Additional context**
Not every function or file may be fit for refactoring; e.g. type files *may* be eligible for lint ignore. @dopenguin What do you think? Shall we drop a few eslint-disable-file statements?

warm-coolguy:
The `packages/core/src/utils/createMap.ts` has been refactored in https://github.com/Dataport/polar/pull/61 out of necessity since I couldn't read/fix it in its current form. Just saying so you don't try to refactor it, too.

PR #84, opened 2024-01-08
refactor MapContainer.vue

warm-coolguy:
## Summary

There was a method that more closely resembled an action since it was based both on store information and made commits to it. The method has been moved to the appropriate position.

Behaviour should be unchanged. Best testable with DISH client with/-out Cors Enabler.

Relates: #33


dopenguin (approved):
🚀 @warm-coolguy

Issue #80, opened 2024-01-05
Refactor store/createInteractions of @polar/plugin-draw to be a proper action


dopenguin:
The function `createInteractions` is written like a util function but uses the same context object as its first parameter like an action.

As described in an [ADR](https://github.com/Dataport/polar/wiki/Architecture-Decision-Records-%F0%9F%93%83#2024-01-02-difference-between-actions-utils-and-lib-packages), this function should be an action.

To be an action, it 

- should have `drawSource` and `drawLayer` in an object as the second function parameter
- shouldn't be called directly but via `dispatch`
- should retrieve `textSize` from `getters`

PR #86, opened 2024-01-09
Refactor/#80 draw createinteractions

dopenguin:
## Summary

Refactored `createinteractions` of `@polar/plugin-draw` to be an actual action and split it in its separate parts for readability.

Run `snowbox` and test whether the draw plugin still works as intended.

Resolves: #80 
Relates: #33


warm-coolguy (changes requested):
I am confused about what is and what isn't an action, especially regarding createTextInteractions being an action while createDeleteInteractions isn't one.

Following the ADR, I'd go with the third option:

>Does the functionality not have state changes, but belongs to a certain action? Either locally in the same file as the action or in a folder named after the action in the path store/ACTIONNAME

Since you want to split, a folder for `createInteractions` seems to be required. That also makes sense from the point where all of the methods are called only there, and where they don't make state changes, except for createModifyInteractions,

1. whiches initial state change may be moved to the caller to avoid this
2. whiches other commits are indirect since they just empower the select and don't change state directly, that is, it's not *really* an action in itself

That all methods work rather via return than state change is another hint in this direction.

I think changes are in order to follow the ADR, but maybe you have a different interpretation?

dopenguin:
> I am confused about what is and what isn't an action, especially regarding createTextInteractions being an action while createDeleteInteractions isn't one.
> 
> Following the ADR, I'd go with the third option:
> 
> > Does the functionality not have state changes, but belongs to a certain action? Either locally in the same file as the action or in a folder named after the action in the path store/ACTIONNAME
> 
> Since you want to split, a folder for `createInteractions` seems to be required. That also makes sense from the point where all of the methods are called only there, and where they don't make state changes, except for createModifyInteractions,
> 
>     1. whiches initial state change may be moved to the caller to avoid this
> 
>     2. whiches other commits are indirect since they just empower the select and don't change state directly, that is, it's not _really_ an action in itself
> 
> 
> That all methods work rather via return than state change is another hint in this direction.
> 
> I think changes are in order to follow the ADR, but maybe you have a different interpretation?
> 
> 🏓 @dopenguin

I moved all the functions to its own folder (537f188) and changed `createTextInteractions` to be a simple function instead of an action (94c8e74).  
However, I lean toward keeping `createModifyInteractions` as an action. Agree?

🏓 @warm-coolguy

warm-coolguy:
> > I am confused about what is and what isn't an action, especially regarding createTextInteractions being an action while createDeleteInteractions isn't one.
> > Following the ADR, I'd go with the third option:
> > > Does the functionality not have state changes, but belongs to a certain action? Either locally in the same file as the action or in a folder named after the action in the path store/ACTIONNAME
> > 
> > 
> > Since you want to split, a folder for `createInteractions` seems to be required. That also makes sense from the point where all of the methods are called only there, and where they don't make state changes, except for createModifyInteractions,
> > ```
> > 1. whiches initial state change may be moved to the caller to avoid this
> > 
> > 2. whiches other commits are indirect since they just empower the select and don't change state directly, that is, it's not _really_ an action in itself
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > That all methods work rather via return than state change is another hint in this direction.
> > I think changes are in order to follow the ADR, but maybe you have a different interpretation?
> > 🏓 @dopenguin
> 
> I moved all the functions to its own folder ([537f188](https://github.com/Dataport/polar/commit/537f1884a97cca6fea30cf4abfefc741d54cda47)) and changed `createTextInteractions` to be a simple function instead of an action ([94c8e74](https://github.com/Dataport/polar/commit/94c8e7480ada640cbaa56d5a0b912117294376db)). However, I lean toward keeping `createModifyInteractions` as an action. Agree?
> 
> 🏓 @warm-coolguy

It's a little confusing, but so is forwarding the bound commit method by parameter.

🚀 @dopenguin

warm-coolguy (approved):

PR #87, opened 2024-01-10
add missing aria label

warm-coolguy:
## Summary

An aria label was missing. It's there now.

Check with NVDA, if need be.


dopenguin (approved):
🚀 @warm-coolguy

PR #90, opened 2024-01-15
remove wildcard on street search

warm-coolguy:
## Summary

See added file's `scripts/overrideMasterportalapi.js` for details. While this is a global change, I decided to document it in the Meldemichel, since it's the only client using this right now *and* the change results in a new Meldemichel beta release.

To save you the search, https://bitbucket.org/geowerkstatt-hamburg/masterportalapi/src/master/src/searchAddress/searchGazetteer.js is the file's current state. In case of masterportalapi update, we can do this properly.

Install (to trigger postinstall) and run the `client:meldemichel:dev`. Should the changes not arrive, it may be required to add `--force` to the vite command in the root package.json once to avoid vite's cache, which plausibly does not detect manual module changes.

An interesting search is for "Roonstraße". This produced plausible results in the previous Meldemichel (and on current /prod), but produces weird results on the /stage system ("Liliencronstraße").

### DIFF
--- a/scripts/overrideMasterportalapi.js
+++ b/scripts/overrideMasterportalapi.js
@@ -1,0 +1,38 @@
+/* eslint-disable no-console */
+/* eslint-disable @typescript-eslint/no-var-requires */
+
+/*
+ * NOTE
+ *
+ * In current masterportalapi versions, this behaviour can be configured.
+ * Since it currently can't, we'll deactive the wildcard on street names. It
+ * may lead to notoriously bad results on an address search without further
+ * filtering, e.g. "Roonstraße" will first result in the "Liliencronstraße",
+ * which is about as confusing as it gets.
+ *
+ * On an update to the masterportalapi, we may make use of this by first
+ * running on the method without wildcard and, if no results came in, with
+ * wildcard.
+ */
+
+const fs = require('fs')
+
+const filePath =
+  './node_modules/@masterportal/masterportalapi/src/searchAddress/searchGazetteer.js'
+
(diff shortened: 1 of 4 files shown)


scripts/overrideMasterportalapi.js
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * In current masterportalapi versions, this behaviour can be configured.
+ * Since it currently can't, we'll deactive the wildcard on street names. It
+ * may lead to notoriously bad results on an address search without further
+ * filtering, e.g. "Roonstraße" will first result in the "Liliencronstraße",
+ * which is about as confusing as it gets.
+ *
+ * On an update to the masterportalapi, we may make use of this by first
+ * running on the method without wildcard and, if no results came in, with
+ * wildcard.
```
dopenguin, line 15: This part should be added to the `arcana.md` instead of describing it here.

dopenguin (changes requested):
Just one small thing, works like a charm otherwise!

Great catch! 🏓 @warm-coolguy

scripts/overrideMasterportalapi.js
warm-coolguy: This particular line is more of a development ticket that's currently not written out. One would automatically find it via the `arcana.md` by searching for the package as described. However, an additional note does not hurt: https://github.com/Dataport/polar/pull/90/commits/76129bb0e003ab407328d0d49ba9da6d14797194

dopenguin (approved):
🪨 @warm-coolguy

PR #95, opened 2024-01-17
Feature/current i18n package

dopenguin:
## Summary

Upgrade to the currently maintained i18next package for vue, upgrade i18next and subsequently typescript to the latest major version.

Usage of `@panter/vue-i18next` was dropped as it was no longer being maintained and recommended `i18next-vue`.

Run the snowbox (or other clients) and check whether the translations are still working.

Blocked by: #93


dopenguin:
I accidentally also added the upgrade of `vite` to the next major version here, woops!

warm-coolguy (approved):

PR #96, opened 2024-01-25
fix markers being lost on reclicks

warm-coolguy:
## Summary

Markers were lost due to technically required closing operations that were mixed with user-intended closing operations. However, pins should only disappear when users intend the closing operation. When re-clicking now, the pin no longer disappears, since that was not the users intention.

Run DISH/Meldemichel/AfM and check whether behaviour is now as expected. Previously, pins were lost after moving them on or clicking on a second feature. This was an unintended side effect of Meldemichel development and only happens when useExtendedMasterportalapiMarkers is off.


dopenguin (approved):
🚀 @warm-coolguy

PR #97, opened 2024-01-25
fix pin colour in DISH

warm-coolguy:
## Summary

Somehow the DISH configuration for pin colours was named wrongly. This has been resolved.

Run the DISH client and check whether the correct tint of blue is displayed. For testing, you may switch the shBlue in DISH's config to any clearly distinguishable colour.


dopenguin (approved):
🚀 @warm-coolguy

PR #98, opened 2024-01-30
roll back incomplete attribution changes

warm-coolguy:
## Summary

Somehow we ended up with `{{YEAR}}` instead of `<YEAR>` in the Attributions code. However, that is a breaking change, and to my knowledge no configurations exist where `{{YEAR}}` is used. Instead, the text just disappears due to our i18n package, and hence this breaking change wasn't even noticed yet.

I've just rolled it back for compatibility with existing configurations. Do you know anything more about this?

Check e.g. DISH client.


dopenguin (approved):
🚀 @warm-coolguy 

I've got no knowledge how this came to be. As the prior implementation utilizing `{{YEAR}}` was already in place when the repository moved to GitHub, we'd have to check in the previous instance, if need be.

PR #99, opened 2024-01-30
improve robustness on misconfiguration

warm-coolguy:
## Summary

Improve robustness for layer ids that are configured, but do not exist on the service registry.

Change any snowbox layer id to an id that does not exist and review client behaviour.

### DIFF
--- a/packages/plugins/LayerChooser/CHANGELOG.md
+++ b/packages/plugins/LayerChooser/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## unpublished
+
+- Fix: The LayerChooser plugin will keep working on the error that a layer without entry in the service register has been configured.
+
 ## 1.2.0
 
 - Feature: Improved implementation to make plugin SPA-ready.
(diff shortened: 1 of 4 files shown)


dopenguin (commented):
Cool change!

However, I'm quite sure that a user might get confused why a layer entry in the layer chooser does nothing in the case that the id is not present in the register; the user probably won't read the console.

I'd suggest that the layer also shouldn't be added to the menu. Do you agree?

warm-coolguy:
> Cool change!
> 
> However, I'm quite sure that a user might get confused why a layer entry in the layer chooser does nothing in the case that the id is not present in the register; the user probably won't read the console.
> 
> I'd suggest that the layer also shouldn't be added to the menu. Do you agree?

Well, it's slightly better, albeit it may also produce broken clients and further obfuscate what exactly is broken. But it may increase chances that everything'll keep working, so ...

https://github.com/Dataport/polar/pull/99/commits/99bd162d5ec49a53b900514ae3a0d267bb73a126

I've deleted an outdated warning while we're at it.

dopenguin (approved):
🚀 @warm-coolguy

PR #101, opened 2024-01-30
narrow down type for EPSG codes

warm-coolguy:
## Summary

I've narrowed down the type for EPSG codes.

It's not *really* a fix, but it surely isn't breaking (non-matching EPSG codes didn't work anyway) or a feature, so I've gone with `fix` since it's safely installable and other strings shouldn't have been allowed in the first place.


dopenguin (approved):
🚀 @warm-coolguy

Issue #104, opened 2024-02-09
<noscript> advisory


warm-coolguy:
Since we can't do anything in browsers that have JS disabled, we should at least advise our users to use `<noscript>` tags in the POLAR divs. For this, these steps should be taken:

1. Test if something funny happens if our target div contains a `<noscript>` tag (just to be safe ...)
2. Update all API.md files to have a `<noscript>` tag in the POLAR rendering div with matching text and a comment explaining this is optional and up to our direct consumers

PR #105, opened 2024-02-09
add noscript advisory

warm-coolguy:
## Summary

Add advisory to use noscript tag in POLAR's render div, should no other place on the page take care of this. No matter now noscript-compatible a page is, we are not.

Test whether `<noscript>` tag (added where possible) interfered with workflow.

I've skipped changelogs since no substantial changes have been made.

Resolves: #104


dopenguin (approved):
🚀 @warm-coolguy

PR #109, opened 2024-02-23
fix type to allow custom strings for custom methods

warm-coolguy:
## Summary

SearchType was previously hard-wired to pre-defined search types. However, since any client may add custom search methods, any string can be allowed as search type.

This allows `string` as search type for clients to use their own defined search methods without errors messages.


dopenguin (approved):
🚀 @warm-coolguy

PR #114, opened 2024-02-26
add optional i18n for feature titles

warm-coolguy:
## Summary

Feature titles may now optionally hold locale keys. If the title is not found within the locales, it will be shown as-is.

No test environment given, will be used in the TextLocator.


dopenguin (approved):
🚀 @warm-coolguy 

Haven't tested, but the translate function shouldn't mess with anything.

PR #116, opened 2024-02-28
add vector source getter to draw plugin

warm-coolguy:
## Summary

The vector source can now be fetched via getter, circumventing the previous necessity to dig it out from the map. The CHANGELOG was intentionally left unchanged as this is internal API only.


dopenguin (approved):
🚀 @warm-coolguy 

Answered the question myself

PR #117, opened 2024-03-04
Add AddressSearch and ReverseGeocoder to 'SINGLE' mode of meldemichel

dopenguin:
## Summary

It was requested that a feature can be reverse geocoded and the address search can be used programmatically in mode `SINGLE`.  
An addition to the API.md was redundant, as the programmatic usage has already been described in the README of `@polar/plugin-address-search`.

- Add the `mapInstance` to the `window` object in `packages/clients/meldemichel/example/single.html`.
- Use the following code snippet to trigger the search.

```js
window.$mapInstance.$store.dispatch('plugin/addressSearch/search', {
  input: 'Spitalerstraße 12',
  autoselect: 'first'
})
```

A pin should be set to `Spitalerstraße 12` and the following fields should have a value

- `vendor_maps_address_str`
- `vendor_maps_address_hnr`
- `vendor_maps_address_plz`
- `vendor_maps_distance_to`

### DIFF
--- a/packages/clients/meldemichel/src/mapConfigurations.ts
+++ b/packages/clients/meldemichel/src/mapConfigurations.ts
@@ -248,11 +248,13 @@ const mapConfigurations: Record<
   }),
   [MODE.SINGLE]: () => ({
     ...commonMapConfiguration,
+    addressSearch: { ...addressSearch, displayComponent: false },
     layers: commonLayers,
     attributions: {
       ...commonAttributions,
     },
     pins: commonPins,
+    reverseGeocoder,
   }),
 }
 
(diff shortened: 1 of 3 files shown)


packages/clients/meldemichel/src/mapConfigurations.ts
```diff
@@ -248,11 +248,13 @@ const mapConfigurations: Record<
   }),
   [MODE.SINGLE]: () => ({
     ...commonMapConfiguration,
+    addressSearch: { ...addressSearch, displayComponent: false },
```
warm-coolguy, line 251: I found this to be confusing with the `displayComponent: true` above and would favour the statement in line 30 to be `mode !== MODE.SINGLE` so we later have less positions to consider for behaviour understanding/change.

warm-coolguy (changes requested):

warm-coolguy (approved):
🚀 @dopenguin

PR #118, opened 2024-03-14
fix marker draggability after map size change

warm-coolguy:
## Summary

Oddly, the map would pan with the marker drag after switching to fullscreen mode, effectively rendering the marker almost immovable. This has been resolved by updating the Translate interaction on map size changes.

Since the file was of >300 lines size and the setup method was too long, I've refactored it while touching it. The actual fix is merely these lines:

```
        // without update, map will pan during drag
        this.watch(
          () => rootGetters.hasSmallWidth || rootGetters.hasSmallHeight,
          () => dispatch('updateMarkerDraggability')
        )
```

`npm run meldemichel:dev` with [index.zip](https://github.com/Dataport/polar/files/14601162/index.zip) used instead of the usual `index.html` in `meldemichel/src` to test the affected mode. Also testing the Pins plugin in other situations to make sure no unwanted side effects occur.

### DIFF
--- a/packages/plugins/Pins/CHANGELOG.md
+++ b/packages/plugins/Pins/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 1.3.1
+
+- Fix: The map dragged along with the pin in some situations, rendering the pin effectively immovable. This has been fixed.
+
 ## 1.3.0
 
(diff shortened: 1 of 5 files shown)


dopenguin (approved):
🚀 @warm-coolguy

PR #119, opened 2024-04-09
add gh-page link and logos

warm-coolguy:
## Summary

A link and some logos have been added to the GitHub page.

Check whether the result is pleasently looking locally.


dopenguin (approved):
🚀 @warm-coolguy

PR #121, opened 2024-04-15
Fix/plugin geo location no zoom outside extent

dopenguin:
## Summary

When using `@polar/plugin-geo-location`, the map tried to center on the users position even if they'd be outside of the maps extent. This behaviour has been fixed.

Note that the implementation could be combined with `@polar/lib-passes-boundary-check`. However, I didn't move forward with that as that would mean a breaking change in that package to optionalize the parameter `boundaryLayerId`.

Change the following parameter in `@polar/client-dish` and run the dev-Script

```json
startResolution: 10.583327618336419,
startCenter: [553655.72, 6004479.25],
extent: [
  562031.2360463801, 5993415.936589229, 578216.7556916607, 5993517.139659579,
],
```

### DIFF
--- a/packages/clients/dish/CHANGELOG.md
+++ b/packages/clients/dish/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## unpublished
+
+- Feature: Add `toastAction` to `geoLocation` configuration.
+
 ## 1.1.1
 
 - Fix: The marker previously disappeared on being moved/reclicked on a second feature. This issue has been resolved.
--- a/packages/plugins/GeoLocation/CHANGELOG.md
+++ b/packages/plugins/GeoLocation/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## unpublished
+
+- Fix: Only zoom to the user position if the user is currently in the extent of the map.
+
 ## 1.3.0
 
 - Feature: Improved implementation to make plugin SPA-ready.
--- a/packages/plugins/GeoLocation/src/store/actions.ts
+++ b/packages/plugins/GeoLocation/src/store/actions.ts
@@ -4,6 +4,7 @@ import VectorLayer from 'ol/layer/Vector'
 import Point from 'ol/geom/Point'
 import { Vector } from 'ol/source'
 import Feature from 'ol/Feature'
+import { containsCoordinate } from 'ol/extent'
 import { Style, Icon } from 'ol/style'
 import * as Proj from 'ol/proj.js'
 import Geolocation from 'ol/Geolocation.js'
@@ -130,11 +131,10 @@ const actions: PolarActionTree<GeoLocationState, GeoLocationGetters> = {
       configuredEpsg
     )
 
-    const boundaryCheckPassed = await passesBoundaryCheck(
-      map,
-      boundaryLayerId,
-      transformedCoords
-    )
+    const boundaryCheckPassed =
+      typeof boundaryLayerId === 'string'
+        ? await passesBoundaryCheck(map, boundaryLayerId, transformedCoords)
+        : containsCoordinate(map.getView().calculateExtent(), transformedCoords)
     const boundaryErrorOccurred = typeof boundaryCheckPassed === 'symbol'
 
     if (
--- a/packages/clients/dish/src/mapConfig.ts
+++ b/packages/clients/dish/src/mapConfig.ts
@@ -83,6 +83,7 @@ export const mapConfiguration = {
   },
   geoLocation: {
     checkLocationInitially: false,
+    toastAction: 'plugin/toast/addToast',
     zoomLevel: 7,
   },
   attributions: {


packages/plugins/GeoLocation/src/store/actions.ts
```diff
@@ -130,11 +131,10 @@ const actions: PolarActionTree<GeoLocationState, GeoLocationGetters> = {
...
 
-    const boundaryCheckPassed = await passesBoundaryCheck(
-      map,
-      boundaryLayerId,
-      transformedCoords
-    )
+    const boundaryCheckPassed =
+      typeof boundaryLayerId === 'string'
+        ? await passesBoundaryCheck(map, boundaryLayerId, transformedCoords)
+        : containsCoordinate(map.getView().calculateExtent(), transformedCoords)
```
warm-coolguy, line 137: According to the [OL docs](https://openlayers.org/en/latest/apidoc/module-ol_View-View.html#calculateExtent), this only addresses the currently visible map's view's extent, and not the map's whole extent, which does not meet the intended functionality change.

warm-coolguy (changes requested):
* The `plugins/GeoLocation/README.md` is missing an udpate regarding the `toastAction` field.

🏓 @dopenguin

packages/plugins/GeoLocation/src/store/actions.ts
dopenguin: Adjusted 489d2f020a531a3df206f0496db7799668b84937

dopenguin:
> * The `plugins/GeoLocation/README.md` is missing an udpate regarding the `toastAction` field.
> 
> 
> 🏓 @dopenguin

The configuration parameter has not been changed for the plugin. What update is missing?

🏓 @warm-coolguy

warm-coolguy (changes requested):
> > * The `plugins/GeoLocation/README.md` is missing an udpate regarding the `toastAction` field.
> > 
> > 🏓 @dopenguin
> 
> The configuration parameter has not been changed for the plugin. What update is missing?
> 
> 🏓 @warm-coolguy

The file in question reads for the parameter:

`If `boundaryLayerId` is set, and the user is not locatable within the boundary, this string will be used as action to send a toast information to the user. If no toast information is desired, leave this field undefined; for testing purposes, you can still find information in the console.`

This no longer holds, as `boundaryLayerId` is no longer a requirement for the plugin to use this string.

Also, I remember we talked about this task beforehand and preferred the GeoLocation icon to be set independent of whether it is within the (map) boundary; on zooming out, one's position would be indicated outside of the extent. This seems to be reflected by the PRs name, which indicates the `zooming` is removed rather than the `setting`. Albeit it's debatable which behaviour is correct.

🏓 @dopenguin

dopenguin:
> > > * The `plugins/GeoLocation/README.md` is missing an udpate regarding the `toastAction` field.
> > > 
> > > 🏓 @dopenguin
> > 
> > 
> > The configuration parameter has not been changed for the plugin. What update is missing?
> > 🏓 @warm-coolguy
> 
> The file in question reads for the parameter:
> 
> `If `boundaryLayerId` is set, and the user is not locatable within the boundary, this string will be used as action to send a toast information to the user. If no toast information is desired, leave this field undefined; for testing purposes, you can still find information in the console.`
> 
> This no longer holds, as `boundaryLayerId` is no longer a requirement for the plugin to use this string.
> 
> Also, I remember we talked about this task beforehand and preferred the GeoLocation icon to be set independent of whether it is within the (map) boundary; on zooming out, one's position would be indicated outside of the extent. This seems to be reflected by the PRs name, which indicates the `zooming` is removed rather than the `setting`. Albeit it's debatable which behaviour is correct.
> 
> 🏓 @dopenguin

Documentation adjusted 3fd45cecdfa31d1b0c05bab0840bff750dee5f3e

I can't seem to remember that discussion. Let's talk about it!

🏓 @warm-coolguy

packages/plugins/GeoLocation/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## unpublished
+
+- Fix: Only zoom to the user position if the user is currently in the extent of the map.
```
warm-coolguy, line 5:
```suggestion
- Fix: Only zoom to the user position if the user is currently in the extent of the map. The user position is no longer indicated when it's outside the map's extent.
```

If we stick to that decision, that would complete this information.

packages/clients/dish/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## unpublished
+
+- Feature: Add `toastAction` to `geoLocation` configuration.
```
warm-coolguy, line 5:
```suggestion
- Feature: If a user is geolocated outside of the map's extent, the client will inform the user of why geolocation did not take effect via a textbox.
```

This should be more readable to the main recipients of this changelog.

warm-coolguy (changes requested):
🏓 @dopenguin 

While we're waiting for the next-best meeting to discuss it, there's two additional things I found.

warm-coolguy (approved):
We've discussed this and figured not showing the icon of users outside of BBOX/BoundaryLayer is fitting and consistent with previous client behaviour.

🚀 @dopenguin

PR #127, opened 2024-04-19
Fix initiallyOpen not working in prod mode

dopenguin:
## Summary

The fix of #123 was not working in production mode. By adding a new state parameter to `@polar/core`, this issue is now resolved.  
While I was at it, I also moved the logic of the watcher to the store to have it in line with other plugins.

- `npm run meldemichel:dev`
- `npm run meldemichel:build && npm run meldemichel:build:serve`


packages/types/custom/core.ts
```diff
@@ -665,6 +666,7 @@ export interface CoreGetters {
   hovered: Feature | null
   errors: PolarError[]
   map: Map
+  mapHasDimensions: boolean | null
```
warm-coolguy, line 669:
```suggestion
  mapHasDimensions: CoreState['mapHasDimensions']
```

Since `extends` doesn't work without a pretty long `Omit` here, I suggest we use this kind of reference to avoid duplications.

packages/lib/testMountParameters/index.ts
```diff
@@ -61,6 +61,7 @@ export default (): MockParameters => {
       moveHandleActionButton: 0,
       plugin: {},
       language: '',
+      mapHasDimensions: null,
```
warm-coolguy, line 64: Why use `null` at all? It seems to duplicate `false`. There is only one value change of interest, `falsy => true`. It does not help to change `null` to `false` since that only means the first render attempt did not run through, which is what usually happens. Is there a case I'm missing?

warm-coolguy (changes requested):
🏓 @dopenguin

packages/types/custom/core.ts
dopenguin: This deviates from the current procedure, as other properties may also be the same as a State Parameter and as a Getter.

If you deem this necessary, we should go through all properties and add a reference to the state.

packages/lib/testMountParameters/index.ts
dopenguin: I have adjusted this to be always a boolean 6710452

dopenguin:
> 🏓 @dopenguin

🏓 @warm-coolguy

packages/types/custom/core.ts
warm-coolguy: I've checked how we do this otherwise, and in the draw plugin, we're actually using `extends` with `Omit` (and everywhere else, to avoid duplication, `extends`). I can't seem to write this as a regular suggestion as many of the affected lines have not been changed by this MR, but I suggest his:

```ts
export interface CoreGetters
  extends Omit<
    CoreState,
    | 'components'
    | 'center'
    | 'hovered'
    | 'map'
    | 'moveHandle'
    | 'moveHandleActionButton'
    | 'selected'
  > {
  // omitted from CoreState as actual getter type diverges
  components: PluginContainer[]
  hovered: Feature | null
  map: Map
  moveHandle: MoveHandleProperties
  moveHandleActionButton: MoveHandleActionButton
  selected: Feature | null

  // regular getters
  hasSmallHeight: boolean
  hasSmallWidth: boolean
  /** Whether the application currently has the same size as the visual viewport of the users browser */
  hasWindowSize: boolean
}
```

Deduplication and clearly marked type changes (i.e. vuex trickery). Do you like it?

warm-coolguy:
> > 🏓 @dopenguin
> 
> 🏓 @warm-coolguy

🏓 @dopenguin

packages/types/custom/core.ts
dopenguin: This seems to be a reasonable approach!

As this has nothing to do with the fix, do you agree that we should add this in another PR?

dopenguin:
> > > 🏓 @dopenguin
> > 
> > 
> > 🏓 @warm-coolguy
> 
> 🏓 @dopenguin

🏓 @warm-coolguy

warm-coolguy (approved):
> > > > 🏓 @dopenguin
> > > 
> > > 
> > > 🏓 @warm-coolguy
> > 
> > 🏓 @dopenguin
> 
> 🏓 @warm-coolguy 

:rocket: @dopenguin

PR #128, opened 2024-04-22
Add tests for the actions of @polar/plugin-geo-location

dopenguin:
## Summary

Create tests for most actions of `@polar/plugin-geo-location`.

- `npm run test`
- Also, check whether any other meaningful tests may be added


warm-coolguy (approved):
🚀 @dopenguin

PR #134, opened 2024-06-24
fix bugs in time frame filter

warm-coolguy:
## Summary

- Selecting a single day filtered features in a way that selected only the day after the day selected.
- Selecting a date range added another day at the end.

These issues have been resolved.

This fix can easily be tested in `npm run meldemichel:dev` since features denote their date in GFI and the filter is configured.


dopenguin (approved):
🚀 @warm-coolguy

PR #135, opened 2024-06-24
fix GeoLocation failing map init on Safari<16

warm-coolguy:
## Summary

Fix the GeoLocation plugin that accessed `navigator.permissions.query` without checking whether it's supported prior, resulting in a failure of the map initialization on Safari versions below 16 and some other browsers.

This can be tested in the Meldemichel on a device with Safari<16. Since the map will not request access to the geolocation feature initially, it will simply proceed to try and fetch the geolocation; this triggers the usual request and will lead to an error toast.


dopenguin (approved):
🚀 @warm-coolguy

PR #137, opened 2024-06-27
fix filter plugin inconsistencies

warm-coolguy:
## Summary

The filter is now applied initially and on layer loads to prevent inconsistent/confusing UI states. Previously, never-visible features were visible initially, and newly loaded features would not have been filtered correctly. (Albeit there's no known client that has the latter issue at the time being, since all currently known clients are using the filter on initially completely loaded WFS.)

The README has been clarified regarding this.

Check the behaviour in the Meldemichel. Initially, less features should be visible than are transported via network. (See current count in anliegen_extern.json, it's constantly changing.)

### DIFF
--- a/packages/plugins/Filter/CHANGELOG.md
+++ b/packages/plugins/Filter/CHANGELOG.md
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+## 1.1.2
+
+- Fix: Features with categories that are not listed in `knownValues` are never displayed now. Previously, they were initially visible, but disappeared once any filter was touched.
+- Fix: It was possible to have features visible that were loaded after the filter was applied and that would have been filtered out. This has been resolved.
+
 ## 1.1.1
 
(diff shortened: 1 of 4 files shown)


packages/plugins/Filter/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+## 1.1.2
```
dopenguin, line 3: Shouldn't this be in `unpublished` at first?

dopenguin (commented):
🏓 @warm-coolguy

packages/plugins/Filter/CHANGELOG.md
warm-coolguy: Yep, fixed that.

dopenguin (approved):
🚀 @warm-coolguy

PR #138, opened 2024-07-02
Feature/add zoom slider to zoom plugin

jedi-of-the-sea:
## Summary

Adds a zoom slider to the plugin "Zoom", if configured. The zoom slider will be added under the zoom buttons. If the zoom buttons are not rendered, neither will the zoom slider.

Add 
  ``
zoom: {
    addZoomSlider: true,
  },
`` 
to the config file of the client.


packages/plugins/Zoom/src/components/ZoomSlider.vue
```diff
@@ -0,0 +1,76 @@
...
+      },
+    },
+  },
+  methods: {
+    ...mapActions('plugin/zoom', ['setZoomLevel']),
+  },
+})
+</script>
+
+<style lang="scss">
```
dopenguin, line 47: Can some of the styles be scoped?

dopenguin (changes requested):
Great addition to the code base!  
Left some comments for you to address.

Two additional things:
- The [OpenLayers example](https://openlayers.org/en/latest/examples/zoomslider.html), which uses the basic OpenLayers control, transitions between the zoom levels with ease. Do you think that adding that functionality here is feasible? Or do you think that that behaviour is not easily implemented with the current setup of `@polar/plugin-zoom`
- @warm-coolguy I've added a screenshot below. Do you think that the contrast of the focus / hover style is enough for WCAG 2.1? I've tried testing with a plugin, but focus and hover styles don't really seem to be registered ._.

contrast

🏓 @jedi-of-the-sea

warm-coolguy:
>     * @warm-coolguy I've added a screenshot below. Do you think that the contrast of the focus / hover style is enough for WCAG 2.1? I've tried testing with a plugin, but focus and hover styles don't really seem to be registered ._.
> 
> 
> contrast

I'd say it's insufficient since we're working with a map as background, so some WMS may decide to use our exact blue (or other configured primary) as water colour. Since we've had the same issue with all buttons and went over it at length, I'd recommend we use the same solution pattern as we used then. As a sketch:

target_sketch

I'm not sure how the currently effective hover effect looks in combination with that. Maybe we need to suppress it or the whole thing might become a tad too fancy.

jedi-of-the-sea:
> * The [OpenLayers example](https://openlayers.org/en/latest/examples/zoomslider.html), which uses the basic OpenLayers control, transitions between the zoom levels with ease. Do you think that adding that functionality here is feasible? Or do you think that that behaviour is not easily implemented with the current setup of `@polar/plugin-zoom`

I tried the OpenLayers control first but did not succeed with implementing them within the plugin. Since the buttons where also implemented without OpenLayers I figured it would be a good idea to use the code already there.

packages/plugins/Zoom/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 - Refactor: Remove redundant prop `isHorizontal` from `Zoom` component. Logic is now placed in `@polar/core`.
 
+- Feature: Add the option to configure a zoom slider.
+
```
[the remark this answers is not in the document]
jedi-of-the-sea: I changed my mind about the rendering of the zoom slider depending on the rendering of zoom buttons and adjusted the code. Might be that some prefer the zoom slider and don't want the buttons? So I did not add the additional information here.

packages/plugins/Zoom/README.md
```diff
@@ -13,6 +13,7 @@ It can be configured as followed.
 |------------|------------------------------|-----------------------------------------------------------------------------------------------|
 | renderType | 'iconMenu' \| 'independent'? | Whether the zoom related buttons are being rendered independently or as part of the IconMenu. Defaults to `'independent'`. |
 | showMobile | boolean?                     | Whether the zoom related buttons should be displayed on smaller devices; defaults to `false`.   |
+| showZoomSlider | boolean?                     | Whether a zoom slider should be displayed; defaults to false.   |
```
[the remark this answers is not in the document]
jedi-of-the-sea: I changed my mind about the rendering of the zoom slider depending on the rendering of zoom buttons and adjusted the code. Might be that some prefer the zoom slider and don't want the buttons? So I did not add the additional information here.

packages/plugins/Zoom/src/store/index.ts
```diff
@@ -69,6 +69,10 @@ export const makeStoreModule = () => {
         typeof rootGetters.configuration?.zoom?.showMobile === 'boolean'
           ? rootGetters.configuration.zoom.showMobile
           : false,
+      addZoomSlider: (_, __, ___, rootGetters) =>
+        typeof rootGetters.configuration?.zoom?.addZoomSlider === 'boolean'
+          ? rootGetters.configuration.zoom.addZoomSlider
+          : false,
```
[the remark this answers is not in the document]
jedi-of-the-sea: I switched the names of the computed property and the configuration parameter - `showZoomSlider` is now the parameter, `addZoomSlider` the computed.

packages/plugins/Zoom/src/components/Zoom.vue
```diff
@@ -40,26 +40,47 @@
...
+      'minimumZoomLevel',
       'renderType',
       'showMobile',
+      'addZoomSlider',
     ]),
     showZoomButtons(): boolean {
       return this.hasSmallHeight ? this.showMobile : true
     },
+    showZoomSlider(): boolean {
+      return !this.hasSmallHeight && this.addZoomSlider
```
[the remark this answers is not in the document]
jedi-of-the-sea: I thought through that again and changed my mind (see comment above / new PR description). This might make more sense to users.

packages/plugins/Zoom/src/components/ZoomSlider.vue
jedi-of-the-sea: unfortunately no, but I added a specific class for the zoom slider so that this style will not be applied to other v-sliders in the app.

packages/plugins/Zoom/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # CHANGELOG
 
 ## unpublished
+- Feature: Add new configuration parameter `showZoomSlider` to configure a zoom slider display.
```
dopenguin, line 5:
```suggestion

- Feature: Add new configuration parameter `showZoomSlider` to configure a zoom slider display.
```

packages/plugins/Zoom/README.md
```diff
@@ -13,6 +13,7 @@ It can be configured as followed.
 |------------|------------------------------|-----------------------------------------------------------------------------------------------|
 | renderType | 'iconMenu' \| 'independent'? | Whether the zoom related buttons are being rendered independently or as part of the IconMenu. Defaults to `'independent'`. |
 | showMobile | boolean?                     | Whether the zoom related buttons should be displayed on smaller devices; defaults to `false`.   |
+| showZoomSlider | boolean? | Whether a zoom slider should be displayed under the zoom buttons. Won't be rendered on mobiles; defaults to `false`.|
```
dopenguin, line 16:
```suggestion
| showZoomSlider | boolean? | Whether a zoom slider should be displayed under the zoom buttons. Won't be rendered on smaller devices; defaults to `false`.|
```

packages/plugins/Zoom/src/components/Zoom.vue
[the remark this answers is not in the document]
dopenguin: However, it is still dependent on the parameter `showZoomButtons` as the outer `div` is only being displayed if `showZoomButtons` is `true`.

I suggest moving the `v-if` to both `v-tooltip`s of the regular buttons

packages/plugins/Zoom/src/components/Zoom.vue
```diff
@@ -40,26 +40,40 @@
       </template>
       <span>{{ $t('common:plugins.zoom.out') }}</span>
     </v-tooltip>
+    <v-tooltip left :disabled="hasSmallDisplay">
+      <template #activator="{ on, attrs }">
+        <div v-bind="attrs" v-on="on">
+          <ZoomSlider v-if="addZoomSlider"></ZoomSlider>
```
dopenguin, line 46:
```suggestion
    <v-tooltip v-if="addZoomSlider" left :disabled="hasSmallDisplay">
      <template #activator="{ on, attrs }">
        <div v-bind="attrs" v-on="on">
          <ZoomSlider></ZoomSlider>
```

packages/plugins/Zoom/src/components/ZoomSlider.vue
```diff
@@ -0,0 +1,67 @@
...
+  <div>
+    <v-slider
+      v-model="zoomLevelRange"
+      vertical
+      class="polar-zoom-slider"
+      :min="minimumZoomLevel"
+      :max="maximumZoomLevel"
+      :aria-label="$t('common:plugins.zoom.slider')"
+    ></v-slider>
+  </div>
```
dopenguin, line 11:
```suggestion
  <v-slider
    v-model="zoomLevelRange"
    vertical
    class="polar-zoom-slider"
    :min="minimumZoomLevel"
    :max="maximumZoomLevel"
    :aria-label="$t('common:plugins.zoom.slider')"
  ></v-slider>
```

As written in https://github.com/Dataport/polar/pull/138#discussion_r1666540302

packages/plugins/Zoom/src/components/ZoomSlider.vue
```diff
@@ -0,0 +1,67 @@
...
+.polar-zoom-slider .v-slider__thumb {
+  width: 16px;
+  height: 16px;
+  border-radius: 2px;
+  left: -8px;
+}
+
+.polar-zoom-slider :hover .v-slider__track-background,
+.polar-zoom-slider :hover .v-slider__track-fill,
+.polar-zoom-slider :hover .v-slider__thumb {
```
dopenguin, line 55:
```suggestion
.polar-zoom-slider :hover .v-slider__thumb,
.polar-zoom-slider :focus .v-slider__thumb {
```

dopenguin (changes requested):
> > * The [OpenLayers example](https://openlayers.org/en/latest/examples/zoomslider.html), which uses the basic OpenLayers control, transitions between the zoom levels with ease. Do you think that adding that functionality here is feasible? Or do you think that that behaviour is not easily implemented with the current setup of `@polar/plugin-zoom`
                               >
                               > I tried the OpenLayers control first but did not succeed with implementing them within the plugin. Since the buttons where also implemented without OpenLayers I figured it would be a good idea to use the code that is already there.
                               >
                               > @warm-coolguy I changed the style of the v-slider according to your suggestions and I removed the original hover effect.
>
> @dopenguin 🏓

Please also add the parameter `showZoomSlider` to `ZoomConfiguration` in `@polar/lib-custom-types` including a Changelog entry.

🏓 @jedi-of-the-sea

Also: It kinda bothers me that the vuetify component allows a user to press Arrow-Left or Arrow-Right to change the slider if it is vertical :D But we can leave it at that

jedi-of-the-sea:
I implemented your new suggestions @dopenguin 🏓 

> 
> Also: It kinda bothers me that the vuetify component allows a user to press Arrow-Left or Arrow-Right to change the slider if it is vertical :D But we can leave it at that

Hu, I was not able to recreate that - or I did something wrong?

packages/plugins/Zoom/src/components/Zoom.vue
dopenguin: The `v-if` is still present on the `ZoomSlider` and not on `v-tooltip`

dopenguin (changes requested):
> I implemented your new suggestions @dopenguin 🏓
> 
> > Also: It kinda bothers me that the vuetify component allows a user to press Arrow-Left or Arrow-Right to change the slider if it is vertical :D But we can leave it at that
> 
> Hu, I was not able to recreate that - or I did something wrong?

It's not a big problem and it can stay. To recreate though:

- Have a normal desktop view
- Tab until you are focused on the ZoomSlider
- Press Arrow-Right or Arrow-Left
- The Slider moves

Also, there seem to be some merge conflicts that need to be addressed.

🏓 @jedi-of-the-sea

packages/plugins/Zoom/src/components/Zoom.vue
jedi-of-the-sea: has been changed

jedi-of-the-sea:
> > I implemented your new suggestions @dopenguin 🏓
> > > Also: It kinda bothers me that the vuetify component allows a user to press Arrow-Left or Arrow-Right to change the slider if it is vertical :D But we can leave it at that
> > 
> > 
> > Hu, I was not able to recreate that - or I did something wrong?
> 
> It's not a big problem and it can stay. To recreate though:
> 
> * Have a normal desktop view
> * Tab until you are focused on the ZoomSlider
> * Press Arrow-Right or Arrow-Left
> * The Slider moves
> 
> Also, there seem to be some merge conflicts that need to be addressed.
> 
> 🏓 @jedi-of-the-sea

OK now it works. But if it's ok I'll not change that for now^^

@dopenguin 🏓

dopenguin (approved):
🚀 @jedi-of-the-sea

PR #140, opened 2024-07-11
update @polar/core dependencies

warm-coolguy:
## Summary

All @polar/core dependencies have been updated where updates, where compatible with our Vue@2 setup.

No breaking change affected us, and most dependencies were implicitly already installed in the version noted now. Clients can be arbitrarily tested for i18n still working, which was the only breaking change.


dopenguin (approved):
🚀

PR #141, opened 2024-07-11
update jspdf dependency

warm-coolguy:
## Summary

The @polar/plugin-export dependency has been updated to the already installed version.

No test required.


dopenguin (approved):
🚀

PR #142, opened 2024-07-12
update various dependencies

warm-coolguy:
## Summary

Trivially updatable dependencies have been pulled up to newer versions.

* For Babel, dev mode and build mode have been tested.
* For rimraf, the scripts using rimraf have been tested.
* For @github/actions, we'll see what happens after merging. I'm optimistically updating since [only old node versions have been dropped as the breaking change](https://github.com/actions/toolkit/blob/main/packages/github/RELEASES.md).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -30,10 +30,10 @@
         "vuex": "^3.6.2"
       },
       "devDependencies": {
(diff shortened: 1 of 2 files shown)


dopenguin (approved):
🚀

PR #143, opened 2024-07-12
remove unused dependencies

warm-coolguy:
## Summary

- "events" is still installed as dependency of another dependency. However, it is not a direct dependency, and thus has been removed.
- SBOM generation did not work and the sbom is outdated and thus misleading. The file and package have been removed. An SBOM mechanism is in development at https://github.com/Dataport/polar/tree/feature/provide-sbom in parallel.

No tests required.


dopenguin (approved):
🚀

PR #144, opened 2024-07-15
Feature/add scale switcher

jedi-of-the-sea:
## Summary

New Feature Scale Switcher was added to the Scale plugin.

As it is now, the scale switcher is shown if zoom options are configured and the parameter showScaleSwitcher is true.
```
scale: {
    showScaleSwitcher: true,
    zoomMethod: 'plugin/zoom/setZoomLevel'
  },
```

I am not sure if the zoom options for the scale switcher should be configured as it brings some problems, e.g. in my case the scale in the map differs from the scale in the options, and if one zooms to '1:50' they might get '1:45' instead. The reason is the difference between the default dpi (96) and the real dpi, which is a different one in my case. The scale will show the right number, but it is not the scale that was chosen from the scale switcher. I tried to fix this by calculating the scale resolution and setting the scale without zoom options but somehow could not solve this problem. Since I'm not sure how to solve this or how to go on from here, I just put it as 'Draft PR' since I'd like some feedback on the implementation so far and my thoughts to the problem.


packages/plugins/Scale/src/store/index.ts
```diff
@@ -76,6 +72,23 @@ export const makeStoreModule = () => {
     },
     getters: {
       ...generateSimpleGetters(getInitialState()),
+      zoomOptions: (_, __, ___, rootGetters) => {
+        const options =
+          rootGetters.configuration?.scale?.options ||
```
dopenguin, line 77: This is probably related to the question in the PR description: 

Why is there a separate option if all the scales are already coming from one place (the fallback `rootGetters.configuration.options`)? I don't really see the necessity

packages/plugins/Scale/src/store/index.ts
```diff
@@ -76,6 +72,23 @@ export const makeStoreModule = () => {
...
+          (option: { scale: number; zoomLevel: number }) =>
+            typeof option.scale === 'number' &&
+            typeof option.zoomLevel === 'number'
+        )
+        filteredOptions.unshift(translate('common:plugins.scale.scaleSwitcher'))
+
+        return filteredOptions.length > 1 ? filteredOptions : []
+      },
+      showScaleSwitcher: (_, __, ___, rootGetters) => {
+        return rootGetters.configuration?.scale?.showScaleSwitcher
```
dopenguin, line 90: This parameter should be added to a table in the READM.md

jedi-of-the-sea:
@dopenguin 🏓

dopenguin (changes requested):
As `scale` now also has configuration parameters, an entry needs to be added to `MapConfiguration` in `@polar/lib-custom-types`.

🏓 @jedi-of-the-sea

packages/plugins/Scale/src/components/Scale.vue
```diff
@@ -4,6 +4,23 @@
...
+      <option
+        v-for="(option, i) in zoomOptions"
+        :key="i"
+        :value="option"
+        class="scale-as-a-ratio"
+      >
+        {{ i === 0 ? option : scaleNumberToScale(option.scale) }}
+      </option>
+    </select>
     <span class="scale-as-a-ratio">
```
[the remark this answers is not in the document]
jedi-of-the-sea: The problem here (which was why I implemented a seperate switcher and a scale display instead of one element that does both) was, that the calculated scale would not be in the options selection, so the select element would not display the current scale. I changed it now so that the next best option.scale from the options selection is displayed. But that means that the displayed scale might differ from the calculated scale. I'm not sure about this solution, but maybe this is ok for the moment?

jedi-of-the-sea:
@dopenguin 🏓

packages/plugins/Scale/src/language.ts
```diff
@@ -8,6 +8,7 @@ const language: LanguageOption[] = [
         scale: {
           toOneNegative: 'Skalierung muss eine positive Zahl sein',
           label: 'Skala',
+          scaleSwitcher: 'Maßstab wählen',
```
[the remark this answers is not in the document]
jedi-of-the-sea: commit 'remove unneded translation keys'

packages/plugins/Scale/src/store/index.ts
```diff
@@ -76,6 +72,21 @@ export const makeStoreModule = () => {
...
+        const filteredOptions = options?.filter(
+          (option: { scale: number; zoomLevel: number }) =>
+            typeof option.scale === 'number' &&
+            typeof option.zoomLevel === 'number'
+        )
+        filteredOptions.unshift(translate('common:plugins.scale.scaleSwitcher'))
+
+        return filteredOptions.length > 1 ? filteredOptions : []
+      },
+      showScaleSwitcher: (_, __, ___, rootGetters) => {
```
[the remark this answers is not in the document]
jedi-of-the-sea: commit 'add interface ScaleGetters to types'

packages/plugins/Scale/src/components/Scale.vue
```diff
@@ -59,6 +120,31 @@ export default Vue.extend({
...
+      0 4px 5px 0 rgba(0, 0, 0, 0.14), 0 1px 10px 0 rgba(0, 0, 0, 0.12);
+    border: 3px solid transparent;
+    box-sizing: border-box;
+  }
+
+  .scale-switcher:hover,
+  .scale-switcher:focus {
+    cursor: pointer;
+    border: 3px solid #ffffff;
+    outline: 3px solid #003064;
```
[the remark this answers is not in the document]
jedi-of-the-sea: commit 'use css variables for styling'

packages/plugins/Scale/src/components/Scale.vue
```diff
@@ -59,6 +120,31 @@ export default Vue.extend({
     padding-right: 1em;
   }
 
+  .scale-switcher {
+    cursor: pointer;
+    background: #003064;
+    color: #ffffff;
```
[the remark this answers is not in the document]
jedi-of-the-sea: commit 'use css variables for styling'

packages/plugins/Scale/src/components/Scale.vue
```diff
@@ -15,12 +32,55 @@
...
+            )
+          ] || this.zoomOptions[0]
+        )
+      },
+      set(value: number) {
+        this.setScaleValue(value)
+      },
+    },
+    showSelectOptions() {
+      return this.showScaleSwitcher && this.zoomOptions.length > 0
```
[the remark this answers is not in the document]
jedi-of-the-sea: commit 'check length of zoom options in showScaleSwitcher'

packages/plugins/Scale/src/components/Scale.vue
```diff
@@ -15,12 +32,55 @@
...
   computed: {
-    ...mapGetters('plugin/scale', ['scaleToOne', 'scaleWithUnit']),
+    ...mapGetters({
+      zoomOptions: 'plugin/scale/zoomOptions',
+      scaleToOne: 'plugin/scale/scaleToOne',
+      scaleValue: 'plugin/scale/scaleValue',
+      scaleWithUnit: 'plugin/scale/scaleWithUnit',
+      showScaleSwitcher: 'plugin/scale/showScaleSwitcher',
+      getMapView: 'map/getMapView',
+    }),
```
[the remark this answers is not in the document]
jedi-of-the-sea: commits
'unnecessary getter'
'shorten mapGetters import in component'

packages/plugins/Scale/src/components/Scale.vue
```diff
@@ -4,6 +4,23 @@
...
+      class="scale-as-a-ratio scale-switcher"
+      @change="setZoomLevelByScale($event.target.selectedIndex)"
+    >
+      <option
+        v-for="(option, i) in zoomOptions"
+        :key="i"
+        :value="option"
+        class="scale-as-a-ratio"
+      >
+        {{ i === 0 ? option : scaleNumberToScale(option.scale) }}
```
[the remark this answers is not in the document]
jedi-of-the-sea: commit
'show next best matching scale of zoom options if no mathcing scale is found'

packages/plugins/Scale/src/components/Scale.vue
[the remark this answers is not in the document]
jedi-of-the-sea: commit
'show next best matching scale of zoom options if no mathcing scale is found'

packages/types/custom/core.ts
```diff
@@ -592,6 +596,7 @@ export interface MapConfig {
   iconMenu?: IconMenuConfiguration
   pins?: PinsConfiguration
   reverseGeocoder?: ReverseGeocoderConfiguration
+  scale?: ScaleConfiguration
```
dopenguin, line 599: Please add a changelog entry for this addition including the reference to the newly added interface `ScaleConfiguration`

packages/types/custom/core.ts
```diff
@@ -431,6 +431,10 @@ export type ToastTypeStyles = {
 /** configuration for toast type styles */
 export type ToastConfiguration = PluginOptions & ToastTypeStyles
 
+export interface ScaleConfiguration {
```
dopenguin, line 434: Just like with every other configuration option, this needs to extend `PluginOptions`.

Also, I've missed that in `packages/plugins/Scale/src/index.ts` the type needs to be adjusted to `ScaleConfiguration` as well.

packages/plugins/Scale/src/components/Scale.vue
```diff
@@ -15,12 +32,57 @@
...
+    showSelectOptions() {
+      return this.showScaleSwitcher
+    },
+  },
+  methods: {
+    ...mapMutations('plugin/scale', ['setScaleValue']),
+    ...mapActions('plugin/zoom', ['setZoomLevel']),
+    // TODO: Finding a different solution may be a task to be tackled in the future
+    setZoomLevelByScale(index: number) {
+      if (!this.$store._actions['plugin/zoom/setZoomLevel']) {
```
dopenguin, line 78:
```suggestion
      if (!this.$store.hasModule(['plugin', 'zoom'])) {
```
As implemented in `packages/plugins/Gfi/src/components/Gfi.vue`

packages/plugins/Scale/src/components/Scale.vue
```diff
@@ -15,12 +32,57 @@
...
+        console.error(
+          'Action "setZoomLevel" is not available. To use this feature, please add the zoom plugin.'
```
dopenguin, line 80:
```suggestion
          '@polar/plugin-scale: "@polar/plugin-zoom" not available. Please install and configure the plugin to be able to use the scale switcher.'
```

packages/plugins/Scale/src/components/Scale.vue
```diff
@@ -15,12 +32,57 @@
...
+        })
+        return bestMatchingScale.scale
+      },
+      set(value: number) {
+        this.setScaleValue(value)
+      },
+    },
+    showSelectOptions() {
+      return this.showScaleSwitcher
+    },
```
dopenguin, line 71: I just had the idea that the `select` of the scale switcher does not get used if `@polar/plugin-zoom` is not available and the normal scale is displayed instead,
One could add a new parameter to `data` called `zoomPluginAvailable` which is set to `false` in `mounted` in that case and `showSelectOptions` is also reliant on that parameter.

Feel free to add it or leave it as is right now 🎩

packages/plugins/Scale/README.md
```diff
@@ -9,3 +9,7 @@ The scale plugin shows the current map resolution as relative scale (1:x) and/or
 ## Configuration
 
 For details on the `displayComponent` attribute, refer to the [Global Plugin Parameters](../../core/README.md#global-plugin-parameters) section of `@polar/core`.
+
+| fieldName | type | description |
+| - | - | - |
+| showScaleSwitcher | boolean? | If set to `true`, a button to switch to a specific scale will be rendered. Defaults to `false`. Requires the configuration parameter `options` set in the `mapConfiguration`, see `<...masterportalAPI.fields>` in the `@polar/core` package documentation for more information. Needs the Zoom Plugin to work, so make sure that it is added in addPlugins
```
dopenguin, line 15:
```suggestion
| showScaleSwitcher | boolean? | If set to `true`, a button to switch to a specific scale will be rendered. Defaults to `false`. Requires the configuration parameter `options` set in the `mapConfiguration`, see `<...masterportalAPI.fields>` in the `@polar/core` package documentation for more information. Also requires the `@polar/plugin-zoom` to be installed and used. |
```
Slight wording change and syntax fix

packages/plugins/Scale/src/components/Scale.vue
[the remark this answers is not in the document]
dopenguin: I think this seems fine for now. I'll have that in mind when writing the issue regarding the dpi problem of scale. 🎩

dopenguin (changes requested):
🏓 @jedi-of-the-sea

packages/types/custom/core.ts
jedi-of-the-sea: 032a181894f6ef38c73fe650f71c0a8c82e77127

packages/types/custom/core.ts
jedi-of-the-sea: 6ffeaeac8721fd673e2156164c8a500b816f7993

jedi-of-the-sea: 1b41081c579aa08e44b16df93e50f38127afe9e6

packages/plugins/Scale/src/components/Scale.vue
jedi-of-the-sea: 74cf864f134df9513118b6b33e0c1223a18f776a

jedi-of-the-sea:
@dopenguin 🏓

jedi-of-the-sea:
@dopenguin I now implemented the changes we discussed today.

packages/plugins/Scale/README.md
```diff
@@ -9,3 +9,8 @@ The scale plugin shows the current map resolution as relative scale (1:x) and/or
 ## Configuration
 
 For details on the `displayComponent` attribute, refer to the [Global Plugin Parameters](../../core/README.md#global-plugin-parameters) section of `@polar/core`.
+
+| fieldName | type | description |
+| - | - | - |
+| showScaleSwitcher | boolean? | If set to `true`, a button to switch to a specific scale will be rendered. Defaults to `false`. Requires the configuration parameter `options` set in the `mapConfiguration`, see `<...masterportalAPI.fields>` in the `@polar/core` package documentation for more information. Also requires the parameter `zoomMethod` to be configured. |
+| zoomMethod | string? | Path to the action that is used for the zoom of the scale switcher. Has to take zoomLevel als payload. Must be configured for the scale switcher, otherwise it will not be rendered.
```
dopenguin, line 16:
```suggestion
| zoomMethod | string? | Path to the action that is used for the zoom of the scale switcher. Has to take `zoomLevel ` as payload. Must be configured for the scale switcher, otherwise it will not be rendered.
```

packages/plugins/Scale/src/store/index.ts
```diff
@@ -70,12 +66,45 @@ export const makeStoreModule = () => {
...
+          getters.zoomOptions.length > 0
+        )
+      },
+      zoomMethod: (_, __, ___, rootGetters) => {
+        return (
+          rootGetters.configuration?.scale?.zoomMethod ||
+          console.error(
+            "@polar/plugin-scale: Configuration parameter `zoomMethod` is missing. Scale switcher won't be rendered."
+          )
+        )
```
dopenguin, line 106:
```suggestion
        const zoomMethod = rootGetters.configuration?.scale?.zoomMethod

        if (!zoomMethod) {
          console.error(
            "@polar/plugin-scale: Configuration parameter `zoomMethod` is missing. Scale switcher won't be rendered."
          )
          return ''
        }
        return zoomMethod
```
This may otherwise lead to the situation that `zoomMethod` may not be a string (contrary to the type definition).

packages/plugins/Scale/src/store/index.ts
```diff
@@ -70,12 +66,45 @@ export const makeStoreModule = () => {
             : `${scaleNumber} m`
         commit('setScaleWithUnit', newValue)
       },
+      zoomToScale(
+        { dispatch, rootGetters: { configuration } },
+        zoomLevel: number
+      ): void {
+        if (configuration?.scale?.zoomMethod) {
+          dispatch(configuration.scale.zoomMethod, zoomLevel, { root: true })
```
dopenguin, line 74:
```suggestion
      zoomToScale({ dispatch, getters }, zoomLevel: number): void {
        const { zoomMethod } = getters
        if (zoomMethod.length > 0) {
          dispatch(zoomMethod, zoomLevel, { root: true })
```
We can simply use the local getter if we have it :D

dopenguin (changes requested):
🏓 @jedi-of-the-sea

packages/plugins/Scale/README.md
```diff
@@ -9,3 +9,8 @@ The scale plugin shows the current map resolution as relative scale (1:x) and/or
 ## Configuration
 
 For details on the `displayComponent` attribute, refer to the [Global Plugin Parameters](../../core/README.md#global-plugin-parameters) section of `@polar/core`.
```
dopenguin, line 11: In other plugins this is written below the table. Mind moving it there as well? :)

jedi-of-the-sea: https://github.com/Dataport/polar/pull/144/commits/bc9722477d4861bf26bbcffd160703a529e286cc

jedi-of-the-sea:
@dopenguin 🏓

packages/plugins/Scale/README.md
```diff
@@ -8,4 +8,11 @@ The scale plugin shows the current map resolution as relative scale (1:x) and/or
 
 ## Configuration
 
+### scale
+
+| fieldName | type | description |
+| - | - | - |
+| showScaleSwitcher | boolean? | If set to `true`, a button to switch to a specific scale will be rendered. Defaults to `false`. Requires the configuration parameter `options` set in the `mapConfiguration`, see `<...masterportalAPI.fields>` in the `@polar/core` package documentation for more information. Also requires the parameter `zoomMethod` to be configured. |
+| zoomMethod | string? | Path to the action that is used for the zoom of the scale switcher. Has to take `zoomLevel` as payload. Must be configured for the scale switcher, otherwise it will not be rendered.
```
dopenguin, line 16:
```suggestion
| zoomMethod | string? | Path to the action that is used for the zoom of the scale switcher. Has to take `zoomLevel` as payload. Must be configured for the scale switcher, otherwise it will not be rendered. |
```

dopenguin (changes requested):
🏓 @jedi-of-the-sea

jedi-of-the-sea:
@dopenguin 🏓

dopenguin (approved):
🚀 @jedi-of-the-sea

PR #150, opened 2024-07-30
Fix typing across the project

dopenguin:
## Summary

Fixed types at multiple places in the project.
After merging this PR and #149, only selected errors in some clients should remain.

`tsc --noEmit` should only yield some errors in clients.

Blocked by #149, as the type fixes of that PR should be added first.


tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es6",
+    "target": "ES2018",
```
warm-coolguy, line 3: This is affecting the produced code's syntax niveau, and hence reduces the set of browsers the clients will work in. Why exactly is this change required? According to https://www.typescriptlang.org/tsconfig/#target , ES6 is still recommended, even for modern browsers.

Since this may be a breaking change for some users, I'd try to avoid this.

warm-coolguy (changes requested):

tsconfig.json
dopenguin: error

We'd need to refactor the respective code parts.
I think we already only support a reduced set of browsers by including this in an always called part of `@polar/core`.

🏓 @warm-coolguy

dopenguin: I've also just added the type check to the pipeline.
It excludes type errors in clients for now.

warm-coolguy: https://caniuse.com/?search=named%20capture

While named capturing groups are implemented in all modern browsers, other features may be not, and I'm missing a complete list of which browser implements which ES2018 features. For a simple solution, we can @ts-expect-error on this and revert the change to the target.

dopenguin: Done 82d4e646be11fae7e64b96cab936323604afd8df

🏓  @warm-coolguy

You can proceed with the review as soon as #149 is merged

dopenguin: #149 is merged 🎉

warm-coolguy (approved):
🚀 @dopenguin

PR #154, opened 2024-08-01
Feature/easing between zoom levels

jedi-of-the-sea:
## Summary

With the new changes, the zoom between zoom levels works smoother now. 

Start a client and zoom in and out via zoom buttons or mouse.

I assumed that the smooth zoom is preferred and should be standard, so I just replaced the old zoom with the new method.

I'm not sure if the duration of the animation should be configurable, I just used a value of 1000. Let me know if this is something that I should add.

### DIFF
--- a/packages/plugins/Zoom/src/store/index.ts
+++ b/packages/plugins/Zoom/src/store/index.ts
@@ -56,7 +56,7 @@ export const makeStoreModule = () => {
         if (view) {
           view.animate({
             zoom: zoomLevel,
-            duration: 1000,
+            duration: 500,
           })
         }
       },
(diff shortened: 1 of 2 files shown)


packages/plugins/Zoom/src/store/index.ts
```diff
@@ -50,6 +51,15 @@ export const makeStoreModule = () => {
       decreaseZoomLevel({ dispatch, getters: { zoomLevel } }): void {
         dispatch('setZoomLevel', zoomLevel - 1)
       },
+      smoothZoom({ rootGetters: { map }, getters: { zoomLevel } }): void {
+        const view = map.getView()
+        if (view) {
+          view.animate({
+            zoom: zoomLevel,
+            duration: 1000,
```
dopenguin, line 59:
```suggestion
            duration: 500,
```
A second felt a bit too slow

dopenguin (changes requested):
🏓 @jedi-of-the-sea

jedi-of-the-sea:
@dopenguin 🏓 I changed the duration to 500

dopenguin (approved):
🚀 @jedi-of-the-sea

PR #155, opened 2024-08-02
Fix sass deprecation issue regarding declaration position

dopenguin:
## Summary

The current behaviour for mixed declarations in `sass` has been deprecated (see https://sass-lang.com/documentation/breaking-changes/mixed-decls/).
This PR changes the only instance of such mixed declarations.

There was a deprecation warning on any `npm run x:dev`. This should no longer occur.


warm-coolguy (approved):
🚀 @dopenguin

PR #156, opened 2024-08-05
custom attributes for the like filter

ludoholic:
addressSearch.searchMethodsObject.queryParameters (type:wfs) now supports custom attributes for the like filter operator.

## Summary
- buildWfsFilter:wfsLike() now reads likeFilterAttributes from WfsParameters and maps them to <ogc:PropertyIsLike> attributes
- updated readme.md in AddressSearch plugin

- Use the following map configuration for addressSearch:
```
  addressSearch: {
    searchMethods: [
      {
        queryParameters: {
          xmlns: 'https://registry.gdi-de.org/id/de.hh.up',
          featurePrefix: 'de.hh.up',
          typeName: 'strassenbaumkataster',
          fieldName: 'gattung_deutsch',
          likeFilterAttributes: {
            wildCard: '%',
            singleChar: '.',
            escapeChar: '!',
          },
        },
        type: 'wfs',
        url: 'https://geodienste.hamburg.de/HH_WFS_Strassenbaumkataster?service=WFS&request=GetFeature&version=1.1.0',
      },
    ],
    minLength: 3,
    waitMs: 300,
  },
```
- try to use "%" as wildcard in search string, e.g. "%pel" will find all "Pappel" trees
- remove likeFilterAttributes in map configuration and try searching with "*" (default) as wildcard


packages/lib/getFeatures/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## unpublished
 
 - Breaking: `getWfsFeatures` now throws errors if required parameters on the wfs configuration are missing instead of only printing error messages on the console.
+- Feature: addressSearch.searchMethodsObject.queryParameters (type:wfs) now supports custom attributes for the like filter operator.
```
dopenguin, line 6: This package can be used independently to `@polar/plugin-address-search` and thus the changelog entry should reflect the changes to this package. Something like

```md
- Feature: `getWfsFeatures` now offers the possibility to adjust the `wildCard`, `singleChar` and `escapeChar` properties of the `PropertyIsLike`-filter.
```

should be fine. Feel free to alter the text as you please though.

packages/lib/getFeatures/wfs/buildWfsFilter.ts
```diff
@@ -23,14 +30,28 @@ const queryPrefix = ({
...
+  )}>
 <ogc:PropertyName>${featurePrefix}:${fieldName}</ogc:PropertyName>
 <ogc:Literal>${input}${
-  typeof useRightHandWildcard === 'boolean' && !useRightHandWildcard ? '' : '*'
-}</ogc:Literal>
+    typeof useRightHandWildcard === 'boolean' && !useRightHandWildcard
+      ? ''
+      : Object.hasOwn(mergedLikeFilterAttributes, 'wildCard')
+      ? mergedLikeFilterAttributes.wildCard
+      : '*'
```
dopenguin, line 51: This might break for some services.

When I implemented the `wfsSearch` for the Masterportal, I found that the key for the three properties may vary as well and thus wrote the following documentation in the Masterportal

```md
#### Portalconfig.menu.tool.wfsSearch.searchInstance.requestConfig.likeFilter

Values inside a filter for a WFS service can be compared with an `equal` or a `like`.
If the comparison should be with a `like` then the filter needs additional properties. These may vary in value and property definition.
For the documentation, it is assumed that the properties are called `wildCard`, `singleChar` and `escapeChar`; variations like e.g. `single` and `escape` are possible and need to be configured in line with the service. All key-value pairs are used in the request as given.
```

If I now have a service that defines the properties as `wildcard`, `escape` and `singleChar`, the default values are being added to the Filter as well.
Also, when I define wildcard differently (value and key), the wrong wildcard char would be used.

I think it is easiest, if either all properties are defined on the object (if present) or none so the fallback case can be used.

packages/plugins/AddressSearch/README.md
```diff
@@ -120,6 +120,7 @@ These fields are interpreted by all implemented services.
 | patterns | string[]? | Allows specifying input patterns. In a single-field search, a pattern can be as easy as `{{theWholeThing}}`, where `theWholeThing` is also the feature field name to search in. In more complex scenarios, you may add separators and multiple fields, e.g. `{{gemarkung}} {{flur}} {{flstnrzae}}/{{flstnrnen}}` would fit many parcel search services. Mutually exclusive to `fieldName`. |
 | patternKeys | Record<string, string>? | Maps field names from patterns to regexes. Each field name has to have a definition. Each regex must have one capture group that is used to search. Contents next to it are ignored for the search and just used for matching. E.g. `'([0-9]+)$'` would be a value for a key that fits an arbitrary number string at the input's end. |
 | srsName | string? | Name of the projection (srs) for the query. |
+| likeFilterAttributes | Record<string, string>? | Key/value pairs that are mapped to custom attributes of the isLike operators used in filters. The default is {wildCard: "\*",singleChar: ".",escapeChar: "!"} which results in <ogc:PropertyIsLike wildCard="*" singleChar="." escapeChar="!"> |
```
dopenguin, line 123: As mentioned in the comment in `packages/lib/getFeatures/wfs/buildWfsFilter.ts`, the key words may vary.
I think this should be reflected here as well.

Also, it should be written that all three parameters have to be defined on the object.

Also also, I've added a [new ADR](https://github.com/Dataport/polar/wiki/Architecture-Decision-Records-%F0%9F%93%83#2024-08-05-configuration-parameters-in-tables-have-to-be-ordered-by-a-required-and-b-alphabetically) regarding the order of parameters in the documentation.

dopenguin (changes requested):
- Please add a changelog entry for the new configuration parameter of `@polar/plugin-address-search`
- Please adjust the type `AddressSearchConfiguration` in `@polar/lib-custom-types` and add a changelog entry

With the given configuration, I can not seem to find any results. The request is being properly sent, the `FeatureCollection` being returned is empty. 
Do I have to use any browser plugins (CORS?) or am I missing something else?

🏓 @ludoholic

packages/lib/getFeatures/wfs/buildWfsFilter.ts
ludoholic: > If I now have a service that defines the properties as wildcard, escape and singleChar, the default values are being added to the Filter as well.
I know and this should not be a problem. Additional attributes should be ignored by the service. You could also add an attribute which is called `killserver=true` and it won`t kill the server. ;)

> Also, when I define wildcard differently (value and key), the wrong wildcard char would be used.
That's right. The only way to fix this is to splitt the configuration like this:
```
{
wildCardKey: "wild",
wildCardChar: "*"
}
```
But I am not happy with this and we will loose some flexibility by this. Because now it is also possible to add any arbitrary attributes to the filter operator. Which might be handy for the admin of the service to send custom data within.
So I chose the same way the masterportal did: the righthand wildcard key must be called "wildcard" (see [here](https://bitbucket.org/geowerkstatt-hamburg/masterportal/src/42f370cdeec25afa4a2b0aa41bb07e2c6e1c529f/src/modules/wfsSearch/js/buildFilter.js#lines-79)) Maybe we just add this last sentence analogously to the documentation?

> I think it is easiest, if either all properties are defined on the object (if present) or none so the fallback case can be used.
Even when all properties are defined on the object, we don`t know which one is the wildcard replacement. E.g. think about `{joker="*", esc="!", single="*"}`. There is no chance to guess which one is for the wildcard. So I thing the only correct way is mentioned above.

packages/plugins/AddressSearch/README.md
ludoholic: > As mentioned in the comment in packages/lib/getFeatures/wfs/buildWfsFilter.ts, the key words may vary.
I think this should be reflected here as well.

It is mentioned here: "**Key**/**value** pairs are mapped to custom attributes". So the key and the value is mapped.

> Also, it should be written that all three parameters have to be defined on the object.

You don't have to define all three parameters.

> Also also, I've added a [new ADR](https://github.com/Dataport/polar/wiki/Architecture-Decision-Records-%F0%9F%93%83#2024-08-05-configuration-parameters-in-tables-have-to-be-ordered-by-a-required-and-b-alphabetically) regarding the order of parameters in the documentation.

I will adjust it.

ludoholic:
> Please add a changelog entry for the new configuration parameter of @polar/plugin-address-search

Will be done.

> Please adjust the type AddressSearchConfiguration in @polar/lib-custom-types

The needed level of configuration is in the readme of the plugin, see [here](https://github.com/Dataport/polar/blob/feature/custom-like-filter-attributes/packages/lib/getFeatures/types.ts#L45). In AddressSearchConfiguration in @polar/lib-custom-types it is also refered to the specific parameters in the plugin, see [here](https://github.com/Dataport/polar/blob/feature/custom-like-filter-attributes/packages/types/custom/core.ts#L42).

> With the given configuration, I can not seem to find any results. The request is being properly sent, the FeatureCollection being returned is empty.
> Do I have to use any browser plugins (CORS?) or am I missing something else?

No, that should work without any browser plugins. Which client did you use?

packages/lib/getFeatures/CHANGELOG.md
ludoholic: I altered the text. It is not only possible to adjust the `wildCard`, `singleChar` and `escapeChar` attributes, it is possible to add any attributes to the `PropertyIsLike` operator.

ludoholic:
🏓 @dopenguin 
Made some changes and startet some discussions. ;)

packages/plugins/AddressSearch/README.md
dopenguin: 📞 ✔️

packages/lib/getFeatures/wfs/buildWfsFilter.ts
dopenguin: 📞 ✔️

packages/lib/getFeatures/CHANGELOG.md
dopenguin: 📞 ✔️

ludoholic:
🏓 @dopenguin

packages/plugins/AddressSearch/README.md
```diff
@@ -117,8 +117,9 @@ These fields are interpreted by all implemented services.
...
-| patterns | string[]? | Allows specifying input patterns. In a single-field search, a pattern can be as easy as `{{theWholeThing}}`, where `theWholeThing` is also the feature field name to search in. In more complex scenarios, you may add separators and multiple fields, e.g. `{{gemarkung}} {{flur}} {{flstnrzae}}/{{flstnrnen}}` would fit many parcel search services. Mutually exclusive to `fieldName`. |
+| likeFilterAttributes | Record<string, string>? | As specified by the [OGC-Standard for filters](https://schemas.opengis.net/filter/) the PropertyIsLike operator requires three attributes (e.g. in WFS 2.0.0: wildCard, singleChar and escapeChar). But these may vary in value and (with other WFS verions) also in property definition. Therefore it is here possible to configure the values of the attributes needed for WFS 2.0.0 and also to add custom attributes needed for other versions. The default is `{wildCard: "\*",singleChar: ".",escapeChar: "!"}`. Configuration examples: WFS 2.0.0 `{wildCard: "%",singleChar: "\*",escapeChar: "\"}`; WFS 1.0.0 `{wildCard: "\%",singleChar: "*",escape: "\"}` |
```
dopenguin, line 120:
```suggestion
| likeFilterAttributes | Record<string, string>? | As specified by the [OGC-Standard for filters](https://schemas.opengis.net/filter/) the `PropertyIsLike` operator requires three attributes (e.g. in WFS 2.0.0: `wildCard`, `singleChar` and `escapeChar`). These may vary in value and (with other WFS versions) also in property definition. Therefore, it is possible to configure the values of the attributes needed for WFS 2.0.0 and also to add custom attributes needed for other versions. Defaults to `{wildCard: "\*", singleChar: ".", escapeChar: "!"}`. |
```
I suggest adding the examples below the table for enhanced readability.

dopenguin (changes requested):
🏓 @ludoholic

ludoholic:
🏓 @dopenguin

dopenguin (approved):
🚀 @ludoholic

PR #157, opened 2024-08-05
Fix/test warnings

dopenguin:
## Summary

When running tests, errors used to popup regarding undefined getters. This was actually the case and has been fixed now by adding the to the testMountParameters.
Also, moved the folder of the tests of `@polar/plugin-gfi` to the correct place.

`npm run test`; only `console.error`s regarding a not defined layer should popup.


warm-coolguy (approved):
🚀 @dopenguin

PR #158, opened 2024-08-07
Add example configuration to documentation

jedi-of-the-sea:
## Summary

Example configurations have been added to the documentations of the core and the plugins.

PR #160, opened 2024-08-19
Feature/mpapi configuration parameters

dopenguin:
## Summary

- Enhance documentation of `@masterportal/masterportalApi` configuration parameters in `@polar/core`
- Add base values for some `@masterportal/masterportalApi` configuration parameters
- Move some configuration documentation of `layers` from `@polar/plugin-layer-chooser` to `@polar/core`
- Adjust `scripts/makeDocs.js` to change all links to anchors in the documentation of `@polar/core`

## TODO

- [x] Changelogs

 #149, #153, #158 have to be merged before this PR.


packages/core/README.md
```diff
@@ -253,48 +313,102 @@ Whitelisted and confirmed parameters include:
...
-| namedProjections | Array | Array of usable projections by proj4 string. |
-| options | Array | Defines all available zoomLevels. Entries define `resolution`, `scale`, and `zoomLevel`. See masterportalAPI for details. |
-| startCenter | number[] | Initial center coordinate. |
-| startResolution | number | Initial resolution; must be in options. See below. |
+| id | string | Service register id in `mapConfiguration.layerConf`. |
+| name | string | Display name in UI. |
+| type | enum["background", "mask"] | Layer handling. Backgrounds are mutually exclusive, masks ("overlays") can be stacked. |
+| maxZoom | number? | If set, layer only available (and selectable) up to this zoom level. |
+| minZoom | number? | If set, layer only available (and selectable) from this zoom level on. |
+| visibility | boolean? | Initial visibility. Defaults to `false`. |
```
warm-coolguy, line 368: Do these have an effect in the Core (resp. masterportalapi)?

packages/core/README.md
```diff
@@ -69,11 +80,10 @@ The mapConfiguration allows controlling many client instance details.
...
 | checkServiceAvailability | boolean? | If set to `true`, all services' availability will be checked with head requests. |
-| extendedMasterportalapiMarkers | extendedMasterportalapiMarkers? | Optional. If set, all configured visible vector layers' features can be hovered and selected by mouseover and click respectively. They are available as features in the store. Layers with `clusterDistance` will be clustered to a multi-marker that supports the same features. Please mind that this only works properly if you configure nothing but point marker vector layers styled by the masterportalAPI. |
+| extendedMasterportalapiMarkers | extendedMasterportalapiMarkers? | Optional. If set, all configured visible vector layers' features can be hovered and selected by mouseover and click respectively. They are available as features in the store. Layers with `clusterDistance` will be clustered to a multi-marker that supports the same features. Please mind that this only works properly if you configure nothing but point marker vector layers styled by the masterportalapi. |
+| language | enum["de", "en"]? | Initial language. |
```
warm-coolguy, line 86: Why has this become optional? What is the default value? 🤔

packages/core/README.md
```diff
@@ -220,16 +230,66 @@ A full documentation of the masterportalapiPolygonFillHatch is available at the
...
+
+The `<...masterportalapi.fields>` means that any masterportalapi field may also be used here _directly_ in the mapConfiguration. The fields described here are fields that are interesting for the usage of POLAR.
+Fields that are not set as required have default values.
+
+| fieldName | type | description |
+| - | - | - |
+| layerConf | layerConf | Layer configuration of all available layers as a service register. Layers defined here are not directly shown in a client, see `mapconfiguration.layers` for that. |
+| layers | layer[] | Configuration of layers that are supposed to be used in the respective client. All layers defined here have to have an entry in `mapConfiguration.layerConf`. If `@polar/plugin-layer-chooser` is installed and configured, all these layers will be displayed in that menu. |
+| startCenter | number[] | Initial center coordinate. Needs to be defined in the chosen leading coordinate system. |
+| backgroundImage | string? | Image to be displayed as a background of the map. Defaults to `''`, so no image being used. |
```
warm-coolguy, line 243: This effectively is supposed to change how some clients are rendered initially. On slow devices or when the WMS is slow, this does make a difference. I'm okay with `''` being the default, but now clients need the image configured.

Theoretically.

Practically, the backgroundImage feature doesn't work anyway, see e.g. the implementation of `setBackgroundImage` in the Meldemichel, where `"masterportalAPI's setBackgroundImage doesn't work in nested shadow roots."` is noted.

packages/core/README.md
```diff
@@ -220,16 +230,66 @@ A full documentation of the masterportalapiPolygonFillHatch is available at the
...
+| epsg | `EPSG:${string}`? | Leading coordinate system. The coordinate system has to be defined in `mapConfiguration.namedProjections` as well. Changing this value should also lead to changes in `mapConfiguration.startCenter`, `mapConfiguration.extent`, `mapConfiguration.options` and `mapConfiguration.startResolution` as they are described in or related to the leading coordinate system. Defaults to `'EPSG:25832'`. |
+| extent | number[]? | Map movement will be restricted to the rectangle described by the given coordinates. Unrestricted by default. |
+| namedProjections | Array<[string,string]>? | Array of usable coordinated systems mapped to a projection as a proj4 string. Defines the `'EPSG:25832'`, `'EPSG:3857'` and `'EPSG:4326'` by default. |
```
warm-coolguy, line 246:
```suggestion
| namedProjections | Array<[string,string]>? | Array of usable coordinated systems mapped to a projection as a proj4 string. Defines the `'EPSG:25832'`, `'EPSG:3857'` and `'EPSG:4326'` by default. If you set a value, please mind that all pre-configured projections are overridden, and requiring e.g. `'EPSG:4326'` will only work if it is also defined in your override. |
```

I see this going wrong the first time someone uses this, so I'd add this extra cautious sentence to make sure people don't run into hard-to-debug problems.

packages/core/src/utils/createMap/index.ts
```diff
@@ -26,13 +27,53 @@ export default async function createMap({
...
+        'EPSG:31467',
+        '+proj=tmerc +lat_0=0 +lon_0=9 +k=1 +x_0=3500000 +y_0=0 +ellps=bessel +nadgrids=BETA2007.gsb +units=m +no_defs +type=crs',
+      ],
+      [
+        'EPSG:4647',
+        '+proj=tmerc +lat_0=0 +lon_0=9 +k=0.9996 +x_0=32500000 +y_0=0 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs',
+      ],
+    ],
+    startResolution: 15.874991427504629,
+  }
```
warm-coolguy, line 68: Due to `Async function 'createMap' has too many lines (68). Maximum allowed is 50`, this should probably end up in a separate file. That'd also make it easier to find.

packages/lib/testMountParameters/index.ts
```diff
@@ -54,11 +54,46 @@ export default (): MockParameters => {
...
+          [
+            'EPSG:31467',
+            '+proj=tmerc +lat_0=0 +lon_0=9 +k=1 +x_0=3500000 +y_0=0 +ellps=bessel +nadgrids=BETA2007.gsb +units=m +no_defs +type=crs',
+          ],
+          [
+            'EPSG:4647',
+            '+proj=tmerc +lat_0=0 +lon_0=9 +k=0.9996 +x_0=32500000 +y_0=0 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs',
+          ],
+        ],
+        startResolution: 15.874991427504629,
```
warm-coolguy, line 96: Due to `Arrow function has too many lines (90). Maximum allowed is 50`, this is probably best in another file. I wonder whether we should publish our default masterportalapi config as a lib, so both core and this may rely upon it? Or maybe have the core export the default config, and use the defaults as parameter ...

This effort will also come in handy when we integrate E2E stuff.

packages/types/custom/CHANGELOG.md
```diff
@@ -11,6 +11,9 @@
 - Feature: Add new parameter `showZoomSlider` to `ZoomConfiguration`.
 - Feature: Add new types `LoaderStyles` and `LoadingIndicatorConfiguration`.
 - Feature: Add `loadingIndicator` to `MapConfig` to configure loader style.
+- Feature: Add new utility type `PartialBy`.
```
warm-coolguy, line 14: Why?

packages/types/custom/core.ts
```diff
@@ -656,7 +661,18 @@ export interface CoreState {
...
+    Required<
+      Pick<
+        MasterportalApiConfig,
+        | 'backgroundImage'
+        | 'epsg'
+        | 'namedProjections'
+        | 'options'
+        | 'startResolution'
+      >
+    >
```
warm-coolguy, line 675: This reads as if e.g. backgroundImage was required, but didn't you explicitly make it optional?

warm-coolguy (changes requested):
🏓 @dopenguin

packages/core/README.md
dopenguin: This has always been optional, the documentation was just wrong

packages/core/README.md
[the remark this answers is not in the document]
dopenguin: I've adjusted this to match the default values 21294279

packages/core/README.md
```diff
@@ -253,48 +313,102 @@ Whitelisted and confirmed parameters include:
...
+          '3': 'Natural monument',
+          '2': 'Water bodies',
+          '1': 'Architectural monument (area)',
+          '0': 'Natural monument (area)',
+        },
+        legend: true,
+      },
+    },
+  },
+]
```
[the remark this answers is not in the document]
dopenguin: Removed 1359f171

packages/core/README.md
```diff
@@ -253,48 +313,102 @@ Whitelisted and confirmed parameters include:
...
+          '25': 'Material group',
+          '4': 'Architectural monument',
+          '3': 'Natural monument',
+          '2': 'Water bodies',
+          '1': 'Architectural monument (area)',
+          '0': 'Natural monument (area)',
+        },
+        legend: true,
+      },
+    },
```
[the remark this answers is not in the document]
dopenguin: LayerChooser already has such an example, removed it here 1359f171

packages/core/README.md
```diff
@@ -253,48 +313,102 @@ Whitelisted and confirmed parameters include:
...
 | - | - | - |
-| epsg | string | Leading coordinate system, e.g. `"EPSG:25832"`. |
-| extent | number[] | Map movement will be restricted to this rectangle. |
-| namedProjections | Array | Array of usable projections by proj4 string. |
-| options | Array | Defines all available zoomLevels. Entries define `resolution`, `scale`, and `zoomLevel`. See masterportalAPI for details. |
-| startCenter | number[] | Initial center coordinate. |
-| startResolution | number | Initial resolution; must be in options. See below. |
+| id | string | Service register id in `mapConfiguration.layerConf`. |
+| name | string | Display name in UI. |
+| type | enum["background", "mask"] | Layer handling. Backgrounds are mutually exclusive, masks ("overlays") can be stacked. |
```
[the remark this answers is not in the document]
dopenguin: Moved to LayerChooser df4a62b6

packages/core/README.md
dopenguin: Just looking at the docs ([services.json.md](https://bitbucket.org/geowerkstatt-hamburg/masterportal/src/dev/doc/services.json.md) and [config.json.md](https://bitbucket.org/geowerkstatt-hamburg/masterportal/src/dev/doc/config.json.md#markdown-header-themenconfighintergrundkarten)) of Masterportal, `minZoom` and `maxZoom` probably are LayerChooser specific, as `maxScale` and `minScale` is used. `visibility` seems to be usable directly though.

I've moved both to LayerChooser though, just to be safe, as I didn't find any code references in the core 71e09ee2

packages/core/README.md
dopenguin: Interestingly enough, Masterportal has switched to defaulting to no backgroundImage, thus this is just the default of `@masterportal/masterportalapi`.
I've just added this here for completion.

I'm not quite sure what I am supposed to change here. Could you elaborate?

packages/core/src/utils/createMap/index.ts
dopenguin: Done d3013524

packages/lib/testMountParameters/index.ts
dopenguin: Exposed it as another export 97818b77

packages/types/custom/CHANGELOG.md
dopenguin: No longer the case 97818b77

dopenguin:
🏓 @warm-coolguy 

I'll have to move the defaults to a separate lib probably, as the tests are now failing because they now have some code from `@polar/core` which results in the tests failing ._.

packages/core/README.md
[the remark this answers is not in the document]
warm-coolguy: The example is still confusing. There's still a memorial service referenced, and there's undeclared variables used in the `id` slots.

packages/core/README.md
warm-coolguy: I'm not sure how to handle this, but now effectively something's documented that doesn't work. I'd either skip documenting it at all, or propose it's fixed (or it's noted somewhere that it needs fixing). However, I remember 0 instances of anyone needing this backgroundImage, so I guess it's safe to drop the doc.

warm-coolguy:
> 🏓 @warm-coolguy
> 
> I'll have to move the defaults to a separate lib probably, as the tests are now failing because they now have some code from `@polar/core` which results in the tests failing ._. If you have a better solution to just fix the problem here, let me know

🏓 @dopenguin 

Seems to be an issue in our babel setup, since `import.meta` is not available. I found some solutions online, but they did not quite work or produced other issues – googling "jest + ${the error message}" produces a lot of solutions, if you want to try that, too.

Another solution would be using `import mpapiDefaults from '@polar/core/src/utils/createMap/defaults'` in `lib/testMountParameters/index.ts`. You may not like it, but this is what peak tree shaking looks like.

packages/core/README.md
dopenguin: Removed it be841eaf642863a410e8a3747cbcddefbb09abe5

packages/core/README.md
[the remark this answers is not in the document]
dopenguin: Adjusted 6e7bb737cb63ff8154c182849fd82c99ddab7681

dopenguin:
> > 🏓 @warm-coolguy
> > I'll have to move the defaults to a separate lib probably, as the tests are now failing because they now have some code from `@polar/core` which results in the tests failing ._. If you have a better solution to just fix the problem here, let me know
> 
> 🏓 @dopenguin
> 
> Seems to be an issue in our babel setup, since `import.meta` is not available. I found some solutions online, but they did not quite work or produced other issues – googling "jest + ${the error message}" produces a lot of solutions, and I only scratched the surface; if you want to try that, too.
> 
> Another solution would be using `import mpapiDefaults from '@polar/core/src/utils/createMap/defaults'` in `lib/testMountParameters/index.ts`. You may not like it, but this is what peak tree shaking looks like.

It feels like, like that goes against [this ADR](https://github.com/Dataport/polar/wiki/Architecture-Decision-Records-%F0%9F%93%83#2024-03-05-how-to-expose-additional-exports-of-a-package), but I've went along for now and have added a new internal issue regarding this. f300533ea783d316ec0a8e57fc7ba3f7909911f9

🏓 @warm-coolguy

warm-coolguy (approved):
🚀 @dopenguin when CHANGELOG.md conflict is resolved

PR #161, opened 2024-08-20
markdown heading IDs

ludoholic:
## Summary

Headings in markdowns will get an ID during docs generation

- `node ./scripts/makeDocs.js ./snowbox`
- http://localhost:63342/polar2/packages/clients/snowbox/docs/core.html#addplugins


dopenguin (approved):
🚀 @ludoholic

PR #165, opened 2024-08-28
Feature/dark mode

Softwarecules:
## Summary

Adds the possibility for a Dark Theme in POLAR that Clients can use in future. Some plugin-backgrounds were updated from a fixed background-color to CSS-Variables.

Run the `Snowbox` client and click on the `Toggle Theme` Button above the Map. It should switch between Light and Dark Theme of the components within the map.

(!) Known Issue: 
The GFI-Button-Group will not change colors with the `Toggle Theme` Button. 
They will however adapt if you set `dark: true` in the `vuetify.ts` within the `theme`-object, save, and reload the site.
It might be fixed with the Vuetify `options: {customProperties: true}`, which has caused other issues however.


packages/core/README.md
```diff
@@ -351,6 +351,33 @@ Additionally to the regular fields, `primaryContrast` and `secondaryContrast` ar
...
+      },
+      dark: {
+        primary: 'white',
+        primaryContrast: 'black',
+        secondary: '#de1e7e',
+        secondaryContrast: '#c0ffee',
+      },
+    },
+  },
+```
```
dopenguin, line 375: As you have added default values for `dark`, this section is misleading, as it would even work without such configuration in the client.

Simply add an example for the `themes.dark` part in the above example.

packages/core/README.md
```diff
@@ -351,6 +351,33 @@ Additionally to the regular fields, `primaryContrast` and `secondaryContrast` ar
...
+
+The theme can either be activated directly by setting `dark: true` within the `theme`-object or by calling the `setTheme` method within the store.
```
dopenguin, line 377:
```suggestion
```
I don't think that this is necessary. If someone wants to have their client in dark mode from the getgo, they'll just look it up in the vuetify configuration. If they want to change it later, they can look at the mutation description

packages/core/README.md
```diff
@@ -375,3 +402,28 @@ map.$store.watch(
...
+| theme | enum['light', 'dark'] | By default the theme is set to 'light'. Setting theme to 'dark' will render the default 'dark' theme. |
+
+### Actions
+
+```js
+changeTheme({ commit }, theme) {
+        commit('setTheme', theme)
+        recolor(theme)
+      },
+```
```
dopenguin, line 429:
```suggestion
```
`snowbox` uses the `mutation` and the pattern you implemented wants the `mutation` to be the primary function to be called by users. So I don't think we should document this.

packages/core/README.md
```diff
@@ -375,3 +402,28 @@ map.$store.watch(
...
+### Mutations
+
+#### setTheme
+
+This can be used to change the theme.
+
+```js
+setTheme(state, theme) {
+        state.theme = theme
+      },
```
dopenguin, line 415:
```suggestion
client.$store.commit('setTheme', 'dark')
```
We don't document how the mutation is written but how it can be used

packages/plugins/Scale/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Breaking: Upgrade peerDependency `ol` from `^7.1.0` to `^9.2.4`.
 - Feature: Add the option to configure a scale switcher via the new configuration parameters `showScaleSwitcher` and `zoomMethod`.
+- Update: fix background-colors to use CSS variables.
```
dopenguin, line 7:
```suggestion
- Fix: Change background colors to use CSS variables to be able to override these values via the `vuetify` configuration.
```

dopenguin (changes requested):
🏓  @Softwarecules 

I was not able to reproduce the known issue. Is it fixed by any chance? :D

packages/core/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 - Fix: Adjust documentation to properly describe optionality of configuration parameters.
 - Fix: Add package `events` as a dependency to fix issue with `xml2js`. See https://github.com/Leonidas-from-XIV/node-xml2js/issues/697 for more information.
 - Chore: Update dependencies to latest versions.
+- Feature: Add dark theme support and updated fix background-colors to use CSS variables.
```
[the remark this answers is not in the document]
Softwarecules: But in that suggestion there's still Feature - Fix - Fix - Chore - Fix ? 🤔  (Line 12.)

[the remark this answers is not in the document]
dopenguin: Changed that!

packages/core/README.md
Softwarecules: 🏓 @dopenguin

dopenguin:
PR closed after internal discussion.

PR #168, opened 2024-08-30
change text locator build setup

warm-coolguy:
## Summary

The distributable now includes the configuration as .js file, so it's no longer hidden in the bundle after building. The README has been modified to reflect on this change. This has been done so that the deployment for different environments works without additional build steps.

After this is merged, an alpha release is due.

- Dev mode should still work as previously
- Prod mode is now configurable after building


packages/clients/textLocator/README.md
```diff
@@ -8,4 +8,6 @@ Please see the CHANGELOG.md for all changes after the initial release.
 
 ## Usage
 
-The product is a hostable HTML page.
+The product is a hostable HTML page. The included index.js file contains the default configuration.
+
+It's also possible to directly import and use the client `client-text-locator.js`; however, this is currently neither tested nor documented. Should the client with matching documentation be required in such a way that further modifications can be made or that the client can used embedded in a page, contact us.
```
dopenguin, line 13: Should we leave the support mail here as well?

packages/clients/textLocator/package.json
```diff
@@ -10,13 +10,15 @@
...
+  "main": "dist/client-text-locator.js",
   "files": [
     "dist/**/**.*",
     "CHANGELOG.md"
   ],
   "scripts": {
     "postversion": "npm run build",
-    "build": "rimraf dist && vite build && bash ./scripts/injectFlag.sh"
+    "build": "rimraf dist && vite build && copyfiles -f src/html/**/* dist",
+    "build:test": "rimraf dist && vite build && bash ./scripts/injectFlag.sh"
```
dopenguin, line 21: This seems to be the only client that has a `build:test` script.

Is the `injectFlag.sh` actually still needed? Refresh my memory please :D

packages/clients/textLocator/src/html/index.js
dopenguin, line 1: IMO the content of this file simply fits in `packages/clients/textLocator/src/html/index.html` similar how `meldemichel` handles this.

After reading `packages/clients/textLocator/src/index.html` I would even go so far to question why `packages/clients/textLocator/src/html/index.html` exists.
Wouldn't it be simpler to just have a script that changes the import `./polar-client.ts'` to `./client-text-locator.js`?

dopenguin (changes requested):
🏓 @warm-coolguy

packages/clients/textLocator/README.md
warm-coolguy: Nah, the direct package consumer is internal and knows how to reach out to us.

packages/clients/textLocator/package.json
warm-coolguy: Forgot to clean up. https://github.com/Dataport/polar/pull/168/commits/dd6ae5368f95b2eac4852e2d4951f1fd53d79b57 Apparently `injectFlag.sh` is not needed as `build:test` was called nowhere.

packages/clients/textLocator/src/html/index.js
warm-coolguy: It was an explicit requirement to pull the config out as a .js file.

Using the index.html as entry point for a build will wrangle any .js files/contents to something like `<script type="module" crossorigin src="./assets/index-C20EeCK4.js"></script>`, see DISH build as an example where it happens (since it doesn't matter there).

Hence I copied how it's done in the Meldemichel.

A third way with an index.html that's only the entry point in dev mode and copied/edited on builds where the TS file is the entry point is a solution that produces less files, but also is more convoluted IMO due to hiding information in modifying scripts.

I'd argue it doesn't matter in which way it's confusing, but at least it's consistent with how Meldemichel `COMPLETE` mode is built to work, so there's one way less to understand.

warm-coolguy:
> 🏓 @warm-coolguy

🏓 @dopenguin

dopenguin (approved):
🚀 @warm-coolguy

PR #173, opened 2024-09-05
Feature/remove deprecated behaviour

dopenguin:
## Summary

Remove behviour marked as deprecated as we are releasing a new major version of each respective package.

Test if one of the three features is still only being used in its deprecated variation.

I found `data-polar` still being present in `dish` and `textLocator` via the `injectFlag.sh` script. Is it still needed?

#149 has to be merged before this branch


warm-coolguy (changes requested):
🏓 @dopenguin 

>I found `data-polar` still being present in `dish` and `textLocator` via the `injectFlag.sh` script. Is it still needed?

It's not needed in TextLocator anymore after merging https://github.com/Dataport/polar/pull/168.

However, DISH still needs it, unless you apply a refactor in the same spirit. If you want to touch that issue in depth, we might 📞 and make up a way on how to solve this in an uniform fashion for Meldemichel, DISH, and TextLocator.

warm-coolguy (approved):
🚀 @dopenguin

PR #174, opened 2024-09-11
fix ts client check

warm-coolguy:
## Summary

*looks at ts client check setup*

I can fix it.

Pipeline should now run through and tsc also checks clients. I've fixed the setup in that I added (mostly unnecessarily) default configuration properties (that will still throw errors in the plugins if not overridden, so no harm done here) and adjusted some types. I've also entertained `any` to resolve some of the problems, where both `unknown` and `never` would result in errors in varying positions. While not quite ideal, this is imo sufficient by the 20-80 rule.

The biggest gain here is that we've now enabled type checking for clients so that we can stop preventable issues from sneaking in.


dopenguin (approved):
🚀  @warm-coolguy

PR #175, opened 2024-09-12
Fix pages:build on MacOS

dopenguin:
## Summary

Like the title says, this change fixed the `pages:build` script on MacOS systems.

Check if it still works for you on Windows with `npm run pages:build && npm run pages:build:serve`


scripts/buildPages.sh
```diff
@@ -3,7 +3,11 @@ for i in "${array[@]}"
 do
   echo "Building $i docs ..."
 	npm run docs:$i
-  mkdir --parents ./pages/docs/$i
+  if [[ "$OSTYPE" == "darwin"* ]]; then
+    mkdir -p ./pages/docs/$i
+  else
+    mkdir --parents ./pages/docs/$i
+  fi
```
warm-coolguy, line 10:
```suggestion
  mkdir -p ./pages/docs/$i
```
We can drop the OS check, `-p` also works on git bash on Windows.

warm-coolguy (changes requested):
🏓 @dopenguin

dopenguin:
> 🏓 @dopenguin

🏓 @warm-coolguy

warm-coolguy (approved):
🥇 @dopenguin

PR #178, opened 2024-09-16
allow selection escape with escape button

### LINKED ISSUE #110: Close address search on Esc press
I've noticed a few times that I tried to close the AddressSearch by pressing Esc. We should probably implement it for users expecting this.

- Don't forget Fullscreen Mode and other Esc interactions on implementation.
  - Suggestion: Use event and stop propagation/default if usable, else propagate.

Alternative solution: Make the X button on the search input accessible to keyboards.

warm-coolguy:
## Summary

When navigating the result selection via keyboard, pressing Escape will now let users close the result list. The focus will then be in the same place as if the result list was closed due to an actual selection.

I'm rather unhappy with the solution as, in Fullscreen mode, the plugin will take precedence and use the escape signal first, stopping its progression. Also, I couldn't make the `clear` button available, since that is a bug fixed in vuetify@3, where migration is not trivial, and it seems they only adressed that in the next major version. If you care:

- https://github.com/vuetifyjs/vuetify/issues/18482
- https://github.com/vuetifyjs/vuetify/pull/18838

Hence, this minimal and IMO okayish solution. At least everything can be done via keyboard, even if not in an ideal fashion.

Check all clients with AddressSearch.

Resolves #110


dopenguin (approved):
🚀 @warm-coolguy

PR #179, opened 2024-09-16
#33 refactor code to match lint rules

warm-coolguy:
## Summary

Utilized utils, undoing unfortunate underuse of useful utility.

(Pulled code to an util so it now fits the linting rules.)

Check whether DISH still works regarding the WMS layer options menu.

Relates #33.


dopenguin (changes requested):

🏓 @warm-coolguy

warm-coolguy:
> In the past, we've added information regarding refactorings to the Changelog. I think some relevant entries should be added here as well
> 
> 🏓 @warm-coolguy

🏓 @dopenguin 

https://github.com/Dataport/polar/pull/182#issuecomment-2375991845

dopenguin (approved):
🚀 @warm-coolguy

PR #180, opened 2024-09-16
refactor method (linting, readability)

warm-coolguy:
## Summary

Improve readability of export implementation. Remove linting warning.

Fiddle with export tool in snowbox to check some combinations. Snippet as reminder:

```json
{
  export: {
    showPng: true,
    showJpg: true,
    showPdf: true,
    download: true,
  }
}
```

#33

### DIFF
--- a/packages/plugins/Export/src/store/actions.ts
+++ b/packages/plugins/Export/src/store/actions.ts
@@ -1,6 +1,91 @@
 import { jsPDF } from 'jspdf'
 import { PolarActionTree } from '@polar/lib-custom-types'
+import { Map } from 'ol'
 import { ExportFormat, ExportGetters, ExportState } from '../types'
+
+// PDF options
+const dims = {
+  a0: [1189, 841],
+  a1: [841, 594],
+  a2: [594, 420],
+  a3: [420, 297],
+  a4: [297, 210],
+  a5: [210, 148],
+}
+
+// Screenshot canvas
+const CANVAS_ID = 'export-canvas'
+
+const convertToPdf = (map: Map, src: string, size: number[]) => {
+  // NOTE: when supporting more formats, scale map accordingly
+  const format = 'a4'
+  const dim = dims[format]
+  // Import of jspdf is in mounted.
+  const jsPdf = new jsPDF('landscape', undefined, format) // eslint-disable-line
+  jsPdf.addImage(src, 'JPEG', 0, 0, dim[0], dim[1])
+  // Reset original map size
+  map.setSize(size)
+  map.getView().setResolution(map.getView().getResolution())
+
+  return {
+    pdfSrc: jsPdf.output('datauristring'),
+    jsPdf,
+  }
+}
+
+const downloadAsImage = (src: string, type: ExportFormat) => {
+  const link = document.createElement('a')
+  link.download = 'map.' + (type === ExportFormat.PNG ? 'png' : 'jpeg')
+  link.href = src
+  document.body.appendChild(link)
+  link.click()
+  document.body.removeChild(link)
+}
+
+// For every ol-layer, get matrix and apply on canvas
+const getDataUrl = (size: number[], type: ExportFormat) => {
+  // The canvas to render the screenshot in
+  const mapCanvas: HTMLCanvasElement = document.createElement('canvas')
+  mapCanvas.id = CANVAS_ID
+  // mapCanvas.setAttribute('crossOrigin', 'anonymous')
+  mapCanvas.width = size[0]
+  mapCanvas.height = size[1]
+
+  // Write on this
+  const mapContext = mapCanvas.getContext('2d')
+  if (!mapContext) {
+    console.error(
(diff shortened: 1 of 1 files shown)


packages/plugins/Export/src/store/actions.ts
```diff
@@ -1,7 +1,92 @@
...
+  mapCanvas.id = CANVAS_ID
+  // mapCanvas.setAttribute('crossOrigin', 'anonymous')
+  mapCanvas.width = size[0]
+  mapCanvas.height = size[1]
+
+  // Write on this
+  const mapContext = mapCanvas.getContext('2d')
+  if (!mapContext) {
+    console.error(
+      '@polar/plugin-export: map does not of a 2d context, export failed.'
```
dopenguin, line 58:
```suggestion
      '@polar/plugin-export: map does not have a 2d context, export failed.'
```
Let's fix this while we are at it

packages/plugins/Export/src/store/actions.ts
```diff
@@ -1,7 +1,92 @@
...
+          .match(/^matrix\(([^(]*)\)$/)[1]
+          .split(',')
+          .map(Number)
+        // Apply the transform to the export map context
+        CanvasRenderingContext2D.prototype.setTransform.apply(
+          mapContext,
+          matrix
+        )
+        mapContext.drawImage(canvas, 0, 0)
+      } else console.warn('@polar/plugin-export: canvas width is 0')
```
dopenguin, line 81:
```suggestion
      } else console.warn('@polar/plugin-export: canvas width is 0.')
```

packages/plugins/Export/src/store/actions.ts
```diff
@@ -12,98 +97,23 @@ const actions: PolarActionTree<ExportState, ExportGetters> = {
...
       if (type === ExportFormat.JPG || type === ExportFormat.PNG) {
-        if (download) {
-          const link = document.createElement('a')
-          link.download = 'map.' + (type === ExportFormat.PNG ? 'png' : 'jpeg')
-          link.href = src
-          document.body.appendChild(link)
-          link.click()
-          document.body.removeChild(link)
-        }
+        if (download) downloadAsImage(src, type)
```
dopenguin, line 111: As I can' suggest it (deleted lines and such), here is what I would've anticipated:

```ts
if (
  (type === ExportFormat.JPG || type === ExportFormat.PNG) &&
  download
) {
  downloadAsImage(src, type)
```

packages/plugins/Export/src/store/actions.ts
```diff
@@ -1,7 +1,92 @@
...
+const convertToPdf = (map: Map, src: string, size: number[]) => {
+  // NOTE: when supporting more formats, scale map accordingly
+  const format = 'a4'
+  const dim = dims[format]
+  // Import of jspdf is in mounted.
+  const jsPdf = new jsPDF('landscape', undefined, format) // eslint-disable-line
+  jsPdf.addImage(src, 'JPEG', 0, 0, dim[0], dim[1])
+  // Reset original map size
+  map.setSize(size)
+  map.getView().setResolution(map.getView().getResolution())
```
dopenguin, line 28: There must've been a reason for the variable beforehand. Are you sure, this doesn't change anything? `map.getSize()` is still set as a variable on the other hand.

Now that I think about it: What happens if neither the size of the map or the resolution is reset?

dopenguin (changes requested):
🏓 @warm-coolguy

packages/plugins/Export/src/store/actions.ts
warm-coolguy: That would be a change in meaning. The else branch would be taken under different circumstances than it is now.

packages/plugins/Export/src/store/actions.ts
warm-coolguy: Upon revisiting this, yeah, it doesn't change anything. In fact, we can even skip the line altogether. I've dropped the whole map view modification thing in https://github.com/Dataport/polar/pull/180/commits/b0edcaf1b2475f2c298169a1d0abbcb611f5de76.

After all, size and resolution of the map are never changed anyway. It's the same on `main`.

warm-coolguy:
> In the past, we've added information regarding refactorings to the Changelog. I think some relevant entries should be added here as well 🏓 @warm-coolguy

🏓 @dopenguin with nod in https://github.com/Dataport/polar/pull/182#issuecomment-2375991845's direction.

packages/plugins/Export/src/store/actions.ts
dopenguin: It would have a change in meaning, but I highly doubt we'll ever have additional features in this if-clause. You can 🎩 though, if you want.

dopenguin (commented):
🏓 @warm-coolguy

warm-coolguy:
> 🏓 @warm-coolguy

🏓 @dopenguin

dopenguin (approved):
🚀 @warm-coolguy

PR #182, opened 2024-09-16
#33 refactor AddressSearch to meet linting requirements

warm-coolguy:
## Summary

Refactored AddressSearch to meet linting requirements.

Check BKG and mpapi implementations on whether the behaviour has changed.

#33


dopenguin (changes requested):

🏓 @warm-coolguy

warm-coolguy:
> In the past, we've added information regarding refactorings to the Changelog. I think some relevant entries should be added here as well
> 
> 🏓 @warm-coolguy

I propose we drop the changelog requirement for pure refactorings that have no visible effect to any consumer.

🏓 @dopenguin

dopenguin (approved):
> have no visible effect to any consumer

Sure, but let's keep an eye out on a per-change-basis

🚀 @warm-coolguy

PR #183, opened 2024-09-16
#33 refactor unused code by deleting it

warm-coolguy:
## Summary

There was gazetteer code here that was not in use.

It also had linting warnings.

It's gone now.

Check if I got this right.

#33


dopenguin (approved):
🚀 @warm-coolguy

PR #186, opened 2024-09-19
Update/update eslint to major 9

warm-coolguy:
## Summary

Update eslint to major 9, and friends.

Just check everything. This required a LOT of changes.


warm-coolguy:
Just made this quickly to have a "Files required" tab. I don't think this one is worth the effort, since the ideal state when we're done is having the exact same behaviour as previously, and it's just a dev tool.

PR #188, opened 2024-09-26
Additionally expose createMap for @polar/client-generic

dopenguin:
## Summary

As the title says, `@polar/client-generic` now additionally exposes `createMap` to be able to use it directly.

`npm run pages:build && npm run pages:build:serve`. This should then also work in the pipeline.


warm-coolguy (approved):
🏓 @dopenguin

PR #203, opened 2024-11-19
Fix/geojson

dopenguin:
## Summary

When testing some GeoJSON I've realized some issues with my previous implementation which are being fixed here.
- `format` is not needed on GeoJSON
- Fix an issue if a value of a property is not a string
- Fix implementation for GeoJSON gfi requests using a coordinate

Meldemichel and snowbox should be tested


warm-coolguy (approved):
@dopenguin 🚀 LGTM

Issue #204, opened 2024-11-25
Assigning parameter "crossOrigin = anonymous" by monkey patch does not work correctly for ol class ImageWMS


jedi-of-the-sea:
The original problem (described below) can be solved if the parameter `crossOrigin: 'anonymous'` is set in the layer config for the layer that is configured with `singleTile: true`

When a WMS is loaded as single Image, it has a different ol class: https://openlayers.org/en/latest/apidoc/module-ol_source_ImageWMS-ImageWMS.html
In the ol documentation it is pointed out that the layer has to be configured with the parameter `crossOrigin` is one wants to access pixel data with the Canvas renderer.

While singleTile=true is not officially supported by POLAR, every layer added to the map should get the parameter `crossOrigin: 'anonymous'` by default. Since this does not work correctly, the bug ticket is still relevant.

---------------------------------
### **POLAR export plugin cannot handle layers with singleTile=true configured**

**Describe the bug**
When trying to create a screenshot, the export plugin throws errors in the browser console (see screenshots below)

This behaviour was recreated with four different layers (e.g. basemap WMS) that had the configuration singletile =true.

**Expected behavior**
When activating the button for the export plugin, a screenshot is taken from the map for downloading or copying.

**Steps to reproduce**

1. Use a client with a WMS that is configured as singleTile=true and uses the export plugin
2. Make sure that the WMS has been already loaded (otherwise the error does not occur)
3. Try to make a screenshot by clicking the export button 
4. Watch the browser console for error message

**Screenshots**
![image](https://github.com/user-attachments/assets/7abb5562-6b35-44bb-a82e-2fb7eafb6c83)
![image](https://github.com/user-attachments/assets/406dfcad-2c02-4a96-8764-b0d5a22c443a)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Firefox, Edge, Chrome
 - Version:  
            - Firefox: 132.0.2
            - Edge: 131.0.2903.63
            - Chrome: 131.0.6778.86

**Smartphone (please complete the following information):**
- not tested on Smartphone

**Additional context**
Was tested in dev environment and on test server

jedi-of-the-sea:
The problem can be solved if the parameter `crossOrigin: 'anonymous'` is set in the layer config for the layer that is configured with `singleTile: true`
Maybe it is enough to update documentation with hints for those cases to prevent this error. 

[...]

PR #206, opened 2024-11-29
Fix/assign cross origin param for image wms

jedi-of-the-sea:
## Summary

The private parameter `crossOrigin` of the ol class ImageWMS is now set to `anonymous` to prevent errors in canvas rendering.

When configuring a WMS with `singleTile: true` the WMS is loaded as an ImageWMS. This ol class has the private parameter crossOrigin which has not been effected by the previous method to set this parameter. If a WMS is configured like described, the Export Plugin which uses canvas rendering throws an error. 

- To recreate this error, start a client which uses the Export plugin and a WMS that is configured with the parameter `singleTile: true`.
- Click the export Button.
- The export does not work. See browser console for error.

#204

### DIFF
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## unpublished
+
+- Fix: Set crossOrigin for ImageWMS to `anonymous` to prevent error in canvas rendering (e.g. when using export plugin).
+
 ## 2.0.0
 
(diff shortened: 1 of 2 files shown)


dopenguin (changes requested):
🏓 @jedi-of-the-sea

jedi-of-the-sea:
🏓 @dopenguin

dopenguin (approved):
🚀 @jedi-of-the-sea

PR #209, opened 2024-12-16
Fix/scale switcher tooltip

jedi-of-the-sea:
## Summary

The tooltip for scale switcher was showing only the translation key. So the missing translation for scale switcher was added.

Start a portal with a configured scale switcher, e.g. start the dish client after adding
```
scale: {
  showScaleSwitcher: true,
  zoomMethod: 'plugin/zoom/setZoomLevel',
}
```
to mapConfig.ts

Then hover with mouse over scale switcher and check if translation is shown.

### DIFF
--- a/packages/clients/dish/CHANGELOG.md
+++ b/packages/clients/dish/CHANGELOG.md
@@ -6,7 +6,6 @@
 - Fix: Extend typing for search result function according to type package update.
 - Fix: Import types `AddressSearchState` and `AddressSearchGetters` from correct position.
 - Fix: Import enum `SearchResultSymbols` from correct position.
-- Fix: Add translations for scale switcher.
 - Chore: Change value of `pins.movable` configuration to `'drag'` as using a boolean has been deprecated in a future release.
 - Chore: Upgrade `@masterportal/masterportalapi` from `2.8.0` to `2.40.0` and subsequently `ol` from `^7.1.0` to `^9.2.4`.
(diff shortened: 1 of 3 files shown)


packages/clients/dish/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Fix: Extend typing for search result function according to type package update.
 - Fix: Import types `AddressSearchState` and `AddressSearchGetters` from correct position.
 - Fix: Import enum `SearchResultSymbols` from correct position.
+- Fix: Add translations for scale switcher.
```
dopenguin, line 9: This changelog entry should be added to `@polar/plugin-scale`. Also, what do you think about "Add missing translations for the scale switcher."?

dopenguin (changes requested):
🏓 @jedi-of-the-sea

packages/clients/dish/CHANGELOG.md
jedi-of-the-sea: ahhh I knew that >___< I must have mixed that up because of working on the dish client for too long.. 

Changed it: 48123d6a496b1519f9cc24ffd5ca296250ab0d56

jedi-of-the-sea:
🏓 @dopenguin

dopenguin (approved):
🚀 @jedi-of-the-sea

PR #211, opened 2025-01-06
Extend multi selection by adding new mode "circle"

dopenguin:
## Summary

Add the possibility to select multiple features via a circle instead of a box.

This also deprecates the configuration parameter `boxSelect` in favour of the new parameter `multiSelect` which can be set to `'box'` or `'circle'`.

`npm run snowbox` and use CTRL / CMD and click once to start the circle and a second time to finish the selection.

#124 has to be merged before this


packages/plugins/Gfi/README.md
```diff
@@ -16,13 +16,14 @@ The GFI plugin can be used to fetch and optionally display GFI (GetFeatureInfo)
...
 | mode | enum["bboxDot", "intersects"]? | Method of calculating which feature has been chosen by the user. `bboxDot` utilizes the `bbox`-url parameter using the clicked coordinate while `intersects` uses a `Filter` to calculate the intersected features. Layers can have their own `gfiMode` parameter which would override this global mode. To apply this, add the desired value to the parameter in the `mapConfiguration`. Defaults to `'bboxDot'`. |
+| multiSelect | enum["box", "circle"]? | If configured, multiple features can be selected at once by using the modifier key (CTRL on Windows or Command on macOS) and dragging the mouse. If set to `'box'`, the selection will be done in a box. If set to `'circle'`, the selection will be done in a circle beginning from the first click. In this case, the user has to click a second time to end the selection. Similar to `gfi.directSelect`, features can be added and removed by selection / unselecting them. The features need to be distinguishable by their properties for the functionality to properly work. Does not work together with `extendedMasterportalapiMarkers` of `@polar/core`. Is disabled by default. |
```
warm-coolguy, line 26: While I understand that this feature is mostly used by clerks in an office/desktop environment, it lacks mobile support and is thus technically inaccessible on some devices. I wonder how we should treat that.

Note in documentation that it is meant for desktop use only and should not be used in citizen-facing clients? Maybe even make an issue of it, albeit that may be overzealous since afaik nobody needs this feature so far.

packages/plugins/Gfi/src/store/actions/setupMultiSelection.ts
```diff
@@ -0,0 +1,91 @@
...
+      dispatch('getFeatureInfo', {
+        // @ts-expect-error | A feature that is drawn has a geometry.
+        coordinateOrExtent: e.feature.getGeometry().getExtent(),
+        modifierPressed: true,
+      }).finally(() =>
+        // Needed so that no pin is set when finishing drawing the circle
+        setTimeout(() => {
+          // @ts-expect-error | internal hack to detect it in @polar/plugin-pins
+          circleDraw._isMultiSelect = false
+        }, 250)
```
warm-coolguy, line 72: Does 0 suffice since all we need to do is get at the end of the execution chain or do we actually need the 250ms as timeout until an asynchronous request is done? In the latter case, I do not consider this a valid solution. We'd probably need a pins getter that returns an awaitable method that resolves on completing all current loading operations.

packages/plugins/Gfi/src/store/actions/setupMultiSelection.ts
```diff
@@ -0,0 +1,91 @@
...
+import { PolarActionContext } from '@polar/lib-custom-types'
+import { GfiGetters, GfiState } from '../../types'
+
+const circleDraw = new Draw({
+  source: new VectorSource(),
+  type: 'Circle',
+  style: new Style({
+    stroke: new Stroke({ color: '#118bee', width: 2 }),
+    fill: new Fill({ color: [17, 139, 238, 0.6] }),
+  }),
```
warm-coolguy, line 15: I'd prefer the multiSelect circle to look more like the multiSelect dragBox.

packages/plugins/Gfi/src/store/actions/setupMultiSelection.ts
```diff
@@ -0,0 +1,91 @@
...
+      )
+    }
+    dragBox.on('boxend', () =>
+      dispatch('getFeatureInfo', {
+        coordinateOrExtent: dragBox.getGeometry().getExtent(),
+        modifierPressed: true,
+      })
+    )
+    map.addInteraction(dragBox)
+  } else if (multiSelect === 'circle') {
```
warm-coolguy, line 57: I'd very much like the circle to behave as the box does in that the dragging operation is continued while the mouse is held down and ends on the `pointerup` or `pointercancel` event, since that is the "default OL behaviour" in the DragBox and thus probably expectable here, too. Also I fathom this to be the more mobile-like user interaction, should we ever come to the point where users may drag on mobile devices by e.g. clicking a button placed somewhere beforehand.

packages/plugins/Pins/src/store/index.ts
```diff
@@ -53,8 +53,8 @@ export const makeStoreModule = () => {
         rootGetters.map.on('singleclick', async ({ coordinate }) => {
           const isDrawing = interactions.getArray().some(
             (interaction) =>
-              // these indicate other interactions are expected now
-              interaction instanceof Draw ||
+              // @ts-expect-error | internal hack to detect it from @polar/plugin-gfi
+              (interaction instanceof Draw && interaction._isMultiSelect) ||
```
warm-coolguy, line 57: This should be `or`ed instead of being `and`ed. Currently, while `Draw` is active, the marker is e.g. placed on the center of any circle drawn.

warm-coolguy (changes requested):
🏓 @dopenguin

packages/plugins/Gfi/src/store/actions/setupMultiSelection.ts
dopenguin: IIRC I tested it with 0 or time lower than 250ms. I'll see what I can do to have a cleaner solution

packages/plugins/Gfi/src/store/actions/setupMultiSelection.ts
dopenguin: Sure! I'll change the styling

packages/plugins/Gfi/src/store/actions/setupMultiSelection.ts
dopenguin: `Draw` offers a configuration parameter `stopClick` 0c38d86a95c04d600c35febe3ebcf96e2bb0971e

packages/plugins/Pins/src/store/index.ts
dopenguin: The problem with simply adding an `or` here is that then the Pin can never be placed. I am not quite sure on how to handle this and will take a closer look at the end, maybe with a clearer idea

packages/plugins/Gfi/src/store/actions/setupMultiSelection.ts
dopenguin: I've fiddled around a bit and I don't seem to find a configuration of [Draw](https://openlayers.org/en/v9.2.4/apidoc/module-ol_interaction_Draw-Draw.html) that allows such interaction possibilities.

I've also realized that [DragBox](https://openlayers.org/en/v9.2.4/apidoc/module-ol_interaction_DragBox-DragBox.html) is not implemented for mobile usage. If we intend to support that, we'd have to either extend OL directly or implement something ourselves

packages/plugins/Gfi/README.md
dopenguin: Also, see https://github.com/Dataport/polar/pull/211#discussion_r1922193778

packages/plugins/Gfi/src/store/actions/setupMultiSelection.ts
dopenguin: Adjusted ea992a6ea5c7b226dc82aa6008cbfbb4dc2b668f

packages/plugins/Pins/src/store/index.ts
dopenguin: Found a solution 9170d0fccd31bf1b3fc16c11cf50c20440decd38

packages/plugins/Gfi/README.md
dopenguin: Adjusted documentation fef2366975e05ef55267b37d865c3dc22a6a8bb7

dopenguin:
> 🏓 @dopenguin

🏓 @warm-coolguy

packages/plugins/Gfi/src/store/actions/setupMultiSelection.ts
warm-coolguy: What do you think about trashing the DragBox then, and instead going for a box draw analoguous to the circle draw? https://openlayers.org/en/latest/examples/draw-shapes.html That, too, would even out the drawing experience.

packages/plugins/Pins/src/store/index.ts
warm-coolguy: No crucially different idea comes to mind, and obviously we've had that problem before ...

Would you please write an issue/ticket regarding this? While I don't reckon there'll be a wildly different solution, we can at least standardize and document the flag for future plugins. I'm e.g. thinking about a `draw._isHeldByPolarPlugin = "Measure"`; this way, `Boolean(draw._isHeldByPolarPlugin)` can be used to determine whether a draw comes from a POLAR plugin, and `draw._isHeldByPolarPlugin === localPluginName` can be used to determine whether it's the current one ...

However, I consider this a separate effort. And before we double down on it, we should lock ourselves in the thinking chamber and think really hard about it. After you made a issue/ticket, 🎩 .

warm-coolguy:
> > 🏓 @dopenguin
> 
> 🏓 @warm-coolguy

🏓 @dopenguin 

Although #124 will soon be a blocker, there are some little leftovers to do. After #124 is merged, I will do another functionality check. I didn't verify the current branch state regarding that and will re-check at the end.

packages/plugins/Gfi/src/store/actions/setupMultiSelection.ts
dopenguin: Sounds like a good idea UX-wise! Will implement that!

dopenguin: Changed 17f91e01107ac37435b835e1782463940e5538a8

packages/plugins/Pins/src/store/index.ts
dopenguin: I've added an internal ticket 👍

dopenguin:
> > > 🏓 @dopenguin
> > 
> > 
> > 🏓 @warm-coolguy
> 
> 🏓 @dopenguin
> 
> Although #124 will soon be a blocker, there are some little leftovers to do. After #124 is merged, I will do another functionality check. I didn't verify the current branch state regarding functionality and will re-check at the end.

We can the rest of the PR after #124 is done.

🏓 @warm-coolguy

dopenguin:
> > > > 🏓 @dopenguin
> > > 
> > > 
> > > 🏓 @warm-coolguy
> > 
> > 
> > 🏓 @dopenguin
> > Although #124 will soon be a blocker, there are some little leftovers to do. After #124 is merged, I will do another functionality check. I didn't verify the current branch state regarding functionality and will re-check at the end.
> 
> We can do the rest of the PR after #124 is done.
> 
> 🏓 @warm-coolguy

This PR is now ready receive its final review
:shipit: @warm-coolguy

packages/plugins/Gfi/README.md
```diff
@@ -16,13 +16,14 @@ The GFI plugin can be used to fetch and optionally display GFI (GetFeatureInfo)
...
+| multiSelect | enum["box", "circle"]? | If configured, multiple features can be selected at once by using the modifier key (CTRL on Windows or Command on macOS) and dragging the mouse. Can only be used in Desktop environments. If set to `'box'`, the selection will be done in a box. If set to `'circle'`, the selection will be done in a circle beginning from the first click. In this case, the user has to click a second time to end the selection. Similar to `gfi.directSelect`, features can be added and removed by selection / unselecting them. The features need to be distinguishable by their properties for the functionality to properly work. Does not work together with `extendedMasterportalapiMarkers` of `@polar/core`. Is disabled by default. |
```
warm-coolguy, line 26: It's actually "Shift" on Windows. This makes the toggling de-/selection behaviour confusing since such behaviour is associated with CTRL. My first reaction was to report the behaviour as a bug. Users will probably do the same.

Proposal:

* Selection with "Shift" is purely additive, users have to click somewhere else for deselection
* Selection with "Shift+Strg" has the behaviour currently seen in "Shift" (toggling de-/selection)

This mimics how file selection works on the Windows Explorer, which most users should be familiar with, if they are familiar with any selection scheme. Is it wildly different on MacOS?

packages/plugins/Gfi/README.md
```diff
@@ -16,13 +16,14 @@ The GFI plugin can be used to fetch and optionally display GFI (GetFeatureInfo)
...
-| boxSelect | boolean? | If set to `true`, multiple features can be selected at once by using the modifier key (CTRL on Windows or Command on macOS) and dragging the mouse. Similar to `gfi.directSelect`, features can be added and removed by selection / unselecting them. The features need to be distinguishable by their properties for the functionality to properly work. Does not work together with `extendedMasterportalapiMarkers` of `@polar/core`. Defaults to `false`. |
+| boxSelect | boolean? | **This parameter has been deprecated. Please use `multiSelect` to `'box'` instead.** If set to `true`, multiple features can be selected at once by using the modifier key (CTRL on Windows or Command on macOS) and dragging the mouse. Can only be used in Desktop environments. Similar to `gfi.directSelect`, features can be added and removed by selection / unselecting them. The features need to be distinguishable by their properties for the functionality to properly work. Does not work together with `extendedMasterportalapiMarkers` of `@polar/core`. Defaults to `false`. |
```
warm-coolguy, line 19: As on the new parameter (see other comment), the documentation is currently wrong regarding CTRL.

packages/plugins/Gfi/README.md
warm-coolguy, line 26: >If set to `'circle'`, the selection will be done in a circle beginning from the first click. In this case, the user has to click a second time to end the selection.

This is not true on Windows, where the `circle` behaves in the same fashion as the `box` does, id est "Hold Shift, mouse down, drag circle to desired size, mouse up" encompasses the whole operation. On the other hand, acting as described ("Hold Shift, Click, Drag, Click") does not work.

Is it different on MacOS?

warm-coolguy (changes requested):
>This PR is now ready receive its final review
>:shipit: @warm-coolguy

You can't shipit me since I am not the one to merge this pull request. Shipitting me would have to happen on a pull request I am the assignee of.

🏌️ @dopenguin

warm-coolguy:
@dopenguin The type `MeasureConfiguration` sneaked into our core types with the last PR merge. Since it's topical, please remove it here.

dopenguin:
> > This PR is now ready receive its final review
> > :shipit: @warm-coolguy
> 
> You can't shipit me since I am not the one to merge this pull request. Shipitting me would have to happen on a pull request I am the assignee of.
> 
> 🏌️ @dopenguin

Removed the type 8b29f6acab447492a7d7ba811e95709d3fc73414 and changed the behaviour as described 42faf6390900820d93f2297d2347f234edeb6207.

It now works on Mac, can you test the same for Windows and if it is the way we talked?

🏓 @warm-coolguy

packages/plugins/Gfi/src/store/actions/setupMultiSelection.ts
```diff
@@ -0,0 +1,94 @@
...
+}: PolarActionContext<GfiState, GfiGetters>) {
+  if (boxSelect || multiSelect === 'box' || multiSelect === 'circle') {
+    if (boxSelect) {
+      console.warn(
+        '@polar/plugin-gfi: Configuration parameter "boxSelect" has been deprecated. Please use the new parameter "multiSelect" set to "box" instead.'
+      )
+    }
+    if (multiSelect !== 'circle') {
+      drawOptions.geometryFunction = createBox()
+    }
```
warm-coolguy, line 61:
```suggestion
    if (multiSelect !== 'circle') {
      drawOptions.geometryFunction = createBox()
    }
    else {
      delete drawOptions.geometryFunction;
    }
```
Just in case we ever have switchable select logic at run-time.

warm-coolguy (approved):
> > > This PR is now ready receive its final review
> > > :shipit: @warm-coolguy
> > 
> > 
> > You can't shipit me since I am not the one to merge this pull request. Shipitting me would have to happen on a pull request I am the assignee of.
> > 🏌️ @dopenguin
> 
> Removed the type [8b29f6a](https://github.com/Dataport/polar/commit/8b29f6acab447492a7d7ba811e95709d3fc73414) and changed the behaviour as described [42faf63](https://github.com/Dataport/polar/commit/42faf6390900820d93f2297d2347f234edeb6207).
> 
> It now works on Mac, can you test the same for Windows and if it is the way we talked?
> 
> 🏓 @warm-coolguy

:shipit: @dopenguin It works. I have one last remark in https://github.com/Dataport/polar/pull/211#pullrequestreview-2584318633 and the PR is good to go.

packages/plugins/Gfi/src/store/actions/setupMultiSelection.ts
dopenguin: Done dbb2cec92897d2881a272bebaaf6f3b3e506fce6

PR #218, opened 2025-01-16
fix tooltip font family

warm-coolguy:
## Summary

Since tooltips are technically a neighbour of the #app, the correct font styles missed the the target.

Check client tooltips. You can e.g. view the difference in DISH.

### DIFF
--- a/packages/core/src/utils/createMap/pullVuetifyStyleToShadow.ts
+++ b/packages/core/src/utils/createMap/pullVuetifyStyleToShadow.ts
@@ -15,7 +15,7 @@ export const pullVuetifyStyleToShadow = (shadowRoot: ShadowRoot) => {
   // tooltips are technically a neighbour of the app; add missing font style
   const tooltipStyle = document.createElement('style')
   tooltipStyle.innerHTML = `.v-tooltip__content {
-  font-family: Arial, sans-serif;
+  font-family: sans-serif;
 }`
   shadowRoot.appendChild(tooltipStyle)
 }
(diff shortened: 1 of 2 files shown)


packages/core/src/utils/createMap/pullVuetifyStyleToShadow.ts
```diff
@@ -11,4 +11,11 @@ export const pullVuetifyStyleToShadow = (shadowRoot: ShadowRoot) => {
     return
   }
   shadowRoot.appendChild(vuetifyStyle)
+
+  // tooltips are technically a neighbour of the app; add missing font style
+  const tooltipStyle = document.createElement('style')
+  tooltipStyle.innerHTML = `.v-tooltip__content {
+  font-family: Arial, sans-serif;
```
dopenguin, line 18: `sans-serif` is used for `font-family` in `MapContainer.vue` which should suffice here as well. But I don't mind trying Arial at first anyway 🎩

dopenguin (approved):
Looks good! Only left one small comment

🚀 @warm-coolguy

packages/core/src/utils/createMap/pullVuetifyStyleToShadow.ts
warm-coolguy: Yeah, makes sense to use the same one here. It'll usually be Arial anyway, but still ... https://github.com/Dataport/polar/pull/218/commits/6d128f79f1ff9b1c4481ea948e6e6a02e796b1f4

PR #221, opened 2025-01-28
Fix/documentation supplement in address search

jedi-of-the-sea:
## Summary

The configuration examples in AddressSearch were not sufficient in case of the bkg and mpapi search methods. To prevent a linting error, it is necessary to typecast the `queryParameters` to `BKGParameters` or `MpApiParameters`.

 - 

## Pull Request Checklist (for Reviewee)

- [x] Changelogs are maintained - not relevant
- [x] Functionality has been tested in Firefox, Chrome, Safari - not relevant
- [x] Functionality has been tested on a smartphone - not relevant
- [x] Functionality has been tested with 200% screen zoom - not relevant
- [x] Screenreader functionality has been manually tested with NVDA - not relevant

UI has been tested in the following tools regarding accessibility (only regarding functionality affected in this PR) - not relevant
  - [x] Chrome Lighthouse - not relevant
  - [x] Firefox Accessibility - not relevant


warm-coolguy (commented):
I would like to reject this pull request since it changes documentation facing users that currently do not have access to our type files (they're not part of the published packages), and we know most of them use `.js` anyway. Furthermore, examples explicitly refer to javascript (as seen on code fences: "\`\`\`js") and this wouldn't be valid syntax there.

@jedi-of-the-sea Should you agree, please close this pull request.

warm-coolguy:
@jedi-of-the-sea Since our users can't use our typescript types, that's still somewhat confusing to them; I'd expect to have a usable `.d.ts file` on reading that, and there'd be none to find.

Alternative suggestion: We extend the `QueryParameters` interface to

```
/**
 * Additional queryParameters for the GET-Request;
 * for the specific parameters for each request,
 * please refer to the types in the plugin
 */
export interface QueryParameters {
  /** sets the maximum number of features to retrieve */
  maxFeatures?: number
  /** depending on backend service, anything may belong to the query */
  [key: string]: unknown
}
```

That's a little unspecific, but saves us various typecasts, and is ultimately true.

@dopenguin Requesting approval for that change.

dopenguin:
I'd go along with

> I would like to reject this pull request since it changes documentation facing users that currently do not have access to our type files (they're not part of the published packages), and we know most of them use `.js` anyway. Furthermore, examples explicitly refer to javascript (as seen on code fences: "```js") and this wouldn't be valid syntax there.
> 
> @jedi-of-the-sea Should you agree, please close this pull request.

However, it may become interesting again if we deliver `d.ts`-files along with out package, as mentioned in https://github.com/Dataport/polar/pull/221#issuecomment-2623834830. But that would probably need a merge of e.g. #108 and an upgrade to `vue`@3 - so nothing relevant for now.

PR #231, opened 2025-02-04
add export and example to diplan client

warm-coolguy:
## Summary

Add export and example to diplan client.

May or may not look at `npm run diplan:dev`.


packages/clients/diplan/src/polar-client.ts
```diff
@@ -91,6 +92,14 @@ polarCore.addPlugins([
     coordinateSources: [],
     layers: {},
   }),
+  Export({
+    layoutTag: NineLayoutTag.BOTTOM_LEFT,
+    /*
+     * icons: Currently not implemented, provided icon set has no fitting icon
+     * atm. Suggestion: Searching for "picture" on icomoon yields nice results.
+     * We should ask implementors of diplanung-style to add them eventually.
+     */
```
dopenguin, line 101: Should we make this a task instead of adding the comment here?

packages/clients/diplan/example/index.html
```diff
@@ -36,6 +36,11 @@ <h2>Informationen aus Kartenklient</h2>
               <td id="subscribed-zoom">uninitialized</td>
               <td>Zoomstufe der <a target="_blank" href="https://openlayers.org/en/latest/apidoc/module-ol_View-View.html#getZoom">OpenLayers/View</a> als <code>Number</code></td>
             </tr>
+            <tr>
+              <td><code>plugin/export/exportedMap</code></td>
+              <td><img onclick="window.open(this.src, '_blank');" id="subscribed-export" /></td>
```
dopenguin, line 41: This doesn't show the image in a new tab in Chrome and Safari. Seems to be a similar issue as in `@polar/snowbox`.
Mind adding that as a task to the issue?

It also seems to be that `img` can always handle base64-strings and `embed` can't in some browsers 🤨

dopenguin (changes requested):
🏓 @warm-coolguy 

Some small things

packages/clients/diplan/example/index.html
warm-coolguy: Fixed it, it now downloads on Chrome/Safari. They actually can handle base64 strings (see data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAgAAAAIAQMAAAD+wSzIAAAABlBMVEX///+/v7+jQ3Y5AAAADklEQVQI12P4AIX8EAgALgAD/aNpbtEAAAAASUVORK5CYII), but it turns out the file was too big for some browsers.

packages/clients/diplan/src/polar-client.ts
warm-coolguy: I created an additional task and sent you the link. I'll leave the comment in so that potential other devs are not confused upon stumbling upon this.

dopenguin (approved):
:shipit: @warm-coolguy

PR #235, opened 2025-02-13
add translate mode to draw

warm-coolguy:
## Summary

A translate option has been added to the Draw interactions.

Run snowbox and translate freshly drawn features.

- [x] Functionality has been tested in Firefox, Chrome, ~~Safari~~

  - [x] Chrome Lighthouse
  - [x] Firefox Accessibility

Regarding a11y, no new warnings have been added.

### DIFF
--- a/packages/plugins/Draw/CHANGELOG.md
+++ b/packages/plugins/Draw/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## unpublished
 
-- Feature: Added a "Translate" mode that allows moving drawn features as-are.
+- Feature: Add a "Translate" mode that allows moving drawn features as they are.
 
 ## 3.0.0
 
--- a/packages/plugins/Draw/CHANGELOG.md
+++ b/packages/plugins/Draw/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## unpublished
+
+- Feature: Added a "Translate" mode that allows moving drawn features as-are.
+
 ## 3.0.0
 
 - Breaking: Upgrade peerDependency `ol` from `^9.2.4` to `^10.3.1`.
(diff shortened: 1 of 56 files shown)


packages/plugins/Draw/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## unpublished
+
+- Feature: Added a "Translate" mode that allows moving drawn features as-are.
```
dopenguin, line 5:
```suggestion
- Feature: Add a "Translate" mode that allows moving drawn features as-are.
```
Would you also go along with "as they are" instead "as-are"?

packages/plugins/Draw/src/store/createInteractions/createModifyInteractions.ts
```diff
@@ -1,40 +1,30 @@
...
+import { Collection, Feature, Map } from 'ol'
 import { CreateInteractionsPayload, DrawGetters, DrawState } from '../../types'
+import { makeLocalSelector } from './localSelector'
 
 const createModify = (
   map: Map,
   drawLayer: CreateInteractionsPayload['drawLayer']
 ) => {
-  let active = false
+  const activeContainer = { active: false }
```
dopenguin, line 12: Why is such a wrapper object needed? Would it be not reactive anymore otherwise?

packages/plugins/Draw/src/store/createInteractions/createTranslateInteractions.ts
dopenguin, line 1: This reads exactly as `createModify` with the only difference that `Translate` is being used as the interaction.

Would it be a good idea to handle both translate and modify under one menu point and a user may select in a new radio whether they would like to modify or translate a feature? Or do you think that is more fitting for the upcoming refactor of this plugin?

dopenguin (changes requested):
Just some small things, not much to change, looks and works good!

🏓 @warm-coolguy

dopenguin (approved):
Only one minor suggestion and two questions so we can go with

:shipit: @warm-coolguy

PR #238, opened 2025-02-17
rename global diplan client to technical name

warm-coolguy:
## Summary

There was a TODO regarding the global name that was set. I figured just writing out the package name would result in a plausible default name akin to what'd've to be written in an import's from part.

Check if `build` and `build:serve` still work for diplan.


dopenguin (approved):
:shipit: @warm-coolguy

PR #241, opened 2025-02-17
Allow layers that have singleTile: true set to be used for gfi requests

dopenguin:
## Summary

If a WMS-Layer had `singleTile` set to `true`, gfi requests were not possible as the `if`-statement only checked for `TileLayer` which is not the case for single-tiled-WMS-layers.

- Configure the Layer with the id `367` in `@polar/client-snowbox` for general usage and gfi
- `npm run snowbox`
- Clicking in the map should yield a result.

- ~~[ ] Screenreader functionality has been manually tested with NVDA~~

  - ~~[ ] Chrome Lighthouse~~
  - ~~[ ] Firefox Accessibility~~

None

### DIFF
--- a/packages/plugins/Gfi/CHANGELOG.md
+++ b/packages/plugins/Gfi/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## unpublished
+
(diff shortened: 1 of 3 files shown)


warm-coolguy (approved):
:shipit: @dopenguin

PR #242, opened 2025-02-18
add lasso logic to draw

warm-coolguy:
WIP

## Summary

- Added a lasso function to the draw tool
  - It allows free-hand drawing a polygon, and all geometries of configured layers completely within that polygon are copied up to the draw layer for further modification/processing

Run `diplan:dev` to check whether the feature works. A button has been added there above the map to allow enabling the lasso function. When drawing that lasso zoomed in to a level that would suffice to display the configured parcel cells, all parcel cells completely covered by the drawn lasso will be copied up to the draw layer.

### DIFF
--- a/packages/clients/diplan/example/config.js
+++ b/packages/clients/diplan/example/config.js
@@ -98,6 +98,11 @@ export default {
   },
   draw: {
     enableOptions: true,
+    lassos: [
+      {
+        id: flurstuecke,
+      },
+    ],
     measureOptions: {
       metres: true,
       kilometres: true,
@@ -111,6 +116,7 @@ export default {
         family: 'Arial',
       },
     },
+    toastAction: 'plugin/toast/addToast',
     style: {
       fill: { color: 'rgb(51 117 212 / 50%)' },
       stroke: {
--- a/packages/plugins/Draw/README.md
+++ b/packages/plugins/Draw/README.md
@@ -33,12 +33,16 @@ The styling of the drawn features can be configured to overwrite the default ol-
 
 | fieldName | type | description |
 | - | - | - |
+| addLoading | string? | Expects the path to a mutation within the store. This mutation is committed with a plugin-specific loading key as payload when starting asynchronous procedures that are intended to be communicated to the user. |
 | enableOptions | boolean? | If `true`, draw options are displayed, like choosing and changing stroke color. Not available for texts features. Defaults to `false`. |
+| lassos | lasso[]? | Allows configuring lasso options. The lasso function allows free-hand drawing a geometry on the map; features completely fitting into that geometry will be copied up to the draw layer from all configured layers. 💡 Please mind that the lasso function currently has no UI in the client, and the functionality must be triggered externally. |
+| removeLoading | string? | Expects the path to a mutation within the store. This mutation is committed with a plugin-specific loading key as payload when finishing asynchronous procedures that are intended to be communicated to the user. |
 | measureOptions | measureOptions? | If set, an additional radio is being shown to the user to be able to let the (then) drawn features display their length and / or area. See [draw.measureOptions](#drawmeasureoptions) for further information. Not shown by default. |
 | selectableDrawModes | string[]? | List 'Point', 'LineString', 'Circle', 'Text' and/or 'Polygon' as desired. All besides 'Text' are selectable by default. |
 | snapTo | string[]? | Accepts an array of layer IDs. If these layers are active, they are used as snapping material for geometry manipulation. The Draw layer will also always snap to its own features regardless. Please mind that used layers must provide vector data. The layers referred to must be configured in `mapConfiguration.layers`. |
 | style | style? | Please see example below for styling options. Defaults to standard OpenLayers styling. |
 | textStyle | textStyle? | Use this object with properties 'font' and 'textColor' to style text feature. |
+| toastAction | string? | This string will be used as action to send a toast information to the user to clarify why something happened in edge cases. If this is not defined, the information will be printed to the console for debugging purposes instead. |
 
 For details on the `displayComponent` attribute, refer to the [Global Plugin Parameters](../../core/README.md#global-plugin-parameters) section of `@polar/core`.
 
@@ -55,8 +59,8 @@ draw: {
     },
   },
   style: {
-    fill: { 
-      color: 'rgba(255, 255, 255, 0.5)' 
+    fill: {
+      color: 'rgba(255, 255, 255, 0.5)'
     },
     stroke: {
       color: '#e51313',
@@ -71,6 +75,13 @@ draw: {
 ```
 
 </details>
+
+#### draw.lasso
+
+| fieldName | type | description |
+| - | - | - |
+| id | string | The layer id of a vector layer to copy up vector features from. |
+| minZoom | boolean | Defaults to `true`. If a boolean is given, the `minZoom` (if configured) of the `LayerConfiguration` in `mapConfiguration.layers` will be adhered to when copying up geometries from the source. This is to prevent the client from overly burdening feature-rich vector services on accident. |
 
 #### draw.measureOptions
 
--- a/packages/plugins/Draw/src/store/createInteractions/createLassoInteractions.ts
+++ b/packages/plugins/Draw/src/store/createInteractions/createLassoInteractions.ts
@@ -1,0 +1,104 @@
+import Draw from 'ol/interaction/Draw'
+import Interaction from 'ol/interaction/Interaction'
+import i18next from 'i18next'
+import { rawLayerList } from '@masterportal/masterportalapi'
+import { PolarActionContext } from '@polar/lib-custom-types'
+import { DrawGetters, DrawState } from '../../types'
+
+const loaderKey = 'drawLasso'
+
+/*
+TODO use snippet to confirm the layer type worked on
+    const source = (
+      map
+        .getLayers()
+        .getArray()
+        .find((layer) => layer.get('id') === layerId) as VectorLayer
+    )?.getSource?.()
+    if (source instanceof VectorSource) {
+      accumulator.push(new Snap({ source }))
+    } else {
+      console.warn(
+        `@polar/plugin-draw: Layer with ID "${layerId}" configured for 'snapTo', but it has no source to snap to. The layer does probably not hold any vector data.`
+      )
+    }
+      */
+
+// TODO un – refactor after done
+// eslint-disable-next-line max-lines-per-function
+export default function ({
+  rootGetters,
+  commit,
+  dispatch,
+}: PolarActionContext<DrawState, DrawGetters>): Interaction[] {
+  const draw = new Draw({
+    type: 'Polygon',
+    freehand: true,
+  })
+  const { addLoading, removeLoading } = rootGetters.configuration?.draw || {}
+  // TODO un
+  // eslint-disable-next-line max-lines-per-function
+  draw.on('drawend', (e) => {
+    dispatch('setMode', 'none')
+
+    const drawnLasso = e.feature
+    const lassos = rootGetters.configuration.draw?.lassos || []
+    const toastAction = rootGetters.configuration.draw?.toastAction || ''
+    const promises = lassos.reduce((accumulator, { id, minZoom = true }) => {
+      const layerConfig = rootGetters.configuration.layers?.find(
+        (layer) => id === layer.id
+      )
+      if (
+        minZoom &&
+        layerConfig &&
+        typeof layerConfig.minZoom !== 'undefined' &&
+        rootGetters.zoomLevel < layerConfig.minZoom
+      ) {
+        if (toastAction) {
+          dispatch(toastAction, {
+            type: 'info',
+            text: i18next.t('plugins.draw.lasso.notInZoomRange', {
+              serviceName: layerConfig.name || id,
+            }),
+            timeout: 10000,
+          })
+        } else {
+          console.warn(
+            `Lasso not used with layer with id "${id}". (minZoom not reached)`
+          )
+        }
+        return accumulator
+      }
+      const serviceDefinition = rawLayerList.getLayerWhere({ id })
+
+      // TODO implement for OAF, use existing getFeatures for WFS
+      // https://api.hamburg.de/datasets/v1/alkis_vereinfacht/collections/Flurstueck/items?f=json&limit=10&bbox=10.0872,53.5357,10.0883,53.5362
+
+      accumulator.push(new Promise(() => ({})))
+      return accumulator
+    }, [] as Promise<object>[])
+
+    Promise.all(promises)
+      .then((resolutions) => {
+        if (addLoading) {
+          commit(addLoading, loaderKey, { root: true })
+        }
+
+        // TODO parse resolutions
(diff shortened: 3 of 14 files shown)


packages/clients/diplan/example/config.js
```diff
@@ -124,6 +132,7 @@ export default {
         family: 'Arial',
       },
     },
+    toastAction: 'plugin/toast/addToast',
```
dopenguin, line 135: Does this better fit in `packages/clients/diplan/src/polar-client.ts`?

packages/plugins/Draw/src/store/index.ts
```diff
@@ -79,14 +79,38 @@ export const makeStoreModule = () => {
...
+            if (
+              minZoom &&
+              layerConfig &&
+              typeof layerConfig.minZoom !== 'undefined' &&
+              rootGetters.zoomLevel < layerConfig.minZoom
+            ) {
+              return accumulator
+            }
+            accumulator.push(id)
+            return accumulator
```
dopenguin, line 109:
```suggestion
            return minZoom &&
              layerConfig &&
              typeof layerConfig.minZoom !== 'undefined' &&
              rootGetters.zoomLevel < layerConfig.minZoom
              ? accumulator
              : [...accumulator, id]
```
🏌️

packages/plugins/Draw/src/store/createInteractions/createLassoInteractions.ts
```diff
@@ -0,0 +1,153 @@
+import Draw from 'ol/interaction/Draw'
+import i18next from 'i18next'
+import { GeoJSON } from 'ol/format'
+import { booleanContains } from '@turf/boolean-contains'
+import { rawLayerList } from '@masterportal/masterportalapi'
+import { PolarActionContext } from '@polar/lib-custom-types'
+import { Feature } from 'ol'
+import { Polygon } from 'ol/geom'
+import { parseWfsResponse } from '@polar/lib-get-features/wfs/parse'
```
dopenguin, line 9: If this is needed, this should be a named export from the package-root, as defined by [this ADR](https://github.com/Dataport/polar/wiki/Architecture-Decision-Records-%F0%9F%93%83#2024-03-05-how-to-expose-additional-exports-of-a-package)

packages/lib/getFeatures/vector/index.ts
dopenguin, line 1: Do you expect the parameters of these function to drastically change in the future? If not, I'd suggest adding the parameters directly to the function instead of using an object.
IMO, that would make the function signature more readable.

packages/lib/getFeatures/vector/index.ts
```diff
@@ -0,0 +1,66 @@
...
+          `request=GetFeature`,
+          `srsName=${projectionCode}`,
+          `typeName=${serviceDefinition.featureType}`,
+          `bbox=${bbox},${projectionCode}`,
+        ].join('&')}`
+
+  return fetch(url)
+}
+
+export const getVectorFeaturesByFeatureRequest = ({
```
dopenguin, line 44: This currently looks like a redundant wrapper function.
Wouldn't exposing and using `getVectorFeaturesByBboxRequest` suffice?

packages/lib/getFeatures/vector/index.ts
```diff
@@ -0,0 +1,66 @@
...
+  projectionCode,
+}: {
+  bbox: number[]
+  fetchLayerId: string
+  projectionCode: string
+}) => {
+  const serviceDefinition = rawLayerList.getLayerWhere({ id: fetchLayerId })
+  if (!supportedFormats.includes(serviceDefinition.typ)) {
+    throw new Error(
+      `@polar/lib-get-features#getVectorFeaturesByBboxRequest: Layer with ID "${fetchLayerId}" of type "${serviceDefinition.typ}" used to retrieve vector data, but is not within the range of supported types [${supportedFormats}].`
```
dopenguin, line 18:
```suggestion
      `@polar/lib-get-features#getVectorFeaturesByBboxRequest: Layer with ID "${fetchLayerId}" of type "${serviceDefinition.typ}" is used to retrieve vector data, but is not within the range of supported types [${supportedFormats}].`
```

packages/lib/getFeatures/vector/index.ts
```diff
@@ -0,0 +1,66 @@
...
+    )
+  }
+
+  const [codeName, codeNumber] = projectionCode.split(':')
```
dopenguin, line 22: What do you think about `srsName` and `srsCode`? Whenever I am working with projections, I always stumble over these two.

packages/plugins/Draw/README.md
```diff
@@ -72,6 +76,13 @@ draw: {
 
 </details>
 
+#### draw.lasso
+
+| fieldName | type | description |
+| - | - | - |
+| id | string | The layer id of a vector layer to copy up vector features from. |
+| minZoom | boolean | Defaults to `true`. If a boolean is given, the `minZoom` (if configured) of the `LayerConfiguration` in `mapConfiguration.layers` will be adhered to when copying up geometries from the source. This is to prevent the client from overly burdening feature-rich vector services on accident. |
```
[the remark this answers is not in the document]
warm-coolguy: It's part of the workflow currently in development to copy polygons from the parcel layer and forward adjustments to follow-up processing steps. I've optionally restricted this to the `minZoom` to avoid loading e.g. all parcels of Hamburg (on accident), since that may heavily and unnecessarily burden the services, as well as slowing the client down.

It may eventually be used in other scenarios that match this "WFS-T-less WFS-T" procedure where a heavier procedures than just firing a CUD-Transaction from the client are needed. But I don't think this'll happen a lot.

packages/plugins/Draw/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Feature: Add a `"translate"` mode that allows moving drawn features as they are.
 - Feature: Add a `"snapTo"` key to the configuration that allows defining vector layers to snap to while drawing, editing, and translating.
+- Feature: Add a lasso mode that allows copying up features from a vector layer that are contained within the user's hand drawn polygon. This feature is currently without UI and needs to be started from the outside, see docs.
```
warm-coolguy, line 7:
```suggestion
- Feature: Add a lasso mode that allows copying up features from a vector layer that are contained within the user's hand drawn polygon.
```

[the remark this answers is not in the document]
warm-coolguy: It now containsn't the last sentence.

packages/plugins/Draw/CHANGELOG.md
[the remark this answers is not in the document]
warm-coolguy: https://github.com/Dataport/polar/pull/242/commits/eb4b2f72f2641aee950ef02a3d42b7299aa1529c

packages/plugins/Draw/src/store/index.ts
warm-coolguy: Yeah, I don't like that.

packages/clients/diplan/example/config.js
warm-coolguy: Yep. TODO

packages/lib/getFeatures/vector/index.ts
warm-coolguy: I've sometimes regretted hard-wiring in which slots which parameters go, but I've never regretted using a parameter object. So I'll go with the latter.

packages/lib/getFeatures/vector/index.ts
warm-coolguy: How is it redundant? It's a different API and handles the required conversion.

packages/lib/getFeatures/vector/index.ts
warm-coolguy: https://github.com/Dataport/polar/pull/242/commits/d39895448f4d79b24593d88ae7abcaa6ba9d21c1 :^)

packages/lib/getFeatures/vector/index.ts
warm-coolguy: I don't have any strong opinions regarding these fellas.

packages/plugins/Draw/src/store/createInteractions/createLassoInteractions.ts
warm-coolguy: Since ol drops its barrel files, shall we, too? If you agree, I'll just delete the ADR. If you insist, I can expose it, sure.

packages/clients/diplan/example/config.js
warm-coolguy: https://github.com/Dataport/polar/pull/242/commits/5edfb485f8a999ca34f1bcdfc8fc5d278ab109ce TODONE

packages/plugins/Draw/src/store/createInteractions/createLassoInteractions.ts
dopenguin: If we intend to get rid of barrel files at some point in the future, we should then document the relevant exports and remove the currently exposed exports from the main file (except the "normal" export).

But as this is not on currently on the roadmap, please expose the export.

packages/lib/getFeatures/vector/index.ts
dopenguin: Mind changing them to the above? 😊

packages/lib/getFeatures/vector/index.ts
dopenguin: To me, it feels weird to call a function that may throw an error which I then have to catch.
Thus, I'd mostly use `getVectorFeaturesByBboxRequest` directly and test whether my feature has a `bbox` beforehand.

I don't care enough though 🎩

packages/plugins/Draw/README.md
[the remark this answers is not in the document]
warm-coolguy: With small screens or services with a high minZoom, being able to lasso a larger region may be desirable, especially since we don't just need to touch the geometries, but encompass them completely. For the service at hand, I found it to be quite a hassle to lasso the features appropriately in some regions of Hamburg where parcels are large (see industrial regions), whilst we need to maintain the rather high minZoom for regions where a ton of small parcels reside.

So whether or not one can select from a layer not currently in minZoom is quite opinionated, and as such, I made it configurable.

packages/lib/getFeatures/vector/index.ts
warm-coolguy: Oh, that was supposed to be a renaming suggestion.

For WFS, `srsName` would e.g. be `"EPSG:1234"` and not `"EPSG"`, as this change would imply. For OAF, `srsName` has no meaning, since the specification does not mention it and directly references `crs` instead.

Now that I looked it up, `srsCode` seems to exist in neither spec.

I don't see a reason to change them.

packages/plugins/Draw/src/store/createInteractions/createLassoInteractions.ts
warm-coolguy: [The debate is on the roadmap now.](https://github.com/Dataport/polar/wiki/Architecture-Decision-Records-%F0%9F%93%83#2025-02-24-revoke-2024-03-05-how-to-expose-additional-exports-of-a-package)

For the time being, https://github.com/Dataport/polar/pull/242/commits/b5fbda45b200801ffc1f899f7ff06f9246411bce.

warm-coolguy:
> Almost finished with the review, but will tackle `packages/plugins/Draw/src/store/createInteractions/createLassoInteractions.ts` with a fresh head.
> 
> 🏓 @warm-coolguy

🏓 @dopenguin

packages/plugins/Draw/src/store/createInteractions/createLassoInteractions.ts
```diff
@@ -0,0 +1,155 @@
...
+  dispatch,
+}: PolarActionContext<DrawState, DrawGetters>) {
+  const draw = new Draw({ type: 'Polygon', freehand: true })
+
+  // should be fine, line bloat is from error handling (only logging/toasting)
+  // eslint-disable-next-line max-lines-per-function
+  draw.on('drawend', (e) => {
+    dispatch('setMode', 'none')
+
+    const toast = (toastObject) =>
```
dopenguin, line 87: Now might be a good time to move the type `Toast` from `@polar/plugin-toast` to `@polar/lib-custom-types` to use it here (and possibly other places in the future)

packages/plugins/Draw/src/store/createInteractions/createLassoInteractions.ts
```diff
@@ -0,0 +1,155 @@
...
+                  type: 'error',
+                  text: i18next.t('plugins.draw.lasso.layerRejected', {
+                    id: getters.activeLassoIds[index],
+                  }),
+                })
+                return false
+              }
+              return true
+            }) as PromiseFulfilledResult<Response>[]
+          ).map(async (resolution, index) =>
```
dopenguin, line 129: Help me out: Why is this called `resolution`? 🤨

dopenguin (changes requested):
> > Almost finished with the review, but will tackle `packages/plugins/Draw/src/store/createInteractions/createLassoInteractions.ts` with a fresh head.
> > 🏓 @warm-coolguy
> 
> 🏓 @dopenguin

🏓 @warm-coolguy

dopenguin (approved):
The last two comments don't seem to be as relevant to the PR, so I'll leave this here already.
:shipit: @warm-coolguy

packages/plugins/Draw/src/store/createInteractions/createLassoInteractions.ts
warm-coolguy: The returned value of a resolved promise can be a resolution to it. But I get how this may be confusing in a map application. I made it less zoomy and called it `result` now. https://github.com/Dataport/polar/pull/242/commits/bb093e2aa47f8b4daea2a2fba4cd5cc394043cfb

packages/plugins/Draw/src/store/createInteractions/createLassoInteractions.ts
warm-coolguy: While yes, I'd also rather not do it. We're practically throwing `any` payloads at dispatchments everywhere since there's currently no type safety for those anyway. Maybe we could cook up a solution like for locales eventually, but I'd like to avoid this bigger picture for now, and introducing a breaking change for minimal gain whilst not adressing the main issue seems a little pointless right now.

warm-coolguy:
> The last two comments don't seem to be as relevant to the PR, so I'll leave this here already. :shipit: @warm-coolguy

:shipit: @dopenguin If my responses seem sufficient, please feel free to merge yourself, since it opens up reviewability to the next PR.

packages/plugins/Draw/src/store/createInteractions/createLassoInteractions.ts
dopenguin: It wouldn't be a breaking change as `@polar/plugin-toast` does not export that type while the types `ToastStyle`, ToastType` and `ToastTypeStyles` are already part of `@polar/lib-custom-types`.

The benefit ain't drastic here, though, so we can leave it at that.

warm-coolguy: Oh, didn't realize it's not exported. 👀

PR #248, opened 2025-02-24
Feature/add download button in bgw gfi

jedi-of-the-sea:
## Summary

A custom gfi was added to Badegewässer Client which contains a download button with a link to a page displaying further information about the chosen bathing area.

Start the Badegewässer client and click on a bathing area to open the gfi. The custom gfi will be displayed. Click the download button. This should open a new browser tab.

- [x] Changelogs are maintained - **the client has not been published yet so there isn't a new entry.**
- [x] Functionality has been tested on a smartphone (emulated with browser)

  - [x] Chrome Lighthouse
  - [x] Firefox Accessibility


warm-coolguy (approved):
🏓 @jedi-of-the-sea (and then 🚀)

I'm a little confused why it's a "download"Link or an bgw-"export"-button since it seems to be merely a link. Since this may confuse others as well ("Is this a bug? Why isn't it downloading?"), please rename at least the "download" part.

I figure that's simple enough to not need further inspection and you can just merge thereafter. 👍

PR #249, opened 2025-02-24
Feature/secure services

dopenguin:
## Summary

Add possibility to use services that are authenticated by OIDC by passing a Bearer token when sending requests.

- Get the configuration from me through a secure channel
- `npm run diplan:dev`
- Start CORS Plugin
- Login with credentials
- Load secure layer

Caveat: If the user tries loading the secure layer while not authenticated, the layer is never shown. This should not be a problem in a proper authentication environment and thus I have not addressed this here.

- ~[ ] Screenreader functionality has been manually tested with NVDA~ Not needed

  - [x] Chrome Lighthouse
  - [x] Firefox Accessibility


packages/clients/diplan/example/authentication/index.js
```diff
@@ -0,0 +1,73 @@
...
+  eraseCookies([
+    'token',
+    'expires_in',
+    'refresh_token',
+    'name',
+    'username',
+    'email',
+    'expiry',
+  ])
+  window.location.reload()
```
warm-coolguy, line 53: Is this an F5? Doesn't it completely eradicate all of the user's progress?

warm-coolguy (changes requested):
🏓 @dopenguin Unfinished review, I'll add more later.

All in all, I do not see how an application is supposed to use us properly. It seems to me that we require both a configuration parameter for the token and an update mechanism to replace it in time that the login-managing application is required to call in time. Or did I just not see the idea here?

packages/clients/diplan/example/authentication/cookie.js
```diff
@@ -0,0 +1,70 @@
...
+export function setCookies(token, expiresIn, refreshToken) {
+  setCookie('token', token)
+  setCookie('expires_in', expiresIn)
+  setCookie('refresh_token', refreshToken)
+
+  const account = parseJWT(token)
+
+  setCookie('name', account?.name)
+  setCookie('email', account?.email)
+  setCookie('username', account?.preferred_username)
```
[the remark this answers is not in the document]
dopenguin: Applications typically save these values, similarly to https://github.com/Dataport/polar/pull/249#discussion_r1969190662. We can keep the example simple you like

packages/clients/diplan/example/authentication/cookie.js
```diff
@@ -0,0 +1,70 @@
...
+
+  setCookie('expiry', account?.exp)
```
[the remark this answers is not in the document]
dopenguin: See above

packages/clients/diplan/example/authentication/index.js
dopenguin: Correct. But if a user would log out in an application that includes POLAR, this would be the case in most situation. Or am I on the wrong track?

packages/clients/diplan/example/authentication/cookie.js
```diff
@@ -0,0 +1,70 @@
+import { parseJWT } from './parseJWT'
+
+/*
+ * Functions are based on https://bitbucket.org/geowerkstatt-hamburg/masterportal/src/dev_vue/src/modules/login/js/utilsCookies.js
+ * and https://bitbucket.org/geowerkstatt-hamburg/masterportal/src/dev_vue/src/modules/login/js/utilsOIDC.js.
+ */
```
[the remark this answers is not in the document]
dopenguin: The configuration option was something I had in mind initially but scrapped quickly.

There are two options that can solve using the token:
- Using the token when calling e.g. `createMap`
- Requesting the token from the Cookie

If we'd get the token from the call to `createMap`, we'd have to handle the situation when a token expires and request a new one as the function call is already done. If we'd be doing that with the configuration, the situation would be similar.

By having the token in a cookie, which is usually done by applications that use OIDC tokens for authentication anyway, we can simply request it on the go and not worry about refreshing it on our own.

Do you think an adjustment to the documentation is necessary so that it is clear that `interceptorUrlRegex` should only be used by integrated clients as we currently do not offer a login module?

packages/core/src/utils/createMap/addInterceptor.ts
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * @param interceptorUrlRegex - URLs fitting this regular expression have the token added.
+ */
+export function addInterceptor(interceptorUrlRegex: string) {
+  const { fetch: originalFetch } = window
+
+  window.fetch = (resource, originalConfig) => {
+    let config = originalConfig
+
+    // @ts-expect-error | Has worked like charm so far. If an error occurs if resource is of type RequestInfo, take another look
```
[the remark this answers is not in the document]
dopenguin: Adjusted 26d14e29d37e19993dde33831f808fdcfb112499

packages/clients/diplan/example/index.html
```diff
@@ -15,6 +15,22 @@ <h1>Testseite <code>@polar/client-diplan</code> (dev)</h1>
...
+            </label>
+            <label>
+              Passwort:
+              <input type="password" id="password" name="password" required>
+            </label>
+            <br>
+            <button id="login-button" type="button">Login</button>
+          </form>
+          <p id="error-message" style="color: red;"></p>
+        </div>
```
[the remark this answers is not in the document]
dopenguin: Added 4a993661

dopenguin:
> 🏓 @dopenguin Unfinished review, I'll add more later.
> 
> All in all, I do not see how an application is supposed to use us properly. It seems to me that we require both a configuration parameter for the token and an update mechanism to replace it in time that the login-managing application is required to call in time. Or did I just not see the idea here?

As mentioned in https://github.com/Dataport/polar/pull/249#discussion_r1969546572, the token requests should be done by the leading application and not POLAR.

packages/core/src/utils/createMap/addInterceptor.ts
```diff
@@ -0,0 +1,32 @@
...
+export function addInterceptor(interceptorUrlRegex: string) {
+  const { fetch: originalFetch } = window
+
+  window.fetch = (resource, originalConfig) => {
+    let config = originalConfig
+
+    if (
+      interceptorUrlRegex &&
+      typeof resource === 'string' &&
+      resource?.match(interceptorUrlRegex)
```
warm-coolguy, line 21:
```suggestion
      resource.match(interceptorUrlRegex)
```
It's not required anymore after the type check.

packages/clients/diplan/example/authentication/cookie.js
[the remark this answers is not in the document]
warm-coolguy: Yes, please keep the example minimal.

packages/clients/diplan/example/authentication/cookie.js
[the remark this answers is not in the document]
warm-coolguy: Assuming it's done together with the other thread. 🎩

packages/clients/diplan/example/authentication/cookie.js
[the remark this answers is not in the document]
warm-coolguy: The `createMap` parameter combined with an action provided to update the token for our application when needed should be the way to go here. This way we initially have a token to start working with and delegate its updates to the leading application that may now refresh and keep it however it desires to – forwarding it to the map is a follow-up step.

This also spares the leading application to be forced to produce a cookie `token`, which it may not have intended to, or which may also already be in use for another token since that's a very generic word – since the client is supposed to be used in multiple spots eventually, even both may hold true in time.

This way we're also keeping data flow to the etablished channels and don't open up a new one.

packages/core/src/utils/createMap/addInterceptor.ts
```diff
@@ -0,0 +1,29 @@
...
+      config = {
+        ...originalConfig,
+        // eslint-disable-next-line @typescript-eslint/naming-convention
+        headers: { Authorization: `Bearer ${getCookie('token')}` },
+      }
+    }
+
+    return originalFetch(resource, config)
+  }
+}
```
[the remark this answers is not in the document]
warm-coolguy: My first idea was to identify the call source, and only act if it was from within POLAR. Since `.caller` is deprecated, checking the call stack seems to be a non-solution, and we can't just modify all local requests since OL and the masterportalAPI may start them, too.

The only thing left seems to be the URL itself. We may modify all URLs intended for interception on arrival. Since everything flows through the client creation (either by config or from the services.json), all URLs may run through a check on whether they fit the `interceptorUrlRegex`. If so, we modify them to hold an appropriate flag.

Ideas on how to mark URLs:

* `${url}+POLAR_INTERCEPTOR+` or another string that will absolutely not randomly appear in the finite time that remains to the universe – however, if there's any operations on the end of the URL, this may or may not break anything?
* I wonder whether ?, &, or # may also be made use of, like e.g. in `${url}?polar_interceptor_was_here=true` – this may survive further URL modifications since it's technically a correct URL, but I wonder if some URLs are required to not have any appendices like that by e.g. the masterportalAPI?
* `${scheme}${subdomain}POLAR_INTERCEPTOR.${domain}` would hide that as an additional subdomain, and I'm pretty sure it will survive all underlying requests there until fetch is finally used.

Then, in the interceptor, the flag is removed before the real URL is used with the Auth header. This way all fetch calls of the outlying application to the URL are unmodified, and all of our own calls are modified.

That or you don't override all original headers and only add `Authorization` when it's not there already.

packages/clients/diplan/example/authentication/index.js
warm-coolguy: Oh, the method `authenticate` is also supposed to do the logout. I expected this to be a token refresh since it calls `revokeToken` that has `grant_type=refresh_token` hard-wired. Now that I look at it, the fetch also always uses the URL `''`? Is this just unfinished? I guess we come back around to this when the other issues are resolved, the code may have changed until then anyway.

warm-coolguy:
> > 🏓 @dopenguin Unfinished review, I'll add more later.
> > All in all, I do not see how an application is supposed to use us properly. It seems to me that we require both a configuration parameter for the token and an update mechanism to replace it in time that the login-managing application is required to call in time. Or did I just not see the idea here?
> 
> As mentioned in [#249 (comment)](https://github.com/Dataport/polar/pull/249#discussion_r1969546572), the token requests should be done by the leading application and not POLAR.
> 
> 🏓 @warm-coolguy

🏓 @dopenguin

packages/clients/diplan/example/authentication/index.js
dopenguin: I can take another look regarding the grant_type (as in: is it the correct one).
The URL is the same thing as with the rest: I'll add this to the repository once we have one that can be put into OSS. Until then, use the URL I provided you through a save channel.

packages/clients/diplan/example/authentication/cookie.js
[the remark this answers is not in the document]
warm-coolguy: >Do you think an adjustment to the documentation is necessary so that it is clear that interceptorUrlRegex should only be used by integrated clients as we currently do not offer a login module?

I'd like it to be used in combination with a config parameter and an obligation to refresh it via action, which is a mechanism that should be documented, but is debated atm.

packages/clients/diplan/example/authentication/index.js
warm-coolguy: Oh right, I forgot that it's in the git patch.

dopenguin: The url used ends with `openid-connect/revoke`. The `grant_type` is also correct for this case.

packages/clients/diplan/example/authentication/cookie.js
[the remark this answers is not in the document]
dopenguin: f2b199f7d3bfec08558fd78960e3f0ec3212fc75

packages/clients/diplan/example/authentication/cookie.js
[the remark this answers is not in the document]
dopenguin: The name of the token can now be configured, as mentioned in https://github.com/Dataport/polar/pull/249#discussion_r1971974735

Whether the authentication shall be achieved by retrieving a cookie or giving the JWT value via createMap and handling the refresh in POLAR (as well) is now something I am discussing with the relevant person that will integrate POLAR later on.

dopenguin:
> > > 🏓 @dopenguin Unfinished review, I'll add more later.
> > > All in all, I do not see how an application is supposed to use us properly. It seems to me that we require both a configuration parameter for the token and an update mechanism to replace it in time that the login-managing application is required to call in time. Or did I just not see the idea here?
> > 
> > 
> > As mentioned in [#249 (comment)](https://github.com/Dataport/polar/pull/249#discussion_r1969546572), the token requests should be done by the leading application and not POLAR.
> > 🏓 @warm-coolguy
> 
> 🏓 @dopenguin

Except https://github.com/Dataport/polar/pull/249#discussion_r1969150117 everything should be g2g now.

🏓 @warm-coolguy

warm-coolguy:
> > > > 🏓 @dopenguin Unfinished review, I'll add more later.
> > > > All in all, I do not see how an application is supposed to use us properly. It seems to me that we require both a configuration parameter for the token and an update mechanism to replace it in time that the login-managing application is required to call in time. Or did I just not see the idea here?
> > > 
> > > 
> > > As mentioned in [#249 (comment)](https://github.com/Dataport/polar/pull/249#discussion_r1969546572), the token requests should be done by the leading application and not POLAR.
> > > 🏓 @warm-coolguy
> > 
> > 
> > 🏓 @dopenguin
> 
> Except [#249 (comment)](https://github.com/Dataport/polar/pull/249#discussion_r1969150117) everything should be g2g now.
> 
> 🏓 @warm-coolguy

Alright. Since that's a pretty fundamental thing, I'll wait for the resolution to that.

packages/clients/diplan/example/authentication/cookie.js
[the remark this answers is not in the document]
warm-coolguy: Closed due to changes and re-review.

packages/core/src/utils/createMap/addInterceptor.ts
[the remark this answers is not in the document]
warm-coolguy: Closed due to changes and re-review.

packages/clients/diplan/example/authentication/index.js
warm-coolguy: Closed due to changes and re-review.

packages/clients/diplan/example/authentication/cookie.js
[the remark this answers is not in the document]
warm-coolguy: Closed due to changes and re-review.

packages/clients/diplan/example/polar-ui/config.js
```diff
@@ -65,6 +65,7 @@ export default {
     minLength: 3,
     waitMs: 300,
   },
+  interceptorUrlRegex: '',
```
warm-coolguy, line 68: It doesn't really makse sense to add the login example to the `polar-ui` client since we have no service we could use in the test environment at [our GitHub page](https://dataport.github.io/polar/docs/diplan/example/polar-ui/). This would, as it currently is, result in dead UI in the example.

packages/core/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Feature: Add `singleTile` as as usable parameter in the configuration of WMS-layers.
 - Feature: Additionally export `MapInstance` type.
+- Feature: Add possibility of using OIDC secured services by using the configuration parameter `authentication`.
```
warm-coolguy, line 7: No such parameter seems to exist.

packages/core/README.md
```diff
@@ -91,6 +91,7 @@ The mapConfiguration allows controlling many client instance details.
 | checkServiceAvailability | boolean? | If set to `true`, all services' availability will be checked with head requests. |
 | extendedMasterportalapiMarkers | extendedMasterportalapiMarkers? | Optional. If set, all configured visible vector layers' features can be hovered and selected by mouseover and click respectively. They are available as features in the store. Layers with `clusterDistance` will be clustered to a multi-marker that supports the same features. Please mind that this only works properly if you configure nothing but point marker vector layers styled by the masterportalapi. |
 | featureStyles | string? | Optional path to define styles for vector features. See `mapConfiguration.featureStyles` for more information. May be a url or a path on the local file system. |
+| interceptorUrlRegex | string | Regular expression defining URLs that belong to secured services. All requests sent to URLs that fit the regular expression will send the JSON Web Token (JWT) found in the store parameter `oidcToken` as a Bearer token in the Authorization header of the request. Requests already including a Authorization header will keep the already present one. |
```
warm-coolguy, line 94: This name is too technical. It describes that a URL regex can be set to intercept (requests), but is not telling of what one would configure it to do. I suggest `secureServiceUrlRegex`, since the underlying implementation (intercepting) does not matter to users.

packages/core/src/components/MapContainer.vue
```diff
@@ -168,6 +169,7 @@ export default Vue.extend({
     }
     addEventListener('resize', this.updateHasSmallDisplay)
     this.updateHasSmallDisplay()
+    this.addInterceptor()
```
warm-coolguy, line 172: This should be done earlier. Best would be to do it before the map is even initialized so we can, in the future, prevent any overlaps where a) the map is already loading tiles b) the interceptor is not set up. This may especially break if any async behaviour is introduced in between them, and will be hard to debug.

packages/core/src/vuePlugins/actions/addInterceptor.ts
```diff
@@ -0,0 +1,34 @@
...
+    if (
+      interceptorUrlRegex &&
+      typeof resource === 'string' &&
+      resource.match(interceptorUrlRegex)
+    ) {
+      config = {
+        ...originalConfig,
+        headers: {
+          // eslint-disable-next-line @typescript-eslint/naming-convention
+          Authorization: `Bearer ${getters.oidcToken}` || '',
```
warm-coolguy, line 26:
```suggestion
    if (
      getters.oidcToken &&
      interceptorUrlRegex &&
      typeof resource === 'string' &&
      resource.match(interceptorUrlRegex)
    ) {
      config = {
        ...originalConfig,
        headers: {
          // eslint-disable-next-line @typescript-eslint/naming-convention
          Authorization: `Bearer ${getters.oidcToken}`,
```
Didn't work, I suppose you meant that.

packages/core/src/components/MapContainer.vue
warm-coolguy, line 172: This should be done optionally and only if an `interceptorUrlRegex` is present. We should not mess with fetch unless we're required to do so.

packages/core/README.md
```diff
@@ -548,6 +549,17 @@ This is, for example, useful to listen to search results, draw features, or mark
 
 To add content to the `MoveHandle`, the mutation `setMoveHandle` can be used. The values needed are described in `@polar/lib-custom-types:MoveHandleProperties`.
 
+### Mutations
```
warm-coolguy, line 552: I do not consider the mutation to be sufficient to solve the issue at hand. The environment will have to a) initialize the map client b) use `setOidcToken` to hand us the token to use. If a secured service is configured to be initially visible, the map may call e.g. tiles with the token missing.

Hence, the initial oidcToken must also be configurable.

The mutation is still needed to update tokens.

warm-coolguy (changes requested):
🏓 @dopenguin 

Since we can't provide a reproducible example on our example page and since this is directly relevant to the usage of the DiPlan client in its target environment, please add a paragraph on how to use it to the DiPlan's `API.md` (after all conversations have been resolved, since the details may change by then).

packages/clients/diplan/example/authentication/index.js
```diff
@@ -0,0 +1,65 @@
...
+let loggedIn = false
+let token
+
+export async function authenticate(username, password, setTokenMethod) {
+  if (loggedIn) {
+    await reset()
+    return
+  }
+
+  const url = ''
```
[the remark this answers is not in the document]
dopenguin: Done 896cfab235364a80ff51ba6d2ef61bd9a93fef4c

packages/clients/diplan/example/authentication/index.js
```diff
@@ -0,0 +1,65 @@
...
+  // TODO: Add UI element to a layer if it can only be used via authentication (lock / unlock)
+}
+
+async function reset() {
+  await revokeToken()
+  window.location.reload()
+}
+
+async function revokeToken() {
+  const url = ''
```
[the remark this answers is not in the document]
dopenguin: Done 896cfab235364a80ff51ba6d2ef61bd9a93fef4c

packages/clients/diplan/example/polar-ui/config.js
dopenguin: Removed from POLAR-UI

I'll add an example page that focuses on the login and the secured service.
As we currently have not publicly facing service, this will still be a dead UI though.

packages/core/README.md
dopenguin: Changed the name 9e1a729c060e7793095e3dc7030eba932c8b5294

packages/core/CHANGELOG.md
dopenguin: Now includes the correct parameter name ef0f1760f4ca3b42050247140830d2a64e321749

packages/core/src/components/MapContainer.vue
dopenguin: Moved it up 91d1c2aff5eaff50d1dceb0c551d2e802ffccee8

dopenguin: Adjusted d0b7453478e02346b9c9cec1c22082386d91e0b2

packages/core/README.md
dopenguin: Adjusted the workflow 40d1a198789b11b63895a3c4e019853cfbafecc7

dopenguin:
> 🏓 @dopenguin
> 
> Since we can't provide a reproducible example on our example page and since this is directly relevant to the usage of the DiPlan client in its target environment, please add a paragraph on how to use it to the DiPlan's `API.md` (after all conversations have been resolved, since the details may change by then).

Done 7a6ec7f1cd268360b5af60dadb9d1281dfa6e6d4

🏓 @warm-coolguy

packages/clients/diplan/example/polar-ui/config.js
warm-coolguy: seriously

That didn't suffice.

Yeah, then that's a future effort.

packages/core/src/vuePlugins/actions/addInterceptor.ts
warm-coolguy: `Bearer ${getters.oidcToken}` || ''`

This statement never falls back to '', no matter what value `getters.oidcToken` holds.

packages/core/README.md
warm-coolguy: Unsure about the solution. It would be better to have this available before `createMap` is called with the same reason as for `addInterceptor`. However, I checked, and it effectively does not matter at the time being. But we may run into a run-time issue with this later on.

🎩 since I'm unsure of the probability of this happening.

packages/clients/diplan/example/polar-ui/config.js
dopenguin: Now for real 238bab4a95fa68603184efbd05dcdf4c0e8b2480 aaa104c7231f5b3a4f3d70d221408ca442c996d6

packages/core/src/vuePlugins/actions/addInterceptor.ts
dopenguin: Adjusted 7abee88c097e0c2b484e046315bd2acffe12e402

dopenguin:
> > > 🏓 @dopenguin
> > > Since we can't provide a reproducible example on our example page and since this is directly relevant to the usage of the DiPlan client in its target environment, please add a paragraph on how to use it to the DiPlan's `API.md` (after all conversations have been resolved, since the details may change by then).
> > 
> > 
> > Done [7a6ec7f](https://github.com/Dataport/polar/commit/7a6ec7f1cd268360b5af60dadb9d1281dfa6e6d4)
> > 🏓 @warm-coolguy
> 
> 🏓 @dopenguin

🏓 @warm-coolguy

warm-coolguy (approved):
> > > > 🏓 @dopenguin
> > > > Since we can't provide a reproducible example on our example page and since this is directly relevant to the usage of the DiPlan client in its target environment, please add a paragraph on how to use it to the DiPlan's `API.md` (after all conversations have been resolved, since the details may change by then).
> > > 
> > > 
> > > Done [7a6ec7f](https://github.com/Dataport/polar/commit/7a6ec7f1cd268360b5af60dadb9d1281dfa6e6d4)
> > > 🏓 @warm-coolguy
> > 
> > 🏓 @dopenguin
> 
> 🏓 @warm-coolguy 

:shipit: @dopenguin

PR #251, opened 2025-02-27
prepare diplan example folder for multiple prod instances

warm-coolguy:
## Summary

I've split the example setup so that we have a `dev` client and a `prod` client that may now deviate. More `prod` clients are supposed to trickle in over time so that we can easily produce multiple examples regarding the various contained features. With this, it is no longer required to keep the `dev` and `prod` `.html` files in sync.

For the time being, the two clients are mostly copies of each other except for the code source. This will change during the course of the development to follow.

Check both dev mode and the produced pages.


packages/clients/diplan/example/README.md
```diff
@@ -0,0 +1,13 @@
+# Example structure
+
+## `.`
+
+The root folder holds files used in all clients. Modify with care. If you need wildly different files for an example, create new files in the relevant example folder.
+
+## `dev`
+
+This folder holds the development example used during development time. You may put anything in here as long as it serves the testing purposes. The client is not required to look good, but is a technical playground, aka "just trash in whatever you want to test, the end of the universe will clean it up".
```
dopenguin, line 9:
```suggestion
This folder holds the development example used during development time. You may put anything in here as long as it serves the testing purposes. The client is not required to look good, but is a technical playground.
```
As it is customer facing

packages/clients/diplan/example/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+## all others
+
+All other folders are examples used in the documentation deployed on the GitHub page. They are supposed to display usage variants and should be understandable/clean. These are supposed to be viewn both by the customer and integrators alike so that they can easily lift from the usage scenarios. They're also useful for testing PRs in build mode.
```
dopenguin, line 13: A link to the github page would be neat here.

dopenguin (changes requested):
Nothing big to say, sounds like a good idea.

🏓 @warm-coolguy

packages/clients/diplan/example/README.md
warm-coolguy: https://github.com/Dataport/polar/pull/251/commits/600aba5058e012e78d31f1058e01ea6b92f3a40d

warm-coolguy:
> Nothing big to say, sounds like a good idea.
> 
> 🏓 @warm-coolguy

🏓 @dopenguin

dopenguin (approved):
🚀 @warm-coolguy

PR #254, opened 2025-02-28
Feature/diplan UI

dopenguin:
## Summary

Various changes towards making the UI interchangeable for diplan.

`npm run diplan:dev` and check whether all the functionality that is already implemented works as intended.

- [x] Screenreader functionality has been manually tested with NVDA <- I still haven't looked for a good alternative to NVDA on Mac, please do this for me @warm-coolguy

  - [x] Chrome Lighthouse
  - [x] Firefox Accessibility

Branch is based on #252


dopenguin:
The PR is ready for review!

Please mind the (still) temporary hack in `packages/core/src/utils/createMap/pullPolarStyleToShadow.ts`. If the needed change is not done in time, we'd have to add your suggested `postinstall`-hack.

In a follow-up PR, there might UI of the groups of the address search might change but as we currently only have on service, this seemed redundant to do now.

🏓 @warm-coolguy

warm-coolguy (changes requested):
🏓 @dopenguin 

My review isn't complete, but I've found some things that may be addressed in parallel.

packages/clients/diplan/example/overview.html
```diff
@@ -17,7 +17,7 @@ <h1>Testseite @polar/client-diplan – Übersicht</h1>
         <h2 class="padded-x">Beispielübersicht</h2>
         <div class="padded-x">
           <li><a href="./polar-ui">POLAR UI test client</a></li>
-          <li>DiPlan UI test client <span class="todo">TODO</span></li>
+          <li><a href="./diplan-ui">DiPlan UI test client</a></li>
```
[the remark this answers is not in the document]
dopenguin: Gave the setup a complete revamp to be able to have all the different possibilities 31151b0eaf233ae2f1511a557025b343d6cbf4a8

packages/clients/diplan/API.md
```diff
@@ -16,6 +16,7 @@ DiPlan-specific configuration parameters belong within the `mapConfiguration` ob
 | - | - | - |
 | mergeToMultiGeometries | boolean? | Defaults to `false`. If `true`, the exported FeatureCollection in getter `revisedDrawExport` will have merged geometries; that is, instead of Points, Lines, and Polygons, only MultiPoints, MultiLines, and MultiPolygons will be featured, created by merging the features of their respective geometry. All geometry types that are enabled in the `Draw` tool may occur. This step is executed before geometry validation and meta service usage. |
 | metaServices | metaService[]? | Specification of a service that contains meta-information regarding geometries. Information from there will be added to features in getter `revisedDrawExport`. |
+| renderType | 'iconMenu' \| 'independent'? | If set to `independent`, the displayed view for the GeoEditing plugin is minified to a vertical list of buttons of icons instead of a list of buttons including descriptions. The plugin then may be used outside of an IconMenu. Defaults to `iconMenu`. |
```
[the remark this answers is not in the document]
dopenguin: Added two configs 5d74ede4666b57eb8a4bf95292ee285c9817288b

packages/clients/diplan/src/plugins/AddressSearch/Input.vue
```diff
@@ -0,0 +1,50 @@
+<template>
```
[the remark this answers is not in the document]
dopenguin: Re-added 208b21088254931f9c58801ccef9f15b8c48f0e4

I initially removed it as the design didn't intend to leave any space for that. But as we already include more space because of the information showing below the input, this seems like the correct approach.

dopenguin:
I've added another button for `Translate` ca1896a13624587953d336522d9f371525e4b5e2, as the combined approach mentioned in https://github.com/Dataport/polar/pull/254#issuecomment-2718943909 is not a good approach UX-wise.

packages/core/src/utils/createMap/pullPolarStyleToShadow.ts
```diff
@@ -10,7 +10,18 @@ export const pullPolarStyleToShadow = (
     const polarStylesheets = stylesheets.filter((el) =>
       el.getAttribute('data-vite-dev-id')
     )
-    polarStylesheets.forEach((style) => shadowRoot.appendChild(style))
+    // TODO: This is a temporary workaround until diplanung-style provides the rule this way; this should not find it's way onto main
```
warm-coolguy, line 13:
```suggestion
    // TODO: This is a temporary workaround until diplanung-style provides the rule this way; this should not find its way onto main, but it will, and, if you read it, probably did
```
Keeping it real and grammatically correct.

packages/core/src/components/layouts/NineLayout.vue
```diff
@@ -62,7 +62,7 @@ export default Vue.extend({
   .nine-layout {
     display: flex;
     flex-direction: column;
-    justify-content: flex-start;
+    justify-content: center;
```
warm-coolguy, line 65: Nested rule `&.mid-mid` now pointless and up for removal.

packages/clients/diplan/src/style.css
```diff
@@ -0,0 +1,124 @@
+/* NOTE: This styling is only a concern for the example environment and does
+ * not directly modify what the POLAR client looks like. Styling the client
+ * itself is done in the src folder. */
```
warm-coolguy, line 3: Why is the example environment style in the source folder now?

packages/clients/diplan/src/mode.ts
```diff
@@ -0,0 +1,10 @@
+export const MODE = {
+  // display for small, integrated view
+  DIPLAN_SMALL: 'DIPLAN_SMALL',
```
warm-coolguy, line 3: This has no example. Intentional?

warm-coolguy (changes requested):
🏓 @dopenguin 

Another partial review. I'll proceed tomorrow.

packages/clients/diplan/example/polar-ui/setup.js
```diff
@@ -107,17 +108,6 @@ export default (client, layerConf, config) => {
...
-        htmlExportImg.setAttribute('src', screenshot)
-        if (navigator.userAgent.toLowerCase().includes('firefox')) {
-          htmlExportImg.onclick = function () {
-            window.open(this.src, '_blank')
-          }
-          htmlExportA.onclick = () => false
-        } else {
-          htmlExportA.setAttribute('href', screenshot)
-        }
-      })
```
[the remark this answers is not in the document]
dopenguin: New information concluded that it should be available programmatically, but not in the UI.

Should I add this here or can we leave the removal and have this be done in a separate PR? This PR has grown a bit already.

packages/clients/diplan/example/polar-ui/index.html
```diff
@@ -86,16 +86,14 @@ <h2>Informationen aus Kartenklient</h2>
...
-    <!-- → window.PolarClientDiPlan is now available. -->
     <link rel="stylesheet" src="../../dist/polar-client.css">
     <script type="module">
+      import diplanMapClient from '../../dist/client-diplan.js'
       import services from '../services.js'
       import config from './config.js'
       import setup from './setup.js'
 
-      setup(PolarClientDiPlan, services, config)
+      setup(diplanMapClient, services, config)
```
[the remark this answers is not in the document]
dopenguin: When I transitioned to the new build mode (so that we can have the different components usable when built) I based the implementation on `@polar/client-meldemichel`.
I also don't seem to get this to work if the root is set to `src` instead of `example` as `PolarClientDiPlan` is not available.

Thus, I don't deem as necessary. The target environment is also using `vue` and is installing the package via `npm`, so importing this here doesn't seem like something that would cause problems.

packages/clients/diplan/src/style.css
dopenguin: It is a copy.
When running `npm run diplan:dev` and importing the file from `example`, the styles are not loaded.

packages/clients/diplan/example/polar-ui/index.html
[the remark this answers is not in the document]
warm-coolguy: I mailed the embedding programmer regarding this. We have the script tag in some of the docs, so maybe they copied the solution from there. There's a lot of stuff that may happen, even with NPM. 😄

Should the answer arrive that they use `import`, this is a 🎩. I also asked whether they can switch to `import` if they're using the `<script>` tag solution. They probably favour that anyway.

packages/types/custom/core.ts
```diff
@@ -419,9 +421,14 @@ export interface Menu {
 
 export interface IconMenuConfiguration extends PluginOptions {
   menus: Menu[]
+  component?: VueConstructor
   initiallyOpen?: string
 }
 
+export interface LayerChooserConfiguration {
```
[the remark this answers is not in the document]
warm-coolguy: While we never used it, it does work out of the box due to all plugins with UI working in the same fashion.

Example:

```
polarCore.addPlugins([
  LayerChooser({
    displayComponent: true,
    layoutTag: NineLayoutTag.MIDDLE_MIDDLE,
  }),
])
```

heh

This renders the documentation effectively incomplete OR announces a breaking change that goes undocumented here. Should you desire to drop support for the usual PluginOptions, please do that as a separate effort and keep the type intact here.

packages/clients/diplan/example/polar-ui/index.html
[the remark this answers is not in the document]
warm-coolguy: They're using `import * as client from '@polar/client-diplan'`.

packages/types/custom/core.ts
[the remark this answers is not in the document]
dopenguin: I_guess


b1459b2d4b56f281138c648fe3223a52169860a8

packages/clients/diplan/example/polar-ui/setup.js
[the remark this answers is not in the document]
dopenguin: Re-added 2098257dfd1e4de0c3dccb9a6c5f423970413d29

packages/types/custom/core.ts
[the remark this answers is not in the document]
warm-coolguy: Purr. 🎩

packages/clients/diplan/src/style.css
warm-coolguy: https://github.com/Dataport/polar/pull/276

Setup shenanigans, probably required to fix all of that before continuing here.

dopenguin: My guess is that it is just not intended to load styles from a different path, e.g. `node_modules/diplanung-style/dist/style.css` is also not loaded. I'll remove that import.

warm-coolguy: Yeah, that's a security constraint of web servers. They're not supposed to expose stuff outside of their scope. The bundler goes everywhere, though.

packages/clients/diplan/example/polar-ui/index.html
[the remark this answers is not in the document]
warm-coolguy: Since behaviour is toggling a little, I'll leave this thread open to check at the end what's happened.

packages/clients/diplan/src/addPlugins.ts
```diff
@@ -0,0 +1,129 @@
...
+      mode !== MODE.DIPLAN_SMALL &&
+        Scale({
+          displayComponent: true,
+          layoutTag: NineLayoutTag.BOTTOM_RIGHT,
+        }),
+      mode !== MODE.DIPLAN_SMALL &&
+        Toast({
+          displayComponent: true,
+          layoutTag: NineLayoutTag.BOTTOM_MIDDLE,
+        }),
```
warm-coolguy, line 126: This one may bite and is invisible, probably better to just keep it always in.

packages/core/src/components/layouts/NineLayoutTag.ts
```diff
@@ -3,7 +3,7 @@ export enum NineLayoutTag {
   TOP_MIDDLE = 'top-mid mid top nine-layout',
   TOP_RIGHT = 'top-right top right nine-layout',
   MIDDLE_LEFT = 'mid-left mid left nine-layout',
-  MIDDLE_MIDDLE = 'mid-mid mid nine-layout',
+  MIDDLE_MIDDLE = 'mid nine-layout',
```
warm-coolguy, line 6:
```suggestion
  MIDDLE_MIDDLE = 'mid-mid mid nine-layout',
```
Many of those classnames aren't actually used. They all share a pattern, and there's no reason to break it now. If you desire to change how NineLayout works, please do that as a separate effort.

packages/core/src/utils/createMap/pullPolarStyleToShadow.ts
```diff
@@ -10,7 +10,18 @@ export const pullPolarStyleToShadow = (
...
+    polarStylesheets.forEach((style) => {
+      if (
+        style.attributes['data-vite-dev-id'].textContent.endsWith(
+          'diplanung-style/src/scss/main.scss'
+        )
+      ) {
+        style.textContent = `:host,${style.textContent}`
+      }
+      shadowRoot.appendChild(style)
+    })
```
warm-coolguy, line 24: TODO, shouldn't reach main, see https://github.com/Dataport/polar/pull/277 for how it can be done.

warm-coolguy (changes requested):
🏓 @dopenguin

I kinda skipped reviewing the copied components. I'll check them from the UI side, but not the code side, due to time constraints. Have the copies changes regarding their template? If not, an implementation as wrapper components that use the plugins' template and override the style by e.g. `v-deep` (or however it's called/written nowadays) may be better than the duplication. Your call.

UI/UX/a11y check starting next time I come around to this.

dopenguin:
> 🏓 @dopenguin
> 
> I kinda skipped reviewing the copied components. I'll check them from the UI side, but not the code side – or do you see a need for that?
> 
> Have the copies changed regarding their template? If not, an implementation as wrapper components that use the plugins' template and override the style by e.g. `v-deep` (or however it's called/written nowadays) would be preferable to the copies; they _will_ diverge, absolutely, and it will hurt a lot.
> 
> UI/UX/a11y check starting next time I come around to this.

Some components also include changes to their template, not only their style. Because of that, diverging seems to be an issue we have to communicate.

packages/clients/diplan/src/style.css
dopenguin: The solution at hand is based on `@polar/client-meldemichel`. So, regarding the time constraint, this issue does not seem relevant right now for me and should be tackled together with changing the setup in meldemichel, if the need persists.

packages/core/src/components/layouts/NineLayoutTag.ts
dopenguin: In another comment, you suggested me to delete the class name as it became redundant. I can surely re-add it here, but shouldn't the class be re-added as well?

packages/clients/diplan/src/addPlugins.ts
dopenguin: Aight 9365506291a8ab2129fa8c7da61f8de3dde6033c

packages/clients/diplan/src/style.css
warm-coolguy: Meldemichel is not comparable to DiPlan since the delivery object there includes a `.html` file in parallel to a self-integration variant. We don't plan a standalone DiPlan client. I advise to at least nest the files only required by the `.html` file in DiPlan somehwere deeper instead of having them right next to the files that are part of the bundle.

🎩 at your own discretion.

packages/core/src/components/layouts/NineLayoutTag.ts
warm-coolguy: I suggested to delete a now-redundant rule that pertained to the class, not to delete the class itself.

MagrittePipe

packages/clients/diplan/src/style.css
dopenguin: Moving this to e.g. `src/example` leads to a constant reload-loop for me; using different [cli options](https://vite.dev/guide/cli) didn't change that for me.
You can take another look at this after the next release, if you deem this relevant.

dopenguin:
As discussed earlier today, the menu now deactivates any active tool if the user closes the window 4d35f92e1f044db0dae0c00cee64c93c5435813a

packages/core/src/components/layouts/NineLayoutTag.ts
warm-coolguy: Surely that must be a joke.

warm-coolguy (changes requested):
🏓 @dopenguin 

I largely question the need for the component copies as it has been implemented. It's mostly solvable by either prop forwarding/spreading or usage of the `:deep` selector in a wrapper component. I do not consider this approach maintainable in the long run. We should probably have some deep thoughts about this in a decisive round. (Stopped reviewing since I started repeating myself, it's all that issue.)

packages/clients/diplan/src/plugins/AddressSearch/Results.vue
```diff
@@ -0,0 +1,232 @@
+<template>
```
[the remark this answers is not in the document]
dopenguin: search

Can't reproduce that bug

packages/clients/diplan/src/plugins/AddressSearch/index.ts
```diff
@@ -0,0 +1 @@
+export { default as AddressSearch } from './AddressSearch.vue'
```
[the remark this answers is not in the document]
dopenguin: Adjusted a2434ea588ce6c0920f23eacfecd1812de617fc1

packages/clients/diplan/src/plugins/AddressSearch/AddressSearch.vue
```diff
@@ -0,0 +1,34 @@
+<template>
+  <SearchContainer id="diplan-search-container" />
+</template>
+
+<script lang="ts">
+import Vue from 'vue'
+// TODO: This import should be changed in the future
```
warm-coolguy, line 7: Is this comment obsolete?

packages/clients/diplan/src/plugins/AddressSearch/AddressSearch.vue
```diff
@@ -0,0 +1,34 @@
...
+  name: 'DiplanAddressSearch',
+  components: {
+    SearchContainer,
+  },
+})
+</script>
+
+<style lang="scss">
+#diplan-search-container {
+  .v-toolbar {
```
warm-coolguy, line 20:
```suggestion
#diplan-search-container {
  z-index: 1;
  .v-toolbar {
```

This way, the result box stops underlapping the attributions.

packages/clients/diplan/src/plugins/AddressSearch/Input.vue
```diff
@@ -0,0 +1,49 @@
...
```
warm-coolguy, line 1: This file seems to be obsolte now.

packages/clients/diplan/src/plugins/AddressSearch/Results.vue
warm-coolguy, line 1: This file seems to be obsolete now.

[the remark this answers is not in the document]
warm-coolguy: With the new solution, it no longer happens. Odd why it happened at all. Maybe it was some kind of build hiccup. 🎩

packages/clients/diplan/src/plugins/LayerChooser/Options.vue
```diff
@@ -0,0 +1,120 @@
+<template>
```
warm-coolguy, line 1: This seems to be 100% identical to the plugin component. I propose it is imported rather than copied.

packages/clients/diplan/src/plugins/LayerChooser/LayerWrapper.vue
```diff
@@ -0,0 +1,90 @@
+<template>
```
warm-coolguy, line 1: This seems to be, apart from the icon, identical to the plugin component. I propose it is imported rather than copied, and that the icon override need is adressed in the same fashion as it is for `Attributions.icons`.

packages/clients/diplan/src/plugins/GeoEditing/locales.ts
```diff
@@ -0,0 +1 @@
+export default {}
```
warm-coolguy, line 1: This seems to be a leftover that can safely be removed.

packages/clients/diplan/src/plugins/GeoEditing/index.ts
```diff
@@ -0,0 +1,16 @@
...
+import { GeoEditing } from './components'
+// import locales from './locales'
+// import { makeStoreModule } from './store'
+
+// TODO: Are additional configuration parameters needed?
+export default (options: PluginOptions) => (instance: Vue) =>
+  instance.$store.dispatch('addComponent', {
+    name: 'geoEditing',
+    plugin: GeoEditing,
+    // locales,
```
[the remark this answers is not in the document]
dopenguin: I initially intended to throw all the parts of the geoEditing store in here.
But I am a bit unsure if that will still happen as we should rather add relevant functions to `@polar/plugin-draw` instead.

packages/clients/diplan/src/plugins/GeoEditing/locales.ts
dopenguin: Done 279eca8f54acf7ee7e24200aed67831b1a46d1e9

packages/clients/diplan/src/plugins/AddressSearch/Results.vue
dopenguin: Done 279eca8f54acf7ee7e24200aed67831b1a46d1e9

packages/clients/diplan/src/plugins/AddressSearch/Input.vue
dopenguin: Done 279eca8f54acf7ee7e24200aed67831b1a46d1e9

packages/clients/diplan/src/plugins/AddressSearch/AddressSearch.vue
dopenguin: Jup, added it when I hadn't exposed this 16fc3a47500e42a3e7c1b6906ae93c6d5e2135dd

packages/clients/diplan/src/plugins/Attributions/AttributionButton.vue
```diff
@@ -0,0 +1,49 @@
+<template>
```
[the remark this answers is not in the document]
dopenguin: Adjusted 64ed372a28a43fbc6e725247fd99b863e7038a24

packages/clients/diplan/src/plugins/GeoEditing/index.ts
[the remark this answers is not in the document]
warm-coolguy: Well, then it's up to you. 🎩

packages/clients/diplan/src/plugins/Zoom/Zoom.vue
```diff
@@ -0,0 +1,89 @@
+<template>
```
warm-coolguy, line 1: I think this, too, can be deduplicated by adding the props `zoomInButtonProps` and `zoomOutButtonProps` and using `v-bind` in those places respectively.

packages/clients/diplan/src/plugins/IconMenu/IconMenu.vue
```diff
@@ -0,0 +1,182 @@
+<template>
```
warm-coolguy, line 1: This, too, looks deduplicatable.

packages/clients/diplan/src/plugins/GeoEditing/index.ts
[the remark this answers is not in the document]
dopenguin: Removed 1570a05d3965785e4127a81a272facfd8c453025

packages/clients/diplan/src/plugins/Zoom/Zoom.vue
dopenguin: Done 131706c014b9fdbf1e41414358f67db9b7776f5b & d64ea9fe225fb64c4950f86e69c56a704e90c3fe

packages/clients/diplan/src/plugins/IconMenu/IconMenu.vue
dopenguin: I've fiddled around with this a bit and to me it seems like the approach I've used for the other plugins does not seem to fit here.

- Using `:deep(.v-btn)` screws with styles of buttons inside any integrated plugin; here: GeoEditing
- To be able to remove the normally needed class `.icon-menu-list-item-content-scrollable-y` the following snippet is something that might be of help, but is something I wouldn't approve if somebody else showed me that as their solution
- Reversing the color scheme for the buttons seems awful to implement as well...

```js
watch: {
  open(newVal: number | null) {
    this.$nextTick(() => {
      if (typeof newVal === 'number') {
        this.$children[0].$refs['item-component'][0].$el.className =
          this.$children[0].$refs['item-component'][0].$el.className.replace(
            /icon-menu-list-item-content-scrollable-y/,
            ' '
          )
      }
    })
    },
  },
```

The only solution that would come to my mind is to add the possibility to change the IconMenu-Buttons instead of the complete component and make aforementioned class configurable.

If you got a different solution regarding the class, please let me know!

packages/clients/diplan/src/plugins/LayerChooser/Options.vue
dopenguin: ✅  bf150c9da506ab49da70d55cdf4b9dbc7e87a90a

packages/clients/diplan/src/plugins/LayerChooser/LayerWrapper.vue
dopenguin: It also needed some style adjustments so I went the way of adding the icon via a optional prop 0ef61b217b56adba9ce178e35ce2774525e1704b

packages/core/src/utils/createMap/pullPolarStyleToShadow.ts
warm-coolguy: IIRC we dropped the `font-family` for now since it's a little fiddly atm, and `polar-ui` is optional.

Regarding `font-family`, I checked the `dist/style.css` in that package and `OpenSans` seems to be included as base64. So loading it is not the issue. I don't know what it is, though.

Regarding `polar-ui`, at the time being I'd consider it sufficient to just add a note in the UI somewhere that it is currently "a little off due to changes for diplanung-style" or something along that line so that viewers know why it is like that. Maybe right at the link in the overview page; or above the map? Or both ... both is good ...

warm-coolguy (approved):
@dopenguin :shipit: after addressing https://github.com/Dataport/polar/pull/254#discussion_r2009501982 

It's kinda optional, but would be nice for clarity when someone looks that up.

packages/core/src/utils/createMap/pullPolarStyleToShadow.ts
dopenguin: Added 3999525878fd78c7e3c45d0d1b20ab32ae03d478

PR #258, opened 2025-03-05
Configuration changes for internal dish client

jedi-of-the-sea:
## Summary

Some configuration changes for the internal client.

Nothing really to test, but if you want you can start the client with mode = `INTERN` and have a look at the changes.

Note: The new configuration of `textStyle` will not have an effect yet as there seems to be a bug in the draw plugin.

  - [x] Chrome Lighthouse
  - [x] Firefox Accessibility


warm-coolguy (approved):
:shipit: @jedi-of-the-sea 

I've had a feeling that this bug was probably just about a misnamed variable and found the spot: https://github.com/Dataport/polar/pull/258/commits/a72af3593b4609227692ecab073b5a0ab7e5eb91

Since the rest is merely configuration I assume it's as intended now.

PR #262, opened 2025-03-11
add screenshot to README.md

warm-coolguy:
## Summary

As per presentation feedback, we figured a screenshot would be nice. I've chosen Dataport's Hamburg address and just used the snowbox in a small format. Do you have another suggestion regarding image/wording? I'm pretty open to anything here.


dopenguin (approved):
:shipit: @warm-coolguy

PR #264, opened 2025-03-11
show cursor pointer on delete interaction

warm-coolguy:
## Summary

When hovering a feature during the activeness of the delete interaction, the cursor now appears as "pointer" when clicking would delete something. This is to have it in sync with the other interactions that indicate activeness by either displaying a point at the cursor or changing the cursor itself.

Run a client with draw deletion active and try it.

- [x] Functionality has been tested in Firefox, Chrome, Safari (@dopenguin Please test on MacOS)
- [ ] ~~Functionality has been tested on a smartphone~~ does not apply
- [ ] ~~Screenreader functionality has been manually tested with NVDA~~ does not apply

  - [ ] ~~Chrome Lighthouse~~ does not apply
  - [ ] ~~Firefox Accessibility~~ does not apply


packages/plugins/Draw/src/store/createInteractions/createDeleteInteractions.ts
```diff
@@ -1,19 +1,36 @@
...
 import VectorLayer from 'ol/layer/Vector'
 import Interaction from 'ol/interaction/Interaction'
 import { DragBox, Select } from 'ol/interaction'
 import { platformModifierKeyOnly } from 'ol/events/condition'
+import { PolarActionContext } from '@polar/lib-custom-types'
+import { Map } from 'ol'
+import { CreateInteractionsPayload, DrawGetters, DrawState } from '../../types'
+
+const pointerStyle = (map: Map, drawLayer: VectorLayer) => (e) => {
+  const found = map.forEachFeatureAtPixel(e.pixel, () => true, {
```
dopenguin, line 10: The method [`hasFeatureAtPixel`](https://openlayers.org/en/latest/apidoc/module-ol_Map-Map.html#hasFeatureAtPixel) seems to be the thing that fits best

packages/plugins/Draw/src/store/createInteractions/createDeleteInteractions.ts
```diff
@@ -1,19 +1,36 @@
...
+    layerFilter: (l) => l === drawLayer,
+  })
+
+  if (found) {
+    map.getTargetElement().setAttribute('style', 'cursor: pointer')
```
dopenguin, line 15: Do you think it would be helpful to use a trashicon here? https://developer.mozilla.org/en-US/docs/Web/CSS/cursor#url would do the trick wich `'pointer'` as the fallback. If you think this is redundant 🎩

dopenguin (changes requested):
Works great, just two small suggestions

🏓 @warm-coolguy

packages/plugins/Draw/src/store/createInteractions/createDeleteInteractions.ts
warm-coolguy: Closing due to reasons e.g. seen in: https://ericwbailey.website/published/dont-use-custom-css-mouse-cursors/

warm-coolguy:
> Works great, just two small suggestions
> 
> 🏓 @warm-coolguy

🏓 @dopenguin

dopenguin (approved):
:shipit: @warm-coolguy

PR #263, opened 2025-03-11
Feature/diplan merge polygons

warm-coolguy:
## Summary

- Add merge functionality for diplan tools
- Refactor a little how interactions are reset
- Blocks on https://github.com/Dataport/polar/pull/252

## Instructions for local reproduction and review

Run `npm run diplan:dev` and test the functionality.

- The "merge polygon" only merges things it intersects, not things it touches
- Polygons not touched by the "merge polygon" are not affected (so that deletion works as expected further down the work)
- Geometries that are neither Polygons nor MultiPolygons are not affected

## Pull Request Checklist (for Assignee)

- [ ] ~~Changelogs are maintained~~
- [ ] ~~Functionality has been tested in Firefox, Chrome, Safari~~
- [ ] ~~Functionality has been tested on a smartphone~~
- [ ] ~~Functionality has been tested with 200% screen zoom~~
- [ ] ~~Screenreader functionality has been manually tested with NVDA~~

UI has been tested in the following tools regarding accessibility (only regarding functionality affected in this PR)
  - [ ] ~~Chrome Lighthouse~~
  - [ ] ~~Firefox Accessibility~~

PR #265, opened 2025-03-11
implement geometry cloning

warm-coolguy:
## Summary

The "Duplicate polygon" button is now a broader "Duplicate geometry" button. It does what is says and shows its availability to do so with a "copy" cursor.

Blocks on https://github.com/Dataport/polar/pull/263

`npm run diplan:dev`, draw some geometries and clone them.

- [x] Functionality has been tested in Firefox, Chrome, Safari ← @dopenguin please test on MacOS
- [ ] ~~Screenreader functionality has been manually tested with NVDA~~ doesn't apply, actual UI will come in different PR

  - [ ] ~~Chrome Lighthouse~~ doesn't apply, actual UI will come in different PR
  - [ ] ~~Firefox Accessibility~~ doesn't apply, actual UI will come in different PR

### DIFF
--- a/packages/clients/diplan/src/store/geoEditing/duplicatePolygons.ts
+++ b/packages/clients/diplan/src/store/geoEditing/duplicatePolygons.ts
@@ -39,7 +39,7 @@ export const duplicatePolygons = ({
   const boundPointerStyle = pointerStyle(rootGetters.map, drawLayer)
   rootGetters.map.on('pointermove', boundPointerStyle)
 
-  selectedFeatures.on(['add'], () => {
+  selectedFeatures.on('add', () => {
     drawSource.addFeature(selectedFeatures.getArray()[0].clone())
     selectedFeatures.clear()
     dispatch('updateDrawMode', null)
--- a/packages/clients/diplan/src/store/geoEditing/duplicatePolygons.ts
+++ b/packages/clients/diplan/src/store/geoEditing/duplicatePolygons.ts
@@ -5,7 +5,7 @@ import { Map } from 'ol'
 import { DiplanGetters, DiplanState } from '../../types'
 
 const pointerStyle = (map: Map, drawLayer: VectorLayer) => (e) => {
-  const found = map.forEachFeatureAtPixel(e.pixel, () => true, {
+  const found = map.hasFeatureAtPixel(e.pixel, {
     layerFilter: (l) => l === drawLayer,
   })
 
--- a/packages/clients/diplan/src/store/geoEditing/duplicatePolygons.ts
+++ b/packages/clients/diplan/src/store/geoEditing/duplicatePolygons.ts
@@ -1,5 +1,4 @@
 import { PolarActionContext } from '@polar/lib-custom-types'
-
 import VectorLayer from 'ol/layer/Vector'
 import { Select } from 'ol/interaction'
 import { Map } from 'ol'
--- a/packages/clients/diplan/src/store/geoEditing/duplicatePolygons.ts
+++ b/packages/clients/diplan/src/store/geoEditing/duplicatePolygons.ts
@@ -1,22 +1,56 @@
 import { PolarActionContext } from '@polar/lib-custom-types'
+
+import VectorLayer from 'ol/layer/Vector'
+import { Select } from 'ol/interaction'
+import { Map } from 'ol'
 import { DiplanGetters, DiplanState } from '../../types'
 
+const pointerStyle = (map: Map, drawLayer: VectorLayer) => (e) => {
+  const found = map.forEachFeatureAtPixel(e.pixel, () => true, {
+    layerFilter: (l) => l === drawLayer,
+  })
+
+  if (found) {
+    map.getTargetElement().setAttribute('style', 'cursor: copy')
+  } else {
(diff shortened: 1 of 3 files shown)


packages/clients/diplan/src/store/geoEditing/duplicatePolygons.ts
```diff
@@ -1,22 +1,56 @@
 import { PolarActionContext } from '@polar/lib-custom-types'
+
```
dopenguin, line 2:
```suggestion
```

packages/clients/diplan/src/store/geoEditing/duplicatePolygons.ts
```diff
@@ -1,22 +1,56 @@
...
+import VectorLayer from 'ol/layer/Vector'
+import { Select } from 'ol/interaction'
+import { Map } from 'ol'
 import { DiplanGetters, DiplanState } from '../../types'
 
+const pointerStyle = (map: Map, drawLayer: VectorLayer) => (e) => {
+  const found = map.forEachFeatureAtPixel(e.pixel, () => true, {
```
dopenguin, line 9: Similar to https://github.com/Dataport/polar/pull/264#discussion_r1992111754, `hasFeatureAtPixel` seems to be the best fit here

packages/clients/diplan/src/store/geoEditing/duplicatePolygons.ts
```diff
@@ -1,22 +1,56 @@
...
+  const selectInteraction = new Select({ layers: [drawLayer] })
+  const selectedFeatures = selectInteraction.getFeatures()
+  // TODO temp solution, not actually _isDeleteSelect; normalizing these flags is part of a future effort
+  // @ts-expect-error | internal hack to detect it in @polar/plugin-pins and @polar/plugin-gfi
+  selectInteraction._isDeleteSelect = true
+
+  const boundPointerStyle = pointerStyle(rootGetters.map, drawLayer)
+  rootGetters.map.on('pointermove', boundPointerStyle)
+
+  selectedFeatures.on(['add'], () => {
```
dopenguin, line 43:
```suggestion
  selectedFeatures.on('add', () => {
```

packages/clients/diplan/src/store/geoEditing/duplicatePolygons.ts
```diff
@@ -1,22 +1,56 @@
...
+  dispatch('updateDrawMode', 'duplicate')
+
+  const drawSource = rootGetters['plugin/draw/drawSource']
+  const drawLayer = rootGetters.map
+    .getLayers()
+    .getArray()
+    .find(
+      (layer) =>
+        layer instanceof VectorLayer && layer.getSource() === drawSource
+    ) as VectorLayer
```
dopenguin, line 33: This seems awfully redundant.
Because of similar reasons, we've added `drawSource` as a getter. Mind doing the same for `drawLayer` as it seems to be needed from now on?

dopenguin (changes requested):
If I duplicate Text, I just get the geometry and not the text. Do you deem this relevant right now as diplan won't have the text-draw-feature activated?

Also, this PR is blocked by #264, right?

🏓 @warm-coolguy

packages/clients/diplan/src/store/geoEditing/duplicatePolygons.ts
warm-coolguy: https://github.com/Dataport/polar/pull/265/commits/841f27091fa9d478a7b4891bbf2371c05a33d7be

packages/clients/diplan/src/store/geoEditing/duplicatePolygons.ts
warm-coolguy: The code to do so would be rather ugly. `drawSource` is a "constant getter" outside of the vuex loop that never changes and is available from the start-up since its value is available from the start-up. The `drawLayer` is created further down the line during initialization and can't be trivially modeled as such a "constant getter" since it's "undefined" at the start and would permanently return "undefined" when implemented in the same fashion. (I actually tried beforehand in hopes of getting it done analoguously.)

## Solutions

1. Add drawLayer to a transport object and make the getter resolve `transportObject.drawLayer`, then the getter should return the correct element at run-time. *This may work*, but I'm not sure if vuex may do it once initially before the value is needed and thus also return `undefined` permanently.
2. Add drawLayer to store with the counter hack since we need to avoid putting ol objects in the store by themselves.
3. Just use the code above.

Depending on how this implementation fares, it may eventually be moved into the Draw plugin, where actions receive the drawLayer on creation anyway. For now, I consider 3 to be the cleanest solution, since both 1 and 2 introduce possibly unnecessary structures.

warm-coolguy:
> If I duplicate Text, I just get the geometry and not the text. Do you deem this relevant right now as diplan won't have the text-draw-feature activated?
> 
> Also, this PR is blocked by #264, right?
> 
> 🏓 @warm-coolguy

This PR doesn't seem to be blocked by that PR. It's a samey solution, but doesn't rely on it. Maybe I missed something or messed up the commit history? But I can't seem to find the spot.

🏓 @dopenguin

dopenguin (approved):
🚀 @warm-coolguy

PR #266, opened 2025-03-12
dish client: Feature/add separate layer for selected monument

jedi-of-the-sea:
## Summary

If there is a selected object in internal mode (= objectId as parameter), the map zooms to the monument. A new layer is created that shows only the selected object in yellow and in so creates a highlighting above the wms monument layer. The selected object can be set to visible / invisible through a new addon, a button with an eye icon, that will be added to the icon menu if an objectId is detected.

- start the dish client in mode `INTERN`
- add `?ObjektID=12` to the url

For this it is necessary to fetch the relevant wfs, so we might need to have a look how to get to those services with other machines. 
Also, I was not able to test it in Firefox because the CORS addon is not working there. I emulated safari and firefox browser with playwright, though (https://playwright.dev/). I only could simulate mobile devices with browser, again.

- [x] Functionality has been tested in Firefox (emulated), Chrome, Safari (emulated)
- [x] Functionality has been tested on a smartphone (emulated)

  - [x] Chrome Lighthouse
  - [x] Firefox Accessibility


jedi-of-the-sea:
I know that parameters from outside should be passed differently to POLAR, but in this case this is the best way due to the maintainer of the main application so I did not change that.

packages/plugins/Pins/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## unpublished
+
+- Feature: `getPointCoordinates` as been exposed as root export.
```
warm-coolguy, line 5:
```suggestion
- Feature: `getPointCoordinates` has been exposed as root export.
```

packages/clients/dish/src/plugins/SelectionObject/store.ts
```diff
@@ -0,0 +1,29 @@
+import { PolarModule } from '@polar/lib-custom-types'
+import {
+  generateSimpleGetters,
+  generateSimpleMutations,
+} from '@repositoryname/vuex-generators'
+import { SelectionObjectState } from '../../types'
+
+const getInitialState = (): SelectionObjectState => ({
+  objectId: 0,
```
warm-coolguy, line 9:
```suggestion
  objectId: '',
```
See next comment.

packages/clients/dish/src/types.ts
```diff
@@ -124,3 +126,12 @@ export interface DishMapConfig
     urlParams: DishUrlParams
   }
 }
+
+export interface SelectionObjectState {
+  objectId: number
```
warm-coolguy, line 131:
```suggestion
  objectId: string
```
While a string containing a number, this still seems to always be a string – the `ObjektID` is taken from the parameters, but never parsed to become an actual number. It still works, but may confuse someone in future iterations.

warm-coolguy (changes requested):
🏓 @jedi-of-the-sea

jedi-of-the-sea:
🏓 @warm-coolguy

warm-coolguy (approved):
:shipit: @jedi-of-the-sea

PR #267, opened 2025-03-12
Feature/link button

dopenguin:
## Summary

Add the possibility to configure a new button that links to a configured site in the same tab.

`npm run diplan:dev` and click the new button in the bottom-left corner

- [ ] Screenreader functionality has been manually tested with NVDA <- @warm-coolguy please test this for me :*

  - [x] Chrome Lighthouse
  - [x] Firefox Accessibility

Blocked by #254


packages/clients/diplan/src/config.js
```diff
@@ -12,6 +12,10 @@ export default {
     248651.73157077, 5227198.20287631, 928366.12236557, 6118661.62507136,
   ],
   diplan: {
+    link: {
+      href: 'https://example.com',
+      icon: '$vuetify.icons.fullscreen-exit',
```
warm-coolguy, line 17: Maybe we shouldn't use the `-exit` variant in the example, it looks wrong starting from a small client.

packages/clients/diplan/src/plugins/LinkButton/LinkButton.vue
```diff
@@ -0,0 +1,53 @@
...
+        small
+        width="40"
+        height="40"
+        v-bind="attrs"
+        v-on="on"
+      >
+        <v-icon color="primary">{{ linkConfig.icon }}</v-icon>
+      </v-btn>
+    </template>
+    <span>{{ $t('TODO') }}</span>
```
warm-coolguy, line 17: This has been forgotten. I suppose the solution is adding a configuration parameter that adds some text here. For the configuration example, it should be a locale key.

warm-coolguy (changes requested):
🏓 @dopenguin 

NVDA does not read the "TODO" tooltip. I suppose it has to be added as `aria-label`, too.

Proposal: Check the solution in the screen reader integrated in macOS. While it probably differs from what NVDA does, it should suffice for such basic cases as buttons.

packages/clients/diplan/src/config.js
warm-coolguy: Nvm, I viewed it in dev mode, where it looks wrong. It's fine for the builds, which is what matters. 🎩

packages/clients/diplan/src/plugins/LinkButton/LinkButton.vue
dopenguin: Added the label 1162f55ca80f6ece8f5845a132bf39a4a489058b

dopenguin:
> 🏓 @dopenguin
> 
> NVDA does not read the "TODO" tooltip. I suppose it has to be added as `aria-label`, too.
> 
> Proposal: Check the solution in the screen reader integrated in macOS. While it probably differs from what NVDA does, it should suffice for such basic cases as buttons.

Done d06225b87ade660131702b7969b9a6679c15d9ea
Works with the default ScreenReader (VoiceOver) as well and seems to be fine for tests.

🏓 @warm-coolguy

warm-coolguy (approved):
:shipit: @dopenguin

PR #269, opened 2025-03-13
DISH Client: pass newTab from URL parameter

jedi-of-the-sea:
## Summary

It has come to my attention that the parameter `NewTab`, that is used in the URL for DishExportMap (and that has been set to `true` because it always was `true` during my tests with the previous application) controls if there is a link for the user to return to the previous page of the application or if that link is missing. This is why `NewTab` can now be passed to the map via URL parameter (this way of passing the parameter is predefined by the application using the map). 

- Start the dish client in mode `INTERN`
- - add `?NewTab=false` to the URL
- Select a monument by clicking on it
- Click on "Map Export PDF"
- Check in the new tab for the link "Zurück zur Auswahlseite"
- fyi: the map in the pdf print will be missing except the basemap is selected as background.

  - [x] Chrome Lighthouse
  - [x] Firefox Accessibility


warm-coolguy (changes requested):
🏓 @jedi-of-the-sea 

In the step `Check in the new tab for the link "Zurück zur Auswahlseite"`, no such link is available. Maybe this is since I'm not logged in there?

warm-coolguy (approved):
:shipit: @jedi-of-the-sea

PR #270, opened 2025-03-14
DISH Client: Fix/image protrudes from narrow gfi

jedi-of-the-sea:
## Summary

The image of the selected monument in the external dish map was protruding if the gfi became too narrow due to small client width.

- start dish client with mode `EXTERN`
- click an a monument in the map to open a gfi
- make the browser smaller (narrower) so that the gfi adjusts accordingly
- the image should be hidden before it gets too big for the gfi window

  - [x] Chrome Lighthouse
  - [x] Firefox Accessibility


warm-coolguy (changes requested):
🏓 @jedi-of-the-sea 

Comparison images: Left is dev mode, right is currently productive page.

The image of the "Opernhaus" becomes painfully small:

pain

The image of "Schrevenpark: Rosengarten" is missing completely, no matter how narrow/wide I make the browser window:

miss

I'm a little out of the loop, but while I'm here:

* Is it intended that "Willkommen in der Denkmalkarte Schleswig-Holstein" is no longer displayed initially?
* Is it intended that images on max width no longer "touch" the sides of their card?
* Is it intended that the draw tool is available in `EXTERN` mode?
* Is it intended that the "Denkmalkarte Schleswig-Holstein" banner is missing now? If so, the spacing above the AddressSearch may be reduced, it was artificially inflated to make room for that banner.
* The 1:X bar's contents not being centered becomes very apparent after zooming in a lot. It looks a little off. Is it intended?
* Is it intended that the PDF export is now outside of the GFI? Makes it difficult to realize what the button is actually doing.

Maybe the answer to all of those questions is simply "Yes", but at least I'm back in the loop then. 😄

warm-coolguy:
🏓 @jedi-of-the-sea Oh, I see. I changed the flag, but I changed it in the README.md's example ... 🤣 

Is it a hard requirement to hide the image? I wondered whether it'll appear at all like this on mobile devices (currently can't test since the GFI window is missing there) and users with small windows might miss the images altogether.

It seems to also suffice to change the CSS declaration of `min-width` to `max(350px, 100%);` so that the image will shrink further, including to the breakpoint where it's supposed to switch to the MoveHandle UI element.

jedi-of-the-sea:
🏓 @warm-coolguy no, it was no hard requirement. I edited the code accordingly.

warm-coolguy:
Unbenannt

🏓 @jedi-of-the-sea I figured out why the solution with less pixels wasn't chosen ...

warm-coolguy (approved):
:shipit: @jedi-of-the-sea

PR #273, opened 2025-03-17
remove peer dependencies from diplan

warm-coolguy:
## Summary

Modeled DiPlan client without the peer dependencies since the leading system does not use `--legacy-peer-deps` and versions brought in with this installation would possibly stay in conflict with other installations there. Since POLAR comes pre-bundled, it's pointless for them to know anyway. The packages are not peer from their point of view.

Not much. Might check whether it still works after reinstall.


dopenguin (approved):
:shipit:

PR #274, opened 2025-03-17
unkink and simplify coordinates in diplan client export

warm-coolguy:
## Summary

The DiPlan-specific exports are now further unkinked and cleaned in an effort to produce an export valid in the eyes of the Simple Geometry specification (without substantial changes to the geometries drawn).

One TODO has been left in the right spot since that's more of a refactor without additional gains.

Check dev mode for whether exports still work. Fiddle around with various inputs to verify.


dopenguin (approved):
:shipit: @warm-coolguy

PR #277, opened 2025-03-18
change build setup & workaround placement

warm-coolguy:
## Summary

- Change placement of diplan .scss mod from core to diplan-local precompile step
- Change build setup back again so that the dev mode may reside in the example folder
  - With this, the src folder only contains source files again
  - With this, the style.css is deduplicated
- Built main file's name now matches the `main` entry in `package.json` again

## Caveat

While the build setup works in the same fashion as before, it doesn't seem to function as expected on my machine:

- When bundling, the resulting file is not importable (neither `import X from './dist/polar-client.js` nor `import * as X from './dist/polar-client.js` seem to work)
- However, the embedding programmer is already using it like that

Proposal: Please check whether this build works locally for you. I've found some hints that this may result from me being on Windows; which would explain why it works for the embedder, the final bundle happens on the GitHub Pipeline in Ubuntu, after all.

If it works locally for you, great success! Elsewise, we'll probably have to test this either in the pipeline directly (ew), or find yet another approach. Due to the bundle failing, I could not check whether the build fully works.

I've set us both as assignee since we'll probably both have to work on this.


dopenguin:
Relevant implementaion added to #254

PR #278, opened 2025-03-19
disable automatic full interaction end on first end

warm-coolguy:
## Summary

The methods cutPolygon/mergePolygon/duplicatePolygon are no longer turned off after the first `drawend` of the Draw method. Instead, they now work until intentionally disabled.

The `/polar-ui` example has been extended to showcase these functions and allows turning them off manually.

Check `/polar-ui` example in build mode.

### DIFF
--- a/packages/clients/diplan/example/polar-ui/index.html
+++ b/packages/clients/diplan/example/polar-ui/index.html
@@ -17,17 +17,27 @@
         <h2 class="padded-x">Beispielinstanz</h2>
         <div class="padded-x">
           <h3>Steuerung von außen</h3>
-          <button id="action-plus">+</button>
-          <button id="action-minus">-</button>
-          <button id="action-toast">Toast</button>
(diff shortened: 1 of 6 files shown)


dopenguin (changes requested):
Please also remove the `dispatch('setMode', 'none')` from the default export in `packages/plugins/Draw/src/store/createInteractions/createLassoInteractions.ts` as well

🏓 @warm-coolguy

warm-coolguy:
> Please also remove the `dispatch('setMode', 'none')` from the default export in `packages/plugins/Draw/src/store/createInteractions/createLassoInteractions.ts` as well
> 
> 🏓 @warm-coolguy

🏓 @dopenguin 

Oh right, thanks. Btw I skipped the changelog since the auto-disable feature wasn't documented anyway and there was no release yet.

dopenguin (approved):
:shipit: @warm-coolguy

PR #279, opened 2025-03-19
Feature/diplan bkg

dopenguin:
## Summary

Add BKG example to `@polar/client-diplan`.

- `npm run diplan:dev`
- Searching anything should now yield results put into separate categories


warm-coolguy (approved):
:shipit:

PR #281, opened 2025-03-20
fix minZoom for layers to work from minZoom on

warm-coolguy:
## Summary

`minZoom` was wrong by one zoom level due to a change in parameter forwarding in the `@masterportal/masterportalapi`. This has been resolved by mapping the value for the masterportalapi appropriately. No changes regarding the usage in POLAR have been made, the parameter is still meant inclusively on our end. This is to avoid the breaking changes in all configurations using it.

Run e.g. `diplan:dev` to see the change in effect.


dopenguin (approved):
:shipit: @warm-coolguy

PR #282, opened 2025-03-20
prevent snapping to invisible layers

warm-coolguy:
## Summary

The `snapTo` configuration has been changed to stop snapping to the layers while they're invisible.

E.g. `diplan:dev` can be used to verify that interactions no longer snap to parcels while they're invisible.


dopenguin (approved):
:shipit: @warm-coolguy

PR #283, opened 2025-03-20
repair handling of duplicate polygon points

warm-coolguy:
## Summary

It was previously possible to produce errors by manually modifying a Polygon so that multiple of its points were on the same coordinate.

1. Draw coordinate with 4 vertices
2. Move third vertex onto second
3. Look at error in console; diplan export stopped updating

To fix this, the order of operations has been adjusted. Coordinates are now cleaned prior to further handling by methods that expect clean coordinates.

Verify functionality in `diplan:dev`.

### DIFF
--- a/packages/clients/diplan/src/store/updateState/index.ts
+++ b/packages/clients/diplan/src/store/updateState/index.ts
@@ -48,8 +48,9 @@ export const updateState = async ({
       (accumulator, feature) => {
         if (['Polygon', 'MultiPolygon'].includes(feature.geometry.type)) {
           accumulator.push(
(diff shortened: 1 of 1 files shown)


warm-coolguy:
🏓 @dopenguin After last week's manual review, I added some minimalistic edge case handling for the found error: https://github.com/Dataport/polar/pull/283/commits/9fc18179a0d9efe798be90333f6e75ebbe30fc45

In the long run, we should probably offer some sort of error indicator that the outlying process can make use of to diplay an error. Dispatching toasts within the map on each catch quickly filled up the screen due to Draw firing updates a lot. For now, the validity flag toggles to false, and that should suffice for the first implementation.

dopenguin (approved):
:shipit: @warm-coolguy

PR #284, opened 2025-03-20
fix duplicate polygon style

warm-coolguy:
## Summary

Previously, duplicated geometries would sometimes have the OL default select style in build mode that did not appear in dev mode. This has been resolved and now works as expected in both modes.

Validate in build mode of diplan client.


dopenguin (approved):
:shipit: @warm-coolguy

PR #285, opened 2025-03-21
Fix issue of addInterceptor not having access yet to the regex

dopenguin:
## Summary

The store was not completely initialized so the regular expression was not yet added to the store. This PR fixes that by adding it as a parameter to the aciton.

`npm run diplan:dev` & use the patch I provided you. The layer should load again


warm-coolguy (approved):
:shipit:

PR #288, opened 2025-03-25
activate all e2e tests, active mobile e2e

warm-coolguy:
## Summary

- Mobile tests are now active
- An inactive test has been activated

Run E2E tests on your device. @dopenguin I'm curious whether these work on Safari! They're active, but not executed on my machine, and neither will they run in the pipeline ...

## Merge order

Both https://github.com/Dataport/polar/pull/288 and https://github.com/Dataport/polar/pull/291 can be merged first; the second one will have to be adjusted.


warm-coolguy:
@dopenguin Looks like `Mobile Safari` actually *is* running in the pipeline, to my surprise. Neat. However, the tests don't work there. Can you look into this and see what actually happens or fails to happen? Difficult to fix this from my machine. Low priority anyway.

dopenguin:
> @dopenguin Looks like `Mobile Safari` actually _is_ running in the pipeline, to my surprise. Neat. However, the tests don't work there. Can you look into this and see what actually happens or fails to happen? Difficult to fix this from my machine. Low priority anyway.

I'm not sure on how to properly check what is going wrong but I received the same results as the pipeline: `drawing.features` has a length of `0` so not feature is seemingly drawn

warm-coolguy:
> > @dopenguin Looks like `Mobile Safari` actually _is_ running in the pipeline, to my surprise. Neat. However, the tests don't work there. Can you look into this and see what actually happens or fails to happen? Difficult to fix this from my machine. Low priority anyway.
> 
> I'm not sure on how to properly check what is going wrong but I received the same results as the pipeline: `drawing.features` has a length of `0` so not feature is seemingly drawn

Playwright has a visual mode where you can watch it work, `npx playwright test --ui` may work (I just have some buttons next to the test in VSC, so I don't even know what's executed atm). If it's too fast, spraying in some `await new Promise((resolve) => setTimeout(resolve, 1000))` will help.

For some reason, `Safari Mobile` doesn't seem to draw. Is the canvas on the screen? Do the mouse coordinates look correct? Maybe the bounding box calculation is different on `Safari Mobile ` ... mystical, but it's good you can make it fail locally.

warm-coolguy:
@dopenguin Updated, but can't see a difference regarding `Mobile Safari`. Can you see anything on your hardware when running the e2e tests? If not, I'll just disable Safari for now. Better to have at least the mobile chromium test than none.

dopenguin:
> @dopenguin Updated, but can't see a difference regarding `Mobile Safari`. Can you see anything on your hardware when running the e2e tests? If not, I'll just disable Safari for now. Better to have at least the mobile chromium test than none.

When checking the test case `two features drawn at the same coordinate can be modified separately` the used `y` coordinate is off the canvas, more than double what it is in `the draw tool can be used to draw polygons on the map`

Click in `two features drawn at the same coordinate can be modified separately`:
Bildschirmfoto 2025-03-31 um 15 05 16
Click in `the draw tool can be used to draw polygons on the map`:
Bildschirmfoto 2025-03-31 um 15 04 04

@warm-coolguy

dopenguin (approved):
:shipit: @warm-coolguy

PR #291, opened 2025-03-26
add nested e2e test setup

warm-coolguy:
## Summary

Setup idea for nested e2e tests. I ~~'ve refrained from~~ add~~ing~~ it to the pipeline ~~since it doesn't currently do anything~~, ~~but~~ it ~~should~~ work there as well. This PR is ~~merely~~ about whether that setup is agreeable and works on your device. ~~Actual client e2e tests would have to follow.~~ There's ~~one~~ three ~~minimalistic~~ well-written example test.

Edit: Looks like it's already broken in the pipeline so far. Reverting to draft state.
Edit2: Fixed.

Run `test:e2e:diplan`.

## Merge order

Both https://github.com/Dataport/polar/pull/288 and https://github.com/Dataport/polar/pull/291 can be merged first; the second one will have to be adjusted.


packages/clients/diplan/e2e/utils/openDiplan.ts
```diff
@@ -0,0 +1,8 @@
+import { Page } from '@playwright/test'
+
+export const openDiPlan = async (page: Page) => {
+  // @ts-expect-error | it's manually added in snowbox client
```
dopenguin, line 4:
```suggestion
  // @ts-expect-error | it's manually added in diplan client
```

dopenguin (approved):
Just a small thing, looks great otherwise!

:shipit: @warm-coolguy

PR #295, opened 2025-03-28
Feature/add more layers to diplan

dopenguin:
## Summary

- Add additional layers to the diplan examples. This includes using the live-version of the xplan-layers instead of the qs one.
- Also, highlight different attributions; I took a look a the capabilities for the information added here.
- Remove the minZoom of Flurstücke for now as the UI is still bugged

Check both dev and prod mode of the diplan client.


warm-coolguy (approved):
:shipit:

PR #296, opened 2025-03-31
fix diplan export example

warm-coolguy:
## Summary

The examples were inconsistent/dysfunct regarding the screenshot button. This has been smoothened. While at it, placeholder rows have been removed as per Curly's Extended Law. ("Do two things at most.")

See dev mode and diplan-ui-one examples. In all others, the feature has been hidden.


dopenguin (approved):
:shipit: @warm-coolguy

PR #297, opened 2025-04-01
refactor store to be initialized initialler

warm-coolguy:
## Summary

The initial initialization has become even more initial! Now with setting up data during data setup.

Please review whether I missed a thing.

* `mapConfiguration`, `oidcToken`, and `secureServiceUrlRegex` are now treated at time of store initialization.
* All other initialization has been postponed until after the map actually exists, e.g. due to requiring the Toast addon in error case.


dopenguin (approved):
Works like a charm

:shipit: @warm-coolguy

PR #299, opened 2025-04-01
Dish client: Feature/adjust scale values

jedi-of-the-sea:
## Summary

The scale description in the layer chooser for the alkis layer is now calculated to show the right scale depending on different devices.

- set the mode to `EXTERN`
- start the dish client
- check the scale value in the layer chooser for alkis layer
- you can change the returned dpi in the function `getDpi` of the scale plugin to get different results.

  - [ ] Chrome Lighthouse
  - [ ] Firefox Accessibility

### DIFF
--- a/packages/clients/dish/src/utils/calculateScaleFromResolution.ts
+++ b/packages/clients/dish/src/utils/calculateScaleFromResolution.ts
@@ -1,0 +1,13 @@
+import { beautifyScale, getDpi, thousandsSeparator } from '@polar/plugin-scale'
+import * as olProj from 'ol/proj'
+
+export function calculateScaleFromResolution(
+  unit: string,
+  resolution: number
+): string {
+  const inchesPerMetre = 39.37
+  const scale = Math.round(
+    resolution * olProj.METERS_PER_UNIT[unit] * inchesPerMetre * getDpi()
+  )
+  return thousandsSeparator(beautifyScale(scale))
+}
--- a/packages/clients/dish/src/utils/calculateScaleFromResolution.ts
+++ b/packages/clients/dish/src/utils/calculateScaleFromResolution.ts
@@ -1,6 +1,25 @@
 import { beautifyScale, getDpi, thousandsSeparator } from '@polar/plugin-scale'
 import * as olProj from 'ol/proj'
-import { options } from '../mapConfigurations/mapConfig'
+
+let zoomLevel = 0
+
+export const options = [
+  { resolution: 264.583190458, scale: 1000000, zoomLevel: zoomLevel++ },
+  { resolution: 132.291595229, scale: 500000, zoomLevel: zoomLevel++ },
+  { resolution: 66.14579761460263, scale: 250000, zoomLevel: zoomLevel++ },
+  { resolution: 26.458319045841044, scale: 100000, zoomLevel: zoomLevel++ },
+  { resolution: 15.874991427504629, scale: 60000, zoomLevel: zoomLevel++ },
+  { resolution: 10.583327618336419, scale: 40000, zoomLevel: zoomLevel++ },
+  { resolution: 5.2916638091682096, scale: 20000, zoomLevel: zoomLevel++ },
+  { resolution: 2.6458319045841048, scale: 10000, zoomLevel: zoomLevel++ },
+  { resolution: 1.3229159522920524, scale: 5000, zoomLevel: zoomLevel++ },
+  { resolution: 0.6614579761460262, scale: 2500, zoomLevel: zoomLevel++ },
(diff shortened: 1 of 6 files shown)


packages/clients/dish/src/utils/calculateScaleFromResolution.ts
```diff
@@ -0,0 +1,36 @@
...
+  { resolution: 1.3229159522920524, scale: 5000, zoomLevel: zoomLevel++ },
+  { resolution: 0.6614579761460262, scale: 2500, zoomLevel: zoomLevel++ },
+  { resolution: 0.2645831904584105, scale: 1000, zoomLevel: zoomLevel++ },
+  { resolution: 0.1322915952292052, scale: 500, zoomLevel: zoomLevel++ },
+  { resolution: 0.06614579761, scale: 250, zoomLevel: zoomLevel++ },
+  { resolution: 0.02645831904, scale: 100, zoomLevel: zoomLevel++ },
+  { resolution: 0.01322915952, scale: 50, zoomLevel: zoomLevel++ },
+]
+
+function calculateScaleFromResolution(
```
dopenguin, line 24: Please add a comment to this function on why inchesPerMetre is needed.

After some searching I found that the dpi can be easily measured in inches, as 1 inch converts to 96 px (see [here](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Styling_basics/Values_and_units#lengths)). This surely would be helpful information to stumble across in the future.

packages/clients/dish/src/utils/calculateScaleFromResolution.ts
```diff
@@ -0,0 +1,36 @@
...
+  unit: string,
+  resolution: number
+): string {
+  const inchesPerMetre = 39.37
```
dopenguin, line 28: Is it interesting, that this is rounded?

packages/clients/dish/src/utils/calculateScaleFromResolution.ts
dopenguin, line 24: In my opionion, this is the way the scale should be calculated in the plugin as well. It is the approach that OpenLayers does anyway and only `@masterportal/masterportalapi` deviated from that.

@warm-coolguy agree?

packages/clients/dish/src/utils/calculateScaleFromResolution.ts
jedi-of-the-sea: I don't know - I use it because it is used in the scale plugin and I want the same result for the scale.

packages/clients/dish/src/utils/calculateScaleFromResolution.ts
```diff
@@ -0,0 +1,36 @@
...
+  { resolution: 0.6614579761460262, scale: 2500, zoomLevel: zoomLevel++ },
+  { resolution: 0.2645831904584105, scale: 1000, zoomLevel: zoomLevel++ },
+  { resolution: 0.1322915952292052, scale: 500, zoomLevel: zoomLevel++ },
+  { resolution: 0.06614579761, scale: 250, zoomLevel: zoomLevel++ },
+  { resolution: 0.02645831904, scale: 100, zoomLevel: zoomLevel++ },
+  { resolution: 0.01322915952, scale: 50, zoomLevel: zoomLevel++ },
+]
+
+function calculateScaleFromResolution(
+  unit: string,
```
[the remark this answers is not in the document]
jedi-of-the-sea: 0c9b2bc500d114fad849c56d3f03bc2fc0e56bc4

packages/clients/dish/src/utils/calculateScaleFromResolution.ts
```diff
@@ -0,0 +1,36 @@
...
+  resolution: number
+): string {
+  const inchesPerMetre = 39.37
+  const scale = Math.round(
+    resolution * olProj.METERS_PER_UNIT[unit] * inchesPerMetre * getDpi()
```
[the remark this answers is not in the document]
jedi-of-the-sea: 01b27669245619783f2db95806386621819360d5

jedi-of-the-sea:
🏓 @dopenguin

packages/clients/dish/src/utils/calculateScaleFromResolution.ts
warm-coolguy: Agreed.

warm-coolguy: Disagreed. I misunderstood the situation and thought that this was about the resolution in the map. It is insufficient to calculate all scales from resolutions of the masterportalAPI's `options` parameter since that only provides "full" zoom levels. However, in OL, we can get between zoom levels by going wild on the mouse wheel, which will produce wild resolutions anywhere in between the zoom levels we specify.

PR #300, opened 2025-04-01
Fix/disabled masks

dopenguin:
## Summary

Wrong layers were disabled when using different types for overlays than `mask` as the getter `disabledMasks` was working on the index.

`npm run diplan:dev`

- ~~[ ] Changelogs are maintained~~ Feature has not been published yet

  - [x] Chrome Lighthouse
  - [x] Firefox Accessibility

### DIFF
--- a/packages/clients/diplan/src/plugins/LayerChooser/LayerChooser.vue
+++ b/packages/clients/diplan/src/plugins/LayerChooser/LayerChooser.vue
@@ -11,7 +11,7 @@
             <template v-for="({ name, id }, index) in backgrounds">
               <LayerWrapper
                 :key="'disabled-background-' + index"
-                :index="index"
+                :index="Number(index)"
                 :disabled-layers="disabledBackgrounds"
                 :layer-id="id"
               >
@@ -56,7 +56,7 @@
                     dense
                     hide-details
                     class="cut-off-top-space"
-                    :disabled="disabledMasks[index]"
+                    :disabled="disabledMasks[id]"
                   />
                 </LayerWrapper>
                 <v-divider :key="index" />
--- a/packages/clients/diplan/src/plugins/LayerChooser/LayerChooser.vue
+++ b/packages/clients/diplan/src/plugins/LayerChooser/LayerChooser.vue
@@ -8,24 +8,22 @@
         <v-divider />
         <v-expansion-panel-content>
           <v-radio-group v-model="activeBackground" dense hide-details>
-            <template v-for="({ name, id }, index) in backgrounds">
+            <template v-for="{ name, id } in backgrounds">
               <LayerWrapper
-                :key="'disabled-background-' + index"
-                :index="index"
+                :key="`background-layer-${id}`"
                 :disabled-layers="disabledBackgrounds"
                 :layer-id="id"
               >
                 <v-radio
-                  :key="index"
                   aria-describedby="polar-label-background-title"
                   dense
                   hide-details
                   :label="$t(name)"
                   :value="id"
-                  :disabled="disabledBackgrounds[index]"
+                  :disabled="disabledBackgrounds[id]"
                 />
               </LayerWrapper>
-              <v-divider :key="index" />
+              <v-divider :key="`background-divider-${id}`" />
             </template>
           </v-radio-group>
         </v-expansion-panel-content>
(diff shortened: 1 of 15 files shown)


packages/clients/diplan/src/plugins/LayerChooser/LayerChooser.vue
```diff
@@ -11,7 +11,7 @@
             <template v-for="({ name, id }, index) in backgrounds">
               <LayerWrapper
                 :key="'disabled-background-' + index"
-                :index="index"
+                :index="Number(index)"
```
warm-coolguy, line 14: Index already is a number. [Example](https://play.vuejs.org/#eNp9kU1PxCAQhv8KmdOa1O1BT5tqomYPelCj3sSYpp1WVgoEhlrT9L8L1P04bJYDMO+8wzwTRrgxZtl7hBUUhJ2RJeE1V5wKL8PJwiqkYP15o+0Vh8VnxoSqcTgLB3tXXsqM7fcPDqmYE2PjyOjXoG4W/wXTxFZRTWGIZmORS5EaFXnqWOQHGJABuUqrRrTLjdMqUI7Ry6HSnRES7ZMhoZXjEJ6ecTmUUuqfh6SR9Zht9eoLq+8j+sYNUePwbNGh7ZHDLkelbZHm9Pr1EYdw3yU7XXsZ3CeSL+i09JFxtt16VQfsA1+ive+MtiRU++bWA6Fy26EiaHROyc8hfNXdidH3uBfLy1TH1QTTH7CSn10=)

packages/plugins/LayerChooser/src/components/LayerWrapper.vue
```diff
@@ -37,7 +34,7 @@ export default Vue.extend({
       required: true,
     },
     disabledLayers: {
-      type: Array,
+      type: [Array, Object],
```
warm-coolguy, line 37: Migrating all `disabledLayers` to `Object` would make the whole thing easier to digest in the future. Or was there a reason why it couldn't/shouldn't be done?

warm-coolguy (changes requested):
🏓 @dopenguin Just some minor stuff.

packages/clients/diplan/src/plugins/LayerChooser/LayerChooser.vue
dopenguin: As discussed, the type allows both Strings and Numbers, we can keep this even though in our case it always is a Number.

packages/plugins/LayerChooser/src/components/LayerWrapper.vue
dopenguin: I didn't migrate it for `disabledBackgrounds` for now as I didn't intend to change too much.

But it makes sense to do this now, while I am at it!

packages/clients/diplan/src/plugins/LayerChooser/LayerChooser.vue
warm-coolguy: If you want to handle strings, you're going to need more than `Number(index)`, since `Number("arbitrary")` is going to give you a lot of `NaN` that's not very usable as index.

dopenguin: Subsequently removed it from every part added through previous PRs e79d3be6f13a577e13def2063feace5853627c3e

packages/plugins/LayerChooser/src/components/LayerWrapper.vue
dopenguin: Adjusted d1c958d858bf91f96f9eda6e8d5502419d0b8698

packages/plugins/LayerChooser/src/utils/isLayerIdIncluded.ts
```diff
@@ -0,0 +1,4 @@
+import { LayerConfiguration } from '@polar/lib-custom-types'
+
+export const isLayerIdIncluded = (layers: LayerConfiguration[], id: string) =>
```
warm-coolguy, line 3: The predicate `is` implies a boolean return, but we get an index. Proposal: Move the `=== -1` comparison from the usages of `isLayerIdIncluded` to this spot.

warm-coolguy (changes requested):
🏓 @dopenguin

packages/plugins/LayerChooser/src/utils/isLayerIdIncluded.ts
dopenguin: Done 841b66af0255d763ec8bfff64619f6fb7598df2d

dopenguin:
> 🏓 @dopenguin

🏓 @warm-coolguy

warm-coolguy (approved):
:shipit: @dopenguin

PR #303, opened 2025-04-03
Fix/diplan layer options

dopenguin:
## Summary

Change the way the options are opened in `@polar/client-diplan`.
Note that only one options menu can be open at once as this is the implementation currently used in the plugin. I wouldn't touch this though, as the implementation of the plugin itself has not been touched.

`npm run diplan:dev`

- ~~[] Changelogs are maintained~~ - not sure if a changelog entry for LayerChooser is needed though; in diplan this is not needed as the feature for the different groups is not published yet
- [x] Screenreader functionality has been manually tested with NVDA - When navigating the options menu via keyboard, the whole application is focused after switching from / to the Options instead of an element in the menu. This is something I'll tackle in a separate PR

  - [x] Chrome Lighthouse
  - [x] Firefox Accessibility

### DIFF
--- a/packages/clients/diplan/src/plugins/LayerChooser/LayerChooser.vue
+++ b/packages/clients/diplan/src/plugins/LayerChooser/LayerChooser.vue
@@ -40,7 +40,8 @@
           </v-expansion-panel-header>
           <v-divider />
           <v-expansion-panel-content>
-            <template v-if="displaySelection">
+            <LayerChooserOptions v-if="displayOptionsForType[type]" />
+            <template v-else>
               <template v-for="({ name, id }, index) in masks">
                 <LayerWrapper
                   :key="`disabled-mask-${type}-${index}`"
@@ -62,7 +63,6 @@
                 <v-divider :key="index" />
               </template>
             </template>
-            <LayerChooserOptions v-else />
           </v-expansion-panel-content>
         </v-expansion-panel>
       </template>
@@ -89,6 +89,7 @@ export default Vue.extend({
       'backgrounds',
       'disabledBackgrounds',
       'disabledMasks',
+      'displayOptionsForType',
       'masksSeparatedByType',
       'openedOptions',
       'shownMasks',
@@ -108,9 +109,6 @@ export default Vue.extend({
       set(value) {
         this.setActiveMaskIds(value)
       },
-    },
-    displaySelection() {
-      return this.openedOptions === null
     },
   },
   methods: {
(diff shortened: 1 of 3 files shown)


packages/plugins/LayerChooser/src/store/index.ts
```diff
@@ -211,17 +211,27 @@ export const makeStoreModule = () => {
...
       idsWithOptions(_, { backgrounds, masks }) {
         return [...backgrounds, ...masks]
           .filter((layer) => Boolean(layer.options))
           .map((layer) => layer.id)
       },
       masksSeparatedByType: (_, { shownMasks }) =>
-        shownMasks.reduce((acc, mask) => {
-          if (Object.keys(acc).includes(mask.type)) {
-            return { ...acc, [mask.type]: [...acc[mask.type], mask] }
+        shownMasks.reduce((acc, masks) => {
```
warm-coolguy, line 230: Misleading pluralization. `shownMasks` is a `LayerConfiguration[]`, hence mask(s) is a singular `LayerConfiguration`.

packages/clients/diplan/src/plugins/LayerChooser/LayerChooser.vue
```diff
@@ -89,6 +89,7 @@ export default Vue.extend({
       'backgrounds',
       'disabledBackgrounds',
       'disabledMasks',
+      'displayOptionsForType',
       'masksSeparatedByType',
       'openedOptions',
```
warm-coolguy, line 94:
```suggestion
```
Unused.

packages/clients/diplan/src/plugins/LayerChooser/LayerChooser.vue
```diff
@@ -40,7 +40,8 @@
           </v-expansion-panel-header>
           <v-divider />
           <v-expansion-panel-content>
-            <template v-if="displaySelection">
+            <LayerChooserOptions v-if="displayOptionsForType[type]" />
```
warm-coolguy, line 43: Currently only one options menu can be active at a time. This is a confusing user experience since the closed menu changes its state while the user is not looking.

confusion_of_the_highest_order

warm-coolguy (changes requested):
🏓 @dopenguin

packages/clients/diplan/src/plugins/LayerChooser/LayerChooser.vue
dopenguin: As mentioned in the PR description and discussed separately, this'll stay for this PR. If we further work on the client, this should be changed though.

packages/plugins/LayerChooser/src/store/index.ts
dopenguin: Woops, should've looked at the type d86f0247ee2e8918db824e5338d316d25e17e99b

dopenguin:
> 🏓 @dopenguin

🏓 @warm-coolguy

warm-coolguy (approved):
🚀 @dopenguin

PR #304, opened 2025-04-04
fix measurement calculation

warm-coolguy:
## Summary

Fix measurement calculation.

Visible in e.g. Snowbox. Außenalster has about a km width at its largest, but you may use other things of known size.

- [x] Functionality has been tested in Firefox, ~~Chrome, Safari~~ numbers work the same way in all browsers


dopenguin (approved):
Measurements fit the scale 👍 

:shipit: @warm-coolguy

PR #306, opened 2025-04-04
Fix/mobile gfi

dopenguin:
## Summary

After removing several computed values and props in #131 the mutation to set the gfi in the MoveHandle through the watcher was no longer triggering as the values of moveHandleProperties were always the same.

Now, the GFI window is displayed again on mobile devices when it's rendered independently and the closeIcon is now again fa-angles-right when featureList is configured.

~~Note that a plugin opened in the IconMenu still disappears when resizing to a smaller width. This is a separate issue I'll tackle in another PR.~~ That is now fixed in #307.

Check both `@polar/client-meldemichel` and `@polar/client-dish` if the window is rendered properly on small devices

  - [x] Chrome Lighthouse
  - [x] Firefox Accessibility

Bug was introduced in #131

### DIFF
--- a/packages/plugins/Gfi/CHANGELOG.md
+++ b/packages/plugins/Gfi/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## unpublished
+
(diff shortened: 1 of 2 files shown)


warm-coolguy (approved):
:shipit:

PR #312, opened 2025-04-09
add default values to config in all spots

warm-coolguy:
## Summary

A prior refactor broke the defaulting; not all targets also got the default values as a baseline, due to what e.g. the `epsg` value was missing in BKG address search as configured in the bgw client. Now it's made sure that the default values are used everywhere, also if we later add additional default values.

See that e.g. `npm run bgw:dev` has a working BKG address search again.

No changelog entry required since there was no release since the bug appeared and it in itself was merely a refactor.

Was broken in https://github.com/Dataport/polar/pull/297.


dopenguin (approved):
:shipit: @warm-coolguy

PR #311, opened 2025-04-08
remove diplan client

warm-coolguy:
## Summary

Remove DiPlan client.

Blocked by https://github.com/Dataport/polar/pull/310. Lockfile update required before merge.

PR #313, opened 2025-04-10
Addition of a login example to @polar/client-snowbox

dopenguin:
## Summary

Add an example on how to login and save the token to @polar/client-snowbox.
This should subsequently be updated including a service and a keycloak server with which one can showcase the usage of secured services. This should be added in a separate PR once we've got ahold of those things.

`npm run snowbox` and check that you can login with the previously provided credentials. Note that a CORS plugin is required.

Example based on removed example from #311


warm-coolguy (approved):
I don't have the test URLs here anymore, but I'll assume that you tested the copypasta.

🍝 @dopenguin

PR #314, opened 2025-04-10
Remove need to call initializeLayerList for local service registers

dopenguin:
## Summary

@masterportal/masterportalapi already handles this. It is important to note that it is an async function that can not be awaited when used with a URL. Thus, it becomes required to still call it a second time in such cases.

Run `bgw`, `dish` and `textLocator` and check whether they still run properly.
Note that dish may have some issues loading one feature layer in production mode, but that issue already exists on `main`.

  - [x] Chrome Lighthouse
  - [x] Firefox Accessibility

### DIFF
--- a/gettingStarted.md
+++ b/gettingStarted.md
@@ -239,8 +239,10 @@ const createMap = (layerConf) => {
     .then(addStoreSubscriptions(/* subscription logic */))
 }
 
-// Initializes the layer list and creates the map instance.
-// This becomes obsolete if a local service register is used.
+/* Initializes the layer list and creates the map instance.
+ * This becomes obsolete if a local service register is used.
+ * In that case, that service register can be directly used as `layerConf` on the `createMap` call.
+ */
 polarCore.rawLayerList.initializeLayerList(
   'https://geodienste.hamburg.de/services-internet.json',
   createMap
--- a/gettingStarted.md
+++ b/gettingStarted.md
@@ -239,7 +239,8 @@ const createMap = (layerConf) => {
     .then(addStoreSubscriptions(/* subscription logic */))
 }
 
-// Initializes the layer list and creates the map instance
+// Initializes the layer list and creates the map instance.
+// This is becomes obsolete once one uses a local service register.
 polarCore.rawLayerList.initializeLayerList(
   'https://geodienste.hamburg.de/services-internet.json',
   createMap
--- a/gettingStarted.md
+++ b/gettingStarted.md
@@ -240,7 +240,7 @@ const createMap = (layerConf) => {
 }
 
 // Initializes the layer list and creates the map instance.
-// This is becomes obsolete once one uses a local service register.
+// This becomes obsolete if a local service register is used.
 polarCore.rawLayerList.initializeLayerList(
   'https://geodienste.hamburg.de/services-internet.json',
   createMap
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -2,9 +2,9 @@
 
 ## unpublished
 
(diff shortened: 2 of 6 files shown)


gettingStarted.md
```diff
@@ -239,7 +239,8 @@ const createMap = (layerConf) => {
     .then(addStoreSubscriptions(/* subscription logic */))
 }
 
-// Initializes the layer list and creates the map instance
+// Initializes the layer list and creates the map instance.
+// This becomes obsolete if a local service register is used.
```
warm-coolguy, line 243:
```suggestion
/* Initializes the layer list and creates the map instance.
 * This becomes obsolete if a local service register is used. In that case, that service
 * register can be used as `layerConf` on the `createMap` above directly. */
```

As someone who's getting started, I may elsewise be confused about what to do next, since I don't know what exactly `initializeLayerList` is feeding as `layerConf` into the `createMap` function. Unluckily, the names diverge due to requirement by `@masterportal/masterportalapi`, and `layerConf` seems by name to imply more than `service register` somehow.

packages/core/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## unpublished
 
+- Feature: Remove need of calling `rawLayerList.initializeLayerList` when creating a client using a local service register.
 - Chore: Add teardown hints for single page application to README file.
```
warm-coolguy, line 6:
```suggestion
- Chore: Add documentation regarding optionality of calling `rawLayerList.initializeLayerList` when creating a client using a local service register.
- Chore: Add teardown hints for single page application to README file.
```
Since there's nothing implemented, I'd rather sort this into `Chore` since it's documentation.

packages/core/README.md
```diff
@@ -60,6 +60,9 @@ If the storeModule features a `setupModule` action, it will be executed automati
 Layers intended to be used in the map have to be initialized by calling `initializeLayerList` with a service register.  
 This register may either be a link to a predefined service register like [the Hamburg service register](https://geodienste.hamburg.de/services-internet.json), or the custom service register that is also used in [mapConfiguration.layerConf](#mapconfigurationlayerconf).
 
+This is an optional step as `@masterportal/masterportalapi` already handles this.
+However, it becomes required again to call this an additional time when using a URL for the service register as some setups need to be done first before anything polar-related can be set up.
```
warm-coolguy, line 64:
```suggestion
Layers registers to be fetched from remote files have to be initialized by calling `initializeLayerList` with their respective URL. This may e.g. point to a predefined service register like [the Hamburg service register](https://geodienste.hamburg.de/services-internet.json). This needs to be done before `createMap` is called, and `initializeLayerList`'s callback will get the retrieved `layerConf` as first parameter.

If you're using a local custom service register, i.e. you have the [mapConfiguration.layerConf](#mapconfigurationlayerconf) as an array at hand before calling `createMap`, it can be used as `layerConf` parameter directly without calling `initializeLayerList`. The `@masterportal/masterportalapi` then handles the layer list setup in time.
```

1. You have to do X
2. ... which is optional ...
3. ... UNLESS ...

I tried to write it more concisely.

warm-coolguy (changes requested):
🏓 @dop

packages/core/CHANGELOG.md
dopenguin: Moved e2fb4df195ac3527430128eee803add2fbc51411

packages/core/README.md
dopenguin: Adjusted dd0c7ec51faca2d6155be3c42861d02c37d8bbd7

warm-coolguy (approved):
@dopenguin

PR #315, opened 2025-04-10
Fix/layer options tabbing

dopenguin:
## Summary

When opening layer options via keyboard the focus was lost. Now, the back button or the options button are focused after opening / closing.

I encountered an issue when closing the options menu that is resolved by using `setTimeout`. I kinda dislike the solution as this should not be needed as we are already in the next tick so the DOM should've been rerendered. Thus, I am open for other solutions!

`npm run dish:dev` and open / close the options menu via keyboard. The respective button should be focused afterwards.

- ~~[] Functionality has been tested on a smartphone~~

  - [x] Chrome Lighthouse
  - [x] Firefox Accessibility


warm-coolguy (changes requested):
🐧 @dopenguin 

hmm

dopenguin:
> 🐧 @dopenguin
> 
> hmm 

Woops e1f59d658f37df592af827f676b2d3aba0f31919

🏓 @warm-coolguy

warm-coolguy (approved):
@dopenguin

PR #317, opened 2025-04-11
add region role to map div

warm-coolguy:
## Summary

A "region" role has been added to the render div.

* Chrome Lighthouse indicated that an `aria-label` on a `div` is insufficient.
* Supposedly JAWS doesn't read that `aria-label` without further ado: https://stackoverflow.com/questions/42681485/div-aria-label-not-being-read-by-jaws
* NVDA didn't care, we can't really tell if it helped except for the Lighthouse warning vanishing (maybe it matters in Safari VoiceOver?)

Run Chrome Lighthouse check.

  - [x] Chrome Lighthouse
  - [x] Firefox Accessibility


dopenguin (approved):
Doesn't really make a difference for VoiceOver, it just additionally reads "Region" now 🤷 

:shipit: @warm-coolguy

PR #324, opened 2025-04-16
Fix bug in dish export print scale

jedi-of-the-sea:
## Summary

The plugin DishExportMap had a bug where the scale in the print sometimes differed from the scale shown in the map. This was due to the scale switcher using the scales from the configuration parameter `options`. This bug has been fixed.

- Start the map client in mode `INTERN`
- Click on a random monument object in the map to activate the button for `DishExportMap`
- Choose "basemap" as background (because the other background layers are not responding in our setup)
- Click on DishExportMap (Button with PDF icon on the left)
- log into the Dish Software (text me if you need the credentials)
- the print won't work the directly after log-in, you need to repeat the printing by clicking the button again
- the result after the second printing should be a PDF with a map. Check if the scale is the same as it was in the map during the trigger of the print
- to simulate different dpi (which will result in different scales) change the function `getDpi` in the Scale Plugin utils folder in a way that it returns different values. Then repeat the decribed steps with the different dpi.

  - [x] Chrome Lighthouse
  - [x] Firefox Accessibility


packages/plugins/Scale/src/utils/getBestMatchingScale.ts
```diff
@@ -0,0 +1,17 @@
+import beautifyScale from './beautifyScale'
+
+function getBestMatchingScale(
+  scaleValue: number,
+  zoomOptions: { resolution: number; scale: number; zoomLevel: number }[]
```
dopenguin, line 5: Please use `PolarMapOptions[]` similar to the type of the related getter.

jedi-of-the-sea: 207dc4b153c544bd93face87e24fb6de10c58148

packages/plugins/Scale/CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## unpublished
 
-- Feature: Expose the functions `beautifyScale`, `thousandsSeparator` and `getDpi`.
+- Feature: Expose the functions `beautifyScale`, `thousandsSeparator`, `getDpi`, `getBestMatchingScale` and `calculateScaleFromResolution`.
```
dopenguin, line 5: `getBestMatchingScale` is no longer needed because of the changes to `zoomOptions`.
See the [attached patch](https://github.com/user-attachments/files/19979648/remove_getBestMatchingScale.patch) to remove the remaining remnants.

packages/plugins/Scale/src/utils/calculateScaleFromResolution.ts
```diff
@@ -0,0 +1,18 @@
+import { METERS_PER_UNIT, type MetersPerUnitLookup } from 'ol/proj/Units'
+import beautifyScale from './beautifyScale'
+import getDpi from './getDpi'
+
+function calculateScaleFromResolution(
```
dopenguin, line 5: Except the final `beautifyScale` this is a duplicate to a lot of `getScale`. I'd suggest removing `beautifyScale` from the return here to be able to use it in `getScale` as well.

packages/plugins/Scale/src/utils/calculateScaleFromResolution.ts
```diff
@@ -0,0 +1,18 @@
...
+  unit: keyof MetersPerUnitLookup,
```
dopenguin, line 6: Please just use `Units` from `'ol/proj/Units'` here as that's the returned type of `rootGetters.map.getView().getProjection().getUnits()` and it seems to be the thing we want here.

packages/plugins/Scale/src/store/index.ts
```diff
@@ -85,7 +87,18 @@ export const makeStoreModule = () => {
...
         }
-        return options
+        return options.map((option) => {
+          return {
+            ...option,
+            scale: calculateScaleFromResolution(
+              rootGetters.map
+                .getView()
+                .getProjection()
+                .getUnits() as keyof MetersPerUnitLookup,
```
dopenguin, line 97: The type assertion is not needed with the suggested change in `packages/plugins/Scale/src/utils/calculateScaleFromResolution.ts`

dopenguin (changes requested):
🏓 @jedi-of-the-sea

packages/plugins/Scale/CHANGELOG.md
jedi-of-the-sea: I thought so, too, until I realized that the real scale value might be not be exactly the value from `zoomOptions`, like in this example
image

However, I just realized that when I use `beautifyScale` it also seems to work, so `getBestMatchingScale` can really be deleted.

jedi-of-the-sea: 0b248959dd24502025da9a18f899980c0788f3c3

packages/plugins/Scale/src/utils/calculateScaleFromResolution.ts
jedi-of-the-sea: b7ecf75c7e24c2dba63ebaeeb8d39a400b39f88d

packages/plugins/Scale/src/utils/calculateScaleFromResolution.ts
jedi-of-the-sea: f48ba62246351ae17c17e7d674c0dfcf544ae76b

packages/plugins/Scale/src/store/index.ts
jedi-of-the-sea: f48ba62246351ae17c17e7d674c0dfcf544ae76b

jedi-of-the-sea:
🏓 @dopenguin

packages/plugins/Scale/src/utils/calculateScaleFromResolution.ts
oeninghe-dataport: This needs further changes as discussed via ☎️

jedi-of-the-sea:
🏓 @oeninghe-dataport

oeninghe-dataport (approved):
🚀 @jedi-of-the-sea

oeninghe-dataport:
Closed and reopened as #328 for technical reasons

PR #335, opened 2025-07-21
ADR: Customer-specific repositories

oeninghe-dataport:
## Proposal for an Architecture Decision Record
Split customer-specific clients to separate repositories

*see the commit for details*


dopenguin (approved):
Can be merged after #334
🚀 @oeninghe-dataport

PR #336, opened 2025-07-21
ADR Pug templating in Vue SFCs

oeninghe-dataport:
## Proposal for an Architecture Decision Record
Use Pug templating engine for Vue SFCs

*see commit for details*


oeninghe-dataport:
Declined by maintainer team

PR #359, opened 2025-09-15
fix: make snowbox run on windows

warm-coolguy:
## Summary

Make snowbox work on Windows devices.

`npm run snowbox` should now work on Mac/Linux/Windows with `cross-env` introduced to set variables.


dopenguin (approved):
🚀 @warm-coolguy

PR #362, opened 2025-09-18
refactor: this button needs less js

warm-coolguy:
## Summary

I am touch-a your spaghett.

I have altered the way PolarIconButton works to skip some JS that can be modeled with CSS and made sure the component can be styled with external classnames. The bonus `<div>` wrapper is due to the component oftentimes needing to be both `position`ed `absolute` (in the map context) and `relative` (as tooltip anchor).

If you like it, just merge it to your branch as if it was a suggestion. Feel free to modify this branch, too.


dopenguin (approved):
Crazy how you made Lasagna from Spaghetti! @warm-coolguy

PR #366, opened 2025-09-24
Vue3/migration plugin geo location

warm-coolguy:
## Summary

Migrating the GeoLocation plugin to Vue@3. While at it, we changed a lot of the behaviour to better meet the usual expectations for such tools.

## Instructions for local reproduction and review

Run and test in snowbox. Also take a walk with your mobile device when trying this.

## Pull Request Checklist (for Assignee)

- [ ] Changelogs are maintained
- [ ] Functionality has been tested in Firefox, Chrome, Safari
- [ ] Functionality has been tested on a smartphone
- [ ] Functionality has been tested with 200% screen zoom
- [ ] Screenreader functionality has been manually tested with NVDA

UI has been tested in the following tools regarding accessibility (only regarding functionality affected in this PR)
  - [ ] Chrome Lighthouse
  - [ ] Firefox Accessibility

PR #374, opened 2025-09-29
feat(plugins/pins): migrate plugin

### LINKED ISSUE #352: Migration of Pins plugin to POLAR@3

dopenguin:
## Summary

Migrate Pins plugin to `@polar/polar`.

- `npm run snowbox`
- Click anywhere in the map
- A pin should now be visible
- If you click outside the boundary of Hamburg or try to drag the pin outside of the boundary, a toast should be displayed and the pin should be in it's previous location

Closes #352

~~If #366 is merged before this PR, then the `PinsOptions` should be extended by `LayerBoundPluginOptions`. Otherwise, this change has to be added to #366.~~ Done

Blocked by #404

### DIFF
--- a/src/core/types/plugin.ts
+++ b/src/core/types/plugin.ts
@@ -35,7 +35,7 @@ export interface PluginOptions {
 	layoutTag?: keyof typeof NineLayoutTag
 }
 
-interface BoundaryOptions {
+export interface BoundaryOptions {
 	/**
 	 * ID of the vector layer to restrict requests to.
 	 * The layer must contain vectors. This is useful for restricted maps to avoid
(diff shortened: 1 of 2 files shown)


src/plugins/pins/store.ts
```diff
@@ -0,0 +1,257 @@
...
+			toZoomLevel: number
+		}
+	>(() =>
+		toMerged(
+			{ minZoomLevel: 0, movable: 'none', toZoomLevel: 0 },
+			coreStore.configuration.pins || {}
+		)
+	)
+	const latLon = computed(() =>
+		toLonLat(coordinate.value, coreStore.configuration.epsg)
```
oeninghe-dataport: I'm a little bit confused here: Is it lonLat or latLon?

If it is lonLat, why not name it like that?

oeninghe-dataport (approved):
🚀 @dopenguin

PR #382, opened 2025-10-15
add reverse geocoder to afm client

warm-coolguy:
## Summary

Adding the reverse geocoder plugin to the AfM client.

The plugin has been configured as part of the AfM's example so that the usual usage can be tested. However, in its first use case, the `AddressSearch` plugin will probably not be present. Thus, no defaulting configuration has been added, as it may not be desired.

No real implementation work done, hence the example test seems to suffice to me. Everything else should work by virtue of having working previously.

### DIFF
--- a/packages/clients/afm/CHANGELOG.md
+++ b/packages/clients/afm/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## unpublished
 
 - Feature: Add `@polar/plugin-reverse-geocoder` to the client. For details about this plugin, see [the plugin's documentation](https://dataport.github.io/polar/docs/afm/plugin-reverse-geocoder.html).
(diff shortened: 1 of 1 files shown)


packages/clients/afm/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## unpublished
+
+- Feature: Add `@polar/plugin-reverse-geocoder` to the client. For details about this plugin, see [the plugin's documentation](https://dataport.github.io/polar/docs/afm/plugin-reverse-geocoder.html).
+
```
dopenguin: A second entry should be added as the AddressSearch now is being displayed.

dopenguin (changes requested):
🏓 @warm-coolguy

packages/clients/afm/CHANGELOG.md
dopenguin: Still seems somewhat relevant. Chore seems fitting.

dopenguin (approved):
🚀 @warm-coolguy

PR #383, opened 2025-10-20
feat: add an inset box-shadow to the map if POLAR is rendered as part of another page

dopenguin:
## Summary

Add an inset shadow for the map and add a border-radius. This addition is based on proposition from the UI design.

`npm run snowbox`. An inset shadow should now be visible. The border-radius is now also updated.

### DIFF
--- a/examples/snowbox/index.html
+++ b/examples/snowbox/index.html
@@ -45,8 +45,8 @@
 			}
 
 			.snowbox {
-			  	width: 90vw;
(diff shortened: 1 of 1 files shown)


oeninghe-dataport (changes requested):
🏓 @dopenguin

dopenguin:
> 🏓 @dopenguin

🏓 @oeninghe-dataport

PR #398, opened 2025-11-03
feat: darkmode

dopenguin:
## Summary

Adds possibility to override color scheme to light or dark depending on embedding system in contrast to receiving the information from the system settings.

- `npm run snowbox`
- POLAR should now be in light mode. The color scheme can also be switched through a button.

#354

### DIFF
--- a/src/core/components/PolarContainer.ce.vue
+++ b/src/core/components/PolarContainer.ce.vue
@@ -58,6 +58,10 @@ mainStore.configuration = mapZoomOffset({
 	...props.mapConfiguration,
 })
 
+if (mainStore.configuration.colorScheme) {
+	mainStore.colorScheme = mainStore.configuration.colorScheme
+}
+
 if (mainStore.configuration.oidcToken) {
 	// copied to a separate spot for usage as it's changeable data at run-time
 	mainStore.oidcToken = mainStore.configuration.oidcToken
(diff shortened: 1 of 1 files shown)


src/core/components/PolarContainer.ce.vue
```diff
@@ -112,6 +121,13 @@ onMounted(() => {
 		mainStore.shadowRoot as ShadowRoot,
 		mainStore.configuration.theme?.kern || {}
 	)
+	if (mainStore.configuration.colorScheme) {
+		mainStore.colorScheme = mainStore.configuration.colorScheme
+	}
```
oeninghe-dataport: This should be moved to line 60 (below the very similar code for the oidcToken).

src/core/components/PolarContainer.ce.vue
```diff
@@ -112,6 +121,13 @@ onMounted(() => {
...
+	;(polarWrapper.value as HTMLDivElement).setAttribute(
+		'data-kern-theme',
+		mainStore.colorScheme
+	)
```
oeninghe-dataport: See above, this could be unnecessary when setting via template.

oeninghe-dataport (changes requested):
🏓 @dopenguin

PR #405, opened 2025-11-13
Edit dish export for new host

jedi-of-the-sea:
## Summary

After changing the internal host the dish export map plugin was broken. It appears, the wms and wfs displaying monumental features are not the same for the print as they are in the map. So the plugin has now new parameters to configure the url, version and feature type for the services. 
To make further adjustments easier, I also added configuration parameters for the different paths used in dish export map.

- set the mode in the index.html to `intern`
- use VPN to request monumental wfs and wms, you might also need CORS plugins in your browser
- select a monument feature
- click on the button with the PDF button (Kartendruck PDF)
- click on "Karte drucken" of the appearing dialog
- a new tab opens with the PDF and a map of the selected object

When testing for the first time, there will be a login mask for the dish software. You have to log in, then repeat the instructions above. Because of the missing login you won't get a map in the created pdf at first try.

- [x] Functionality has been tested on a smartphone -- the internal map does not (need to) work on smartphone

  - [x] Chrome Lighthouse
  - [x] Firefox Accessibility

### DIFF
--- a/packages/clients/dish/src/plugins/DishExportMap/DishExportMap.vue
+++ b/packages/clients/dish/src/plugins/DishExportMap/DishExportMap.vue
@@ -68,23 +68,6 @@ export default Vue.extend({
     showOverlay: false,
     dialog: false,
     title: '',
-    printApproach: '',
-    printRequester: '',
-    xPrint: '',
-    yPrint: '',
-    versionHintergrund: '',
-    proxyHintergrund: '',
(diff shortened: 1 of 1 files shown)


warm-coolguy (changes requested):
🏓 @jedi-of-the-sea

jedi-of-the-sea:
🏓 @warm-coolguy

PR #407, opened 2025-11-17
Add new parameter backend host

jedi-of-the-sea:
## Summary

Add new parameters added to make backend host configurable.

- set mode to "INTERN"
- select feature from monument service
- click button with PDF symbol
- click on "Karte drucken"
- beware that the first time might not work due to missing login (PDF will not have a map)
- login to dish and repeat the steps
- PDF should be generated

  - [x] Chrome Lighthouse
  - [x] Firefox Accessibility


warm-coolguy (approved):
🚀 @jedi-of-the-sea

PR #409, opened 2025-11-27
Dish client rename variable in url params

jedi-of-the-sea:
## Summary

Dish software devs asked for rename of the parameter 'printHost' due to similar variable in backend.

Missing documentation was added while editing this.

The configuration for urlParams have been edited to resemble the ones now used in the test environment for better debugging.

- set mode to "INTERN"
-  check if the WMS for monument services is rendered (you might need to be in VPN and use CORS Everywhere addon in Browser)

DISH EXPORT MAP testing:
-  select feature from monument service
-  click button with PDF symbol
-  click on "Karte drucken"
-  beware that the first time might not work due to missing login (PDF will not have a map)
-  login to dish and repeat the steps (pm me for login)
-  PDF should be generated

  - [x] Chrome Lighthouse
  - [x] Firefox Accessibility


packages/clients/dish/src/index.html
```diff
@@ -27,31 +27,31 @@
...
       client.createMap({
         containerId: 'polarstern',
-        mode: 'EXTERN', // INTERN, EXTERN
+        mode: 'INTERN', // INTERN, EXTERN
         urlParams,
         // only needed for the internal map
         configOverride: {
           gfi: {
-            internalHost: urlParams.internalHost,
+            internalHost: 'https://dish-testumgebung.dsc.dataport.de',
```
warm-coolguy: Is this intentional? I'd have expected this to be the result of something from above.

packages/clients/dish/src/index.html
```diff
@@ -27,31 +27,31 @@
...
-        internServicesBaseUrl: `${internalHost}:${internalPort}/${internalPath}`
+        printHostDeegree,
+        printServicesBaseUrl: `${printHostDeegree}:${printHostPort}/${printPath}`,
+        internServicesBaseUrl: `${internalHost}/${internalPath}`
       }
 
       client.createMap({
         containerId: 'polarstern',
-        mode: 'EXTERN', // INTERN, EXTERN
+        mode: 'INTERN', // INTERN, EXTERN
```
warm-coolguy: Since this is merely the test client, we may switch any time, but I feel like this is accidentally commited. We can also leave it like that if you're mostly working on `'INTERN'` atm. 🎩

warm-coolguy (approved):
🚀 @jedi-of-the-sea

PR #412, opened 2025-12-09
test(fullscreen): remove global coreStore mock (side-effects)

oeninghe-dataport:
## Summary

Make coreStore references in tests work again.

`npm run test`

## Why does this work?

The old `vi.mock` call wasn't limited to the tests it was supposed to work with.


dopenguin (approved):
🚀 @oeninghe-dataport 

Great! However, the tests for GeoLocation sadly still fail if not skipped. The issue there persists with `useCoreStore`.

PR #444, opened 2026-01-07
(#424) Pins coordinate should be null

### LINKED ISSUE #424: POLAR@3 (Pins): coordinate should be null
Just realized that I didn't check the functionality:

Bildschirmfoto 2025-12-16 um 21 33 11

EDIT: Note that this is unrelated to the implementation here but it seems like the service is currently not available. It works as expected tho.

However, on start, I get the following error. My guess is, that the `immediate` of the watcher triggers this unwanted call.
Bildschirmfoto 2025-12-16 um 21 36 37

_Originally posted by @dopenguin in https://github.com/Dataport/polar/issues/423#issuecomment-3662277751_

oeninghe-dataport:
## Summary

ReverseGeocoder does not error on loading.

- Open snowbox.
- See no error.

Closes #424

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
 			"hasInstallScript": true,
 			"license": "EUPL-1.2",
 			"devDependencies": {
-				"@actions/github": "^6.0.1",
+				"@actions/github": "^7.0.0",
 				"@commitlint/cli": "^20.2.0",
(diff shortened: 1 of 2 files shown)


dopenguin (approved):
🚀 @oeninghe-dataport

PR #454, opened 2026-01-12
Dish toggle label visibility with belonging geomlayer

jedi-of-the-sea:
## Summary

- label layer for monumental services will automatically made visible / invisible when toggling the visibility of matching geometry layer.

- mode to `INTERN`
- zoom in to see "beschriftungslayer" 
- toggle visibility of groups "Kulturdenkmale", "Objekte mit Kontrollbedarf", "Verlust" and check if the labels are switched on / of accordingly
- toggle visibility of layers of one of the groups named above and check if labels are switched on / of accordingly
- switch every group of: beschriftunglayer should dissapear from menu
- switch one group on again: beschriftungslayer should appear again in menu

  - [x] Chrome Lighthouse
  - [x] Firefox Accessibility

### DIFF
--- a/packages/clients/dish/CHANGELOG.md
+++ b/packages/clients/dish/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - Feature: Monumental label layer toggles its visibility depending on visible monumental layer geometries.
 - Fix: Add new configuration parameters for DishExportMap to configure different host (backend host might differ from `internalHost`) and to simplify adjustments for backend changes.
+- Fix: Add terms of use for internal map.
+- Fix: Open links for BKG and their terms of use in new tab.
 - Chore: Edit urlParams configuration for new testing environment.
 
 ## 1.2.0
--- a/packages/clients/dish/src/mapConfigurations/mapConfigIntern.ts
+++ b/packages/clients/dish/src/mapConfigurations/mapConfigIntern.ts
@@ -68,7 +68,7 @@ export const mapConfigIntern = (urlParams: DishUrlParams): DishMapConfig => ({
       attributionsAlkisWms,
     ],
     staticAttributions: [
-      `<span>Geobasisdaten: © GeoBasis-DE / <a href="https://www.bkg.bund.de/">BKG</a> 2024 <a href="http://sg.geodatenzentrum.de/web_public/nutzungsbedingungen.pdf">Nutzungsbedingungen</a></span>`,
+      `<span>Geobasisdaten: © GeoBasis-DE / <a href="https://www.bkg.bund.de/" target="_blank">BKG</a> <YEAR> <a href="http://sg.geodatenzentrum.de/web_public/nutzungsbedingungen.pdf" target="_blank">Nutzungsbedingungen</a></span>`,
       '<a href="#" onclick="window.openBenutzungshinweise(true)">Benutzungshinweise</a>',
     ],
   },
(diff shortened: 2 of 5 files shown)


packages/clients/dish/CHANGELOG.md
```diff
@@ -1,9 +1,12 @@
 # CHANGELOG
 
 ## unpublished
+
+- Feature: Monumental label layer toggle visibility depending on matching monumental geometry layer.
```
warm-coolguy:
```suggestion
- Feature: Monumental label layer toggles its visibility depending on visible monumental layer geometries.
```

I figure this is more what's done here, and there was a little typo. Do you agree, or did I make it worse?

packages/clients/dish/src/polar-client.ts
```diff
@@ -117,3 +125,87 @@ function zoomToInternalFeature(
...
+    })
+    .filter((s) => s)
+}
+
+function updateBeschriftungsLayer(instance: MapInstance, LAYERS: string) {
+  const olLabelLayer = getOlLabelLayer(instance)
+  const olSource = olLabelLayer?.getSource()
+  if (olSource) {
+    const updatedParams = { ...olSource.getParams(), LAYERS }
+    olSource.updateParams(updatedParams)
```
warm-coolguy: Maybe it'd help with one of the reported "bugs" (jumpy labels) to check whether LAYERS is the same as before. (Although I don't know whether order is deterministic?) 🎩

packages/clients/dish/src/polar-client.ts
```diff
@@ -40,6 +46,8 @@ export default {
     const objektId = parameters.get('ObjektID')
     if (mode === 'INTERN') {
       subscribeToExportedMap(map)
+      // watch for changes in activeMaskIds to update beschriftung layer
+      watchActiveMaskIds(map)
```
warm-coolguy: I suggest `watchActiveMaskIds` and all of its functions should be moved to a separate file in `utils`, since they work well as a module. That would make the separation of what does what and belongs where clearer.

packages/clients/dish/src/polar-client.ts
```diff
@@ -117,3 +125,85 @@ function zoomToInternalFeature(
...
+      getOlLabelLayer(instance)?.setVisible(false)
+    }
+  }
+
+  instance.$store.watch(
+    (_, getters) => ({
+      activeLayerIds: getters['plugin/layerChooser/activeLayerIds'],
+      activeMaskIds: getters['plugin/layerChooser/activeMaskIds'],
+    }),
+    updateLabelLayers
```
warm-coolguy:
```suggestion
    updateLabelLayers,
```

My fault, I forgot the comma on my last suggestion ...

>/home/runner/work/polar/polar/packages/clients/dish/src/polar-client.ts
>  166:4  error  Parsing error: ',' expected

jedi-of-the-sea:
> I didn't find instances of "Objekte mit Kontrollbedarf > Mehrheit von baulichen Anlagen mit Kontrollbedarf". All other (sub-)types I could test and see them work. Maybe there are just none of that specific type at the moment?
> 

I also did not find those, I guess there just are none? (to be sure I even checked with main branch)

> When disabling and/or enabling the labeled layers on 1:5000, the label layer will reappear despite being disabled. This does not happen on other resolutions. 
> 

Hopefully this will fix this behavior: 53d1532e6b614a4081efba6f4b6044c7f7fa0001

🏓 @warm-coolguy

warm-coolguy (approved):
🚀 @jedi-of-the-sea

PR #459, opened 2026-01-13
Add PR preview to GitHub pages

oeninghe-dataport:
## Summary

Provides a preview of the PR to GitHub pages.
This is limited on branches created here (in this repository, not in a fork) at the moment.

### DIFF
--- a/.github/workflows/preview.yaml
+++ b/.github/workflows/preview.yaml
@@ -29,11 +29,21 @@ jobs:
           node-version: 24.12.0
           registry-url: https://registry.npmjs.org/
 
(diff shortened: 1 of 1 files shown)


oeninghe-dataport:
> Screenshot 2026-01-14 133825
> Link and QR code don't work. Are they supposed to work already?

Yes, they are supposed to work. However, I think I located the issue here.
The pipeline on the `main` branch deploys to gh-pages and deletes everything there.
This probably also was an issue with the `next` documentation, that did not occur too often as there are many PRs on `next` and few on `main`.

However, this should probably be fixed.

PR #463, opened 2026-01-14
Feature/migrate scale

warm-coolguy:
## Summary

Migrating the Scale plugin to Vue@3.

In the process, I've dropped old util methods that are now done by native browser functions, and refactored some code for better pinia compatibility.

Configured in Snowbox (footer) and Iceberg (classic style) for checks.

  - [x] Chrome Lighthouse
  - [x] Firefox Accessibility

### DIFF
--- a/src/components/PolarSelect.ce.vue
+++ b/src/components/PolarSelect.ce.vue
@@ -7,6 +7,7 @@
 			<select
 				:id="id"
 				class="kern-form-input__select"
+				:aria-label="ariaLabel"
 				:disabled
 				:required
 				:multiple
@@ -16,7 +17,12 @@
 				<option v-if="defaultLabel" value="">{{ defaultLabel }}</option>
 				<!-- Intentional for straightforward API -->
 				<!-- eslint-disable-next-line vue/no-template-shadow -->
-				<option v-for="{ value, label } of options" :key="value" :value="value">
+				<option
+					v-for="{ value, label, ariaLabel } of options"
+					:key="value"
+					:value="value"
+					:aria-label="ariaLabel"
+				>
 					{{ label }}
 				</option>
 			</select>
@@ -30,9 +36,15 @@ import { computed, useId } from 'vue'
 const props = withDefaults(
 	defineProps<{
 		label?: string
+		ariaLabel?: string
 		defaultLabel?: string
 		value: string | string[]
-		options: { value: string | number; label: string; [key: string]: unknown }[]
+		options: {
+			value: string | number
+			label: string
+			ariaLabel?: string
+			[key: string]: unknown
+		}[]
 		disabled?: boolean
 		required?: boolean
 		multiple?: boolean
@@ -41,6 +53,7 @@ const props = withDefaults(
 	}>(),
 	{
 		label: '',
+		ariaLabel: undefined,
 		defaultLabel: '',
 		disabled: false,
 		required: false,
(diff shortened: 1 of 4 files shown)


src/core/stores/index.ts
```diff
@@ -100,6 +100,14 @@ export const useCoreStore = defineStore('core', () => {
 		 */
 		language: mainStoreRefs.language,
 
+		/**
+		 * Current zoom level of the map.
+		 *
+		 * @alpha
+		 * @readonly
+		 */
+		zoom: mainStoreRefs.zoom,
```
oeninghe-dataport: This property is not readonly. Therefore, the `@readonly` should be removed.

src/lib/dpi.ts
```diff
@@ -0,0 +1,75 @@
...
+		dpiDiv.setAttribute(
+			'style',
+			'position: absolute; height: 1in; width: 1in; top: -100%; left: -100%;'
+		)
+		body.appendChild(dpiDiv)
+
+		dpi = dpiDiv.offsetWidth * (window.devicePixelRatio || 1)
+		body.removeChild(dpiDiv)
+	} catch (e) {
+		console.error('@polar/plugin-scale', e)
```
oeninghe-dataport: Using the plugin name in `console.*` is no longer necessary. We use a custom vite plugin (see https://github.com/Dataport/polar/blob/next/vitePlugins/enrichedConsole.ts) that builds the relevant information on plugin and file into the logging automatically.

Besides that, as this is a global lib, the plugin name would be wrong anyway ;)

src/plugins/scale/store.ts
```diff
@@ -0,0 +1,113 @@
...
+		scaleToOne,
+
+		/** @internal */
+		scaleWithUnit,
+
+		/** @internal */
+		showScaleSwitcher,
+
+		/** @internal */
+		zoomOptions,
```
oeninghe-dataport: These properties should be marked at least as `@alpha` instead of `@internal` as they are relevant for plugin developers. They should also have a brief description in their TSDoc comment.

src/plugins/scale/types.ts
```diff
@@ -0,0 +1,12 @@
+import type { PluginOptions } from '@/core'
+
+export const PluginId = 'scale'
+
+export interface ScaleOptions extends PluginOptions {
+	/**
+	 * If set to `true`, the `1 : x` text will be replaced with a select element
+	 * that allows switching between scale values. Defaults to `false`. Requires
```
oeninghe-dataport: The "Defaults to `false`" could be replaced with `@defaultValue false`

src/components/PolarSelect.ce.vue
```diff
@@ -0,0 +1,97 @@
...
+const props = withDefaults(
+	defineProps<{
+		label?: string
+		defaultLabel?: string
+		value: string | string[]
+		options: { value: string | number; label: string; [key: string]: unknown }[]
+		disabled?: boolean
+		required?: boolean
+		multiple?: boolean
+		small?: boolean
```
oeninghe-dataport: As discussed, I'd prefer `size?: 'default' | 'small'` here.

src/components/PolarSelect.ce.vue
```diff
@@ -0,0 +1,97 @@
...
+	}
+}
+</script>
+
+<style scoped>
+.kern-form-input__select-wrapper.small {
+	height: var(--kern-metric-dimension-default);
+
+	.kern-form-input__select {
+		padding: 0 var(--kern-40) 0 var(--kern-metric-space-default);
```
oeninghe-dataport: Why the usage of `kern-40` instead of a semantic KERN variable? Is there no suitable `metric-space` size?

PR #465, opened 2026-01-14
refactor: rename core store file

warm-coolguy:
## Summary

In `src/core/stores`, renamed the `exports.ts` to `index.ts` to match usual pattern used throughout the codebase, according to our consensus.

I ran the examples just to check whether the automagically updated imports are right. Looks good to me.


oeninghe-dataport:
...and https://github.com/Dataport/polar/blob/next/typedoc.json#L6

oeninghe-dataport (approved):
🚀 @warm-coolguy

PR #475, opened 2026-01-15
Dish toggle layer visibility in alkis search

jedi-of-the-sea:
## Summary

When searching for a parcel, the user should be able to see the result in a layer.
After choosing a parcel search result, the alkis wms layer will be switched on and the zoom level will be adjusted so that the layer is visible. 

- switch to mode `INTERN`
- search for something, e.g. 'Lauenburg' or '12/3' 
- choose a search result from the parcel search (Flurstückssuche)
- the zoom level should be 10 or above so that the layer is visible
- the alkis layer should be switched on if the layer was unchecked before

- [x] Functionality has been tested in Firefox, Chrome, Safari (cannot really be tested in Chrome or Safari, because the alkis service is not available in either of those browsers)
- [x] Functionality has been tested on a smartphone (see above)

  - [x] Chrome Lighthouse
  - [x] Firefox Accessibility


packages/clients/dish/src/utils/watchSearchResultForAlkisSearch.ts
```diff
@@ -0,0 +1,46 @@
...
+    if (
+      configPatternKeysArray.every((item) =>
+        patternKeysSearchResult.includes(item)
+      )
+    ) {
+      const zoomLevel = instance.$store.getters['plugin/zoom/zoomLevel']
+
+      instance.$store.getters.map
+        .getView()
+        .setZoom(zoomLevel <= alkisMinZoom ? alkisMinZoom : zoomLevel)
```
warm-coolguy:
```suggestion
      if (zoomLevel <= alkisMinZoom) {
        instance.$store.getters.map
          .getView()
          .setZoom(alkisMinZoom)
      }
```

This avoids retriggering the `setZoom` from 10 to 10. I'm not sure if it matters at all, but think it's slightly better not to set the same value again to avoid getter recomputation across the system. 🎩

jedi-of-the-sea: I first used your version, but unfortunately when no zoom level is set, the search plugin seems to set its own zoom level and it will always zoom to 1:5000 after a search result was chosen. This is why I changed it back.

PR #483, opened 2026-01-19
feat: make font family selectable

warm-coolguy:
## Summary

Allow `fontFamily` configuration for map client.

I've ~~added~~ modified ~~a snippet~~ the dataportTheme in the snowbox that can be uncommented to check the results in it. I suggest trying `'Consolas'`, since it's easily distinguishable, and `''`, since it has a special meaning. When using `''`, set e.g. `font-family: Book Antiqua;` on the POLAR host node in the DOM manipulator to check whether it works as expected.

I'm currently having an issue with vite and couldn't check the build. Please verify that the fonts are bundled separately to the other .css files so that they can only optionally be imported during run-time.

- [x] Functionality has been tested in Firefox, Chrome, ~~Safari~~ (Please verify in Safari!)

  - [ ] Chrome Lighthouse
  - [ ] Firefox Accessibility

### DIFF
--- a/src/core/utils/loadKern.ts
+++ b/src/core/utils/loadKern.ts
@@ -37,6 +37,11 @@ export function loadKern(host: ShadowRoot, theme: Partial<KernTheme> = {}) {
 	`)
 	host.adoptedStyleSheets.push(kernSheet)
 
+	// @ts-expect-error | It's fine, we're getting `undefined` for an access on string, too.
+	if (typeof theme.typography?.font?.family?.default !== 'undefined') {
+		void import('@kern-ux/native/dist/fonts/fira-sans.css')
+	}
+
 	const kernTheme = buildKernTheme(theme)
 	host.adoptedStyleSheets.push(kernTheme)
 }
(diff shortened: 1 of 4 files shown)


src/core/utils/loadKern.ts
```diff
@@ -26,13 +26,18 @@ function buildKernTheme(theme: Partial<KernTheme>): CSSStyleSheet {
...
+	theme: Partial<KernTheme> = {},
+	fontFamily: string | undefined
+) {
 	host.adoptedStyleSheets.push(kernExtraIcons)
 
 	const kernSheet = new CSSStyleSheet()
 	kernSheet.replaceSync(`
 		@layer kern-ux {
 			${kernCss.replaceAll(':root', ':host')}
+			${typeof fontFamily !== 'undefined' ? `:host { --kern-typography-font-family-default: ${fontFamily}; }` : ''}
```
oeninghe-dataport: Why is this a separate parameter? Couldn't this be done using `theme`? (theme allows overriding arbitrary KERN variables, although the type probably needs to be adjusted.)

warm-coolguy:
> 🏓 @warm-coolguy

🏓 @oeninghe-dataport I have ... MOVED IT.

PR #503, opened 2026-01-29
Bugfix map export new tab param

jedi-of-the-sea:
## Summary

The previous implementation searched for a parameter "newTab" in the URL, but the parameter was not detected, most likely because it is named "NewTab". 

image

The fix makes the parameter detection case insensitive.

- set mode to 'INTERN' 

first step:
- visit "http://localhost:1234/?newTab=false" OR "http://localhost:1234/?NewTab=false" OR "http://localhost:1234/?newtab=false" OR "http://localhost:1234/?nEwTaB=false"
- select monument
- use dish map export 
image

- the map export should open in the same tab

second step:
- visit "http://localhost:1234"
- select monument
- use dish map export
- the map export should open in a new tab

  - [x] Chrome Lighthouse
  - [x] Firefox Accessibility

### DIFF
--- a/packages/clients/dish/CHANGELOG.md
+++ b/packages/clients/dish/CHANGELOG.md
@@ -1,6 +1,6 @@
 # CHANGELOG
 
-## unpublished
+## 1.3.1
 
 - Fix: Correct parameter detection for `NewTab` in DishMapExport Plugin.
 


packages/clients/dish/src/plugins/DishExportMap/DishExportMap.vue
```diff
@@ -72,10 +72,16 @@ export default Vue.extend({
...
-        ?.toLowerCase() === 'true',
+    newTab: (() => {
+      const params = new URL(document.location as unknown as string)
+        .searchParams
+      for (const [key, value] of params) {
+        if (key.toLowerCase() === 'newtab') {
+          return value.toLowerCase() === 'true'
+        }
+      }
+      return true
```
warm-coolguy: Previously, the default behaviour for unset `nEwTaB` was `false`, since `undefined === 'true'` is `false`. Is this an intentional change?

warm-coolguy (approved):
:shipit:  @jedi-of-the-sea

PR #509, opened 2026-02-02
feat(attributions): migrate plugin

### LINKED ISSUE #348: Migration of Attributions plugin to POLAR@3

dopenguin:
## Summary

Migration of attributions. Should work with `layout: 'nineRegions'` in all configuration options for `renderType` and as currently added in the snowbox as part of the footer.

- `npm run snowbox` and test the visibility of the attributions when toggling layers.
- Change the `layout` to `'nineRegions'` and test through the three situations for `renderType`. Don't forget to move the addition of the plugin to a different position. 

Closes #348

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3560,9 +3560,9 @@
 			"license": "MIT"
 		},
 		"node_modules/@types/node": {
-			"version": "24.11.0",
-			"resolved": "https://registry.npmjs.org/@types/node/-/node-24.11.0.tgz",
-			"integrity": "sha512-fPxQqz4VTgPI/IQ+lj9r0h+fDR66bzoeMGHp8ASee+32OSGIkeASsoZuJixsQoVef1QJbeubcPBxKk22QVoWdw==",
+			"version": "24.12.0",
+			"resolved": "https://registry.npmjs.org/@types/node/-/node-24.12.0.tgz",
+			"integrity": "sha512-GYDxsZi3ChgmckRT9HPU0WEhKLP08ev/Yfcq2AstjrDASOYCSXeyjDsHg4v5t4jOj7cyDX3vmprafKlWIG9MXQ==",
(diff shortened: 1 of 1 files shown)


src/plugins/attributions/utils/getVisibleLayers.ts
```diff
@@ -0,0 +1,18 @@
...
+ */
+export function getVisibleLayers(layers: Collection<BaseLayer>) {
+	const layerIDs: string[] = []
+	layers.forEach((layer) => {
+		// Only layers added through the services include the id property
+		if (layer.getVisible() && layer.get('id')) {
+			layerIDs.push(layer.get('id'))
+		}
+	})
+	return layerIDs
```
oeninghe-dataport:
```suggestion
	return layers.filter((layer) => layer.getVisible() && layer.get('id')).map((layer) => layer.get('id'))
```

src/plugins/attributions/components/NineLayoutAttributions.ce.vue
```diff
@@ -0,0 +1,65 @@
...
+			/>
+			<AttributionContent v-if="windowIsOpen" />
+		</template>
+		<template v-else>
+			<AttributionContent v-if="windowIsOpen" />
+			<PolarIconButton
+				:hint="
+					$t(($) => $.button[`${windowIsOpen ? 'close' : 'open'}Title`], {
+						ns: PluginId,
+					})
```
oeninghe-dataport: I'd prefer a solution here that keeps i18next's type safety.

Probably a solution similar to fullscreen would work well here; see https://github.com/Dataport/polar/blob/next/src/plugins/fullscreen/locales.ts#L19 and https://github.com/Dataport/polar/blob/next/src/plugins/fullscreen/components/FullscreenUI.ce.vue#L7

PR #525, opened 2026-02-09
Dish change service availability check

jedi-of-the-sea:
## Summary

Change configuration for `checkServiceAvailability` in `INTERN` mode because some services do not allow HEAD Requests.

 No real test necessary. Just start the map, there should be no toasts about services that are not avaiblable.

  - [x] Chrome Lighthouse
  - [x] Firefox Accessibility


packages/clients/dish/src/mapConfigurations/mapConfig.ts
```diff
@@ -7,6 +7,7 @@ import { mapConfigIntern } from './mapConfigIntern'
 import { mapConfigExtern } from './mapConfigExtern'
 
 const commonMapConfiguration: Partial<MapConfig> = {
+  checkServiceAvailability: false,
```
dopenguin:
```suggestion
```
It defaults to `false`

PR #556, opened 2026-02-23
feat: update layer type documentation

dopenguin:
## Summary

Allow layer type to be an arbitrary string so mask layers may be grouped freely.

Check in `iceberg` if you now can set the type of a layer to any value.

### DIFF
--- a/src/core/types/layer.ts
+++ b/src/core/types/layer.ts
@@ -25,8 +25,7 @@ export interface LayerConfigurationOptionLayers {
 	title?: boolean | Record<string, string>
 }
 
-// TODO: It should be allowed to set LayerType to a random string to allow for the grouping of different mask layers
-export type LayerType = 'background' | 'mask'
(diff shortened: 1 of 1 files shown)


warm-coolguy (changes requested):
🏓 @dopenguin 

Documentation on how the appearing titles (currently `dankTitle` for custom type `"dank"`) can be properly localized seems to be fitting as part of the `LayerType` documentation, as that would be my immediate follow-up question after trying the feature.

dopenguin:
> 🏓 @dopenguin
> 
> Documentation on how the appearing titles (currently `dankTitle` for custom type `"dank"`) can be properly localized seems to be fitting as part of the `LayerType` documentation, as that would be my immediate follow-up question after trying the feature.

Added c0681f33527b9e55039d56788d324737168fd548

warm-coolguy (approved):
:shipit: @dopenguin

PR #563, opened 2026-02-27
refactor(addressSearch): remove h2 element

dopenguin:
## Summary

Just use a non-semantic html tag and add connection to labelled element via `aria-labelledby`.

e.g. `npm run snowbox`


oeninghe-dataport (approved):
🚀 @dopenguin

PR #571, opened 2026-03-02
fix(reverseGeocoder): abort ongoing requests if a new reverse geocode request comes in

dopenguin:
## Summary

As the title states, reverse geocoding is now being aborted once a new request comes in. If this were not to be added, I could e.g. drag the pin around a few times and the map would keep centering on ALL position.

Drag the pin around a few times in `iceberg`. It should now only center the map on the last position.

### DIFF
--- a/src/plugins/reverseGeocoder/store.ts
+++ b/src/plugins/reverseGeocoder/store.ts
@@ -9,7 +9,7 @@ import type { Mock } from 'vitest'
 import { easeOut } from 'ol/easing'
 import { Point } from 'ol/geom'
 import { acceptHMRUpdate, defineStore } from 'pinia'
-import { computed, type Reactive } from 'vue'
+import { computed, type Reactive, ref, toRaw } from 'vue'
 
 import { usePluginStoreWatcher } from '@/composables/usePluginStoreWatcher'
 import { useCoreStore } from '@/core/stores'
@@ -33,6 +33,8 @@ export const useReverseGeocoderStore = defineStore(
 	'plugins/reverseGeocoder',
 	() => {
 		const coreStore = useCoreStore()
+
+		const abortController = ref<AbortController | null>(null)
 
 		const configuration = computed(
 			() => coreStore.configuration[PluginId] as ReverseGeocoderPluginOptions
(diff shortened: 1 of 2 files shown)


oeninghe-dataport (changes requested):
🏓 @dopenguin

warm-coolguy (changes requested):
🏓 @dopenguin @oeninghe-dataport I closed the prior thread; if that pattern is to be changed, it should happen globally as a separate effort as proposed, and should prior to change be discussed in our Weekly.

Skimming the code, I found various instances of `signal.aborted` checks that were, in other spots, used to prevent spamming the console. I propose this technique is used here, too, to avoid confusion regarding "unerroring errors" further down the line.

dopenguin:
> 🏓 @dopenguin @oeninghe-dataport I closed the prior thread; if that pattern is to be changed, it should happen globally as a separate effort as proposed, and should prior to change be discussed in our Weekly.
> 
> Skimming the code, I found various instances of `signal.aborted` checks that were, in other spots, used to prevent spamming the console. I propose this technique is used here, too, to avoid confusion regarding "unerroring errors" further down the line.

So you propose moving the `AbortController` from the util to the store and passing the signal to the utility function?

warm-coolguy:
> > 🏓 @dopenguin @oeninghe-dataport I closed the prior thread; if that pattern is to be changed, it should happen globally as a separate effort as proposed, and should prior to change be discussed in our Weekly.
> > Skimming the code, I found various instances of `signal.aborted` checks that were, in other spots, used to prevent spamming the console. I propose this technique is used here, too, to avoid confusion regarding "unerroring errors" further down the line.
> 
> So you propose moving the `AbortController` from the util to the store and passing the signal to the utility function?
> 
> 🏓 @warm-coolguy @oeninghe-dataport

Yes, please.

🏓 @dopenguin

dopenguin:
> > > 🏓 @dopenguin @oeninghe-dataport I closed the prior thread; if that pattern is to be changed, it should happen globally as a separate effort as proposed, and should prior to change be discussed in our Weekly.
> > > Skimming the code, I found various instances of `signal.aborted` checks that were, in other spots, used to prevent spamming the console. I propose this technique is used here, too, to avoid confusion regarding "unerroring errors" further down the line.
> > 
> > 
> > So you propose moving the `AbortController` from the util to the store and passing the signal to the utility function?
> > 🏓 @warm-coolguy @oeninghe-dataport
> 
> Yes, please.
> 
> 🏓 @dopenguin

Been there, done that b032ba5c4294af0f14541a0c2baad9dd39b02843

🏓 @warm-coolguy

warm-coolguy (approved):
:shipit: @dopenguin

PR #574, opened 2026-03-03
fix: cleanup after a map has been removed

dopenguin:
## Summary

Some parts were not properly being cleaned up leading to an explosion in RAM usage after multiple additions and removals of an additional map in `snowbox`. This, too, would be the case for any SPA that rerenders POLAR.

`npm run snowbox` here and on `next`. Check the Snapshot size difference after multiple additions and removals. It should be fairly less now.
Note that there is still some overhead of about 100kb - 200kb per instance. I was not able to remove that and it may be an issue with `vue` directly.

### DIFF
--- a/vite.config.github-io.ts
+++ b/vite.config.github-io.ts
@@ -1,7 +1,6 @@
 import vue from '@vitejs/plugin-vue'
 import { resolve } from 'node:path'
 import { defineConfig } from 'vite'
-
 import kernExtraIcons from 'vite-plugin-kern-extra-icons'
 
 export default defineConfig({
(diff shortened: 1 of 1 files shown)


warm-coolguy (changes requested):
🏓 @dopenguin 

Compared `next` to `vue3/fix-memory-leak` on Windows, Chrome. Ran `npm run dev` with both branches.

Testing sequence:

1. Load snowbox, wait a little, take snapshot
2. Add 20 clients, remove all clients
3. Take snapshot
4. GOTO 2

Repeated testing resulted in an additional 5MB snapshot size per loop. Growth behaviour was identical between branches. Didn't review the code since changes seem to be ineffective.

PR #585, opened 2026-03-06
feat: add plugin store watcher

### LINKED ISSUE #484: Plugin watchers don't work if another plugin is added later on
**Describe the bug**

In `reverseGeocoder` and `pins` the configuration parameters `coordinateSources` are being used to add watchers to react to changes in another plugins store. However, if the watched plugin is added with `addPlugin` after the plugin that is watching, the watcher can not be established.

**Expected behavior**

It should be possible to add plugins in any order and have watchers connect / disconnect on the fly.

**Steps to reproduce**

1. `npm run snowbox`
2. Setting a pin does not trigger a reverse geocode

jedi-of-the-sea:
## Summary

This PR should solve the problem of the dependency of plugin orders by adding a reusable plugin store watcher. It can be used within plugins to manahge watchers for stores from other plugins. 
 
Within this PR, plugin store watchers where added to `geoLocation` plugin and `pins` plugin.

⚠️ Please note: to test the new plugin watchers, the order of the plugins `geolocation`, `addressSearch`, `attributions`, `layerChooser` and `pins` within the snowbox client was changed.

- Start the snowbox client
------------------------
- click in the map (stay in the vicinity of Hamburg)
- check if the address of the place that was clicked on the map is displayed in the search result input field.
------------------------
- search for address
- check if map zooms to search result and a pin is put on the map 
------------------------
- open the attributions
- change layer visibiliity in layer chooser
- check if the attributions change accordingly
------------------------

  - [x] Chrome Lighthouse
  - [x] Firefox Accessibility

closes https://github.com/Dataport/polar/issues/484


src/core/composables/usePluginStoreWatcher.ts
dopenguin: There should be some console warnings if something is not working when setting up the watchers.

dopenguin (changes requested):
Attributions also need this new composable.

🏓 @jedi-of-the-sea

src/composables/usePluginStoreWatcher.ts
```diff
@@ -0,0 +1,178 @@
...
+		if (!watcherConfig.source.plugin) {
+			watcherConfig.handle = watch(
+				() =>
+					(coreStore as unknown as Record<string, unknown>)[
+						watcherConfig.source.key
+					],
+				watcherConfig.callback
+			)
+			return
+		}
```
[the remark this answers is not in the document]
jedi-of-the-sea: 2aac0a44071fed3a8a7d0fba80319526e6903c3c

jedi-of-the-sea:
> Attributions also need this new composable.
> 
fecfba85a494cceb916c355c2b66e26e40fe5729

🏓  @dopenguin

dopenguin:
> I noticed an inconsistency within the attributions plugin when a layer is selected but not visible in the current zoom level (can be tested in client "snowbox"). While selecting / deselecting layers, the layer "Ausgleichsflächen" is sometimes listed in the attributions and sometimes not. This seems to be already a problem within the `next` branch, so I suggest I ignore it in this branch and it gets a separate issue. Would you aggree?

This either seems to have already been fixed on `next` or I can not reproduce this bug at least for iceberg.
Please update the branch so we may see if this is the case for this PR as well.

However, this does not seem to work at all for iceberg whereas it works for snowbox without any problems.

src/plugins/reverseGeocoder/store.ts
```diff
@@ -37,35 +38,19 @@ export const useReverseGeocoderStore = defineStore(
...
 				}
-				watchHandles.value.push(
-					watch(
-						() => store[source.key],
-						async (coordinate) => {
-							if (coordinate) {
-								await reverseGeocode(coordinate)
-							}
-						},
-						{ immediate: true }
```
dopenguin, line 58: It would be nice if `usePluginStoreWatcher` would accept additional options so it would be possible to e.g. pass the above `immediate: true` to the underlying watcher

src/plugins/attributions/store.ts
```diff
@@ -59,20 +53,12 @@ export const useAttributionsStore = defineStore('plugins/attributions', () => {
...
-				continue
-			}
-			watchHandles.value.push(
-				watch(() => store[listenReference.key], updateLayers)
-			)
-		}
+	usePluginStoreWatcher(
+		() => configuration.value.listenToChanges || [],
+		updateLayers
+	)
```
dopenguin: This does not seem to work yet (in iceberg)

src/plugins/attributions/store.ts
```diff
@@ -32,14 +31,9 @@ export const useAttributionsStore = defineStore('plugins/attributions', () => {
...
 	const windowIsOpen = ref(false)
 
-	const watchHandles = ref<WatchHandle[]>([])
-
 	const configuration = computed(
 		() => coreStore.configuration.attributions || {}
 	)
-	const listenToChanges = computed<StoreReference[]>(
-		() => configuration.value.listenToChanges || []
-	)
```
dopenguin, line 42: I'd prefer keeping this getter. Might directly solve the issue with `iceberg`

dopenguin (changes requested):
🏓 @jedi-of-the-sea 

Please also rebase the branch / merge the changes from `next` onto this branch

jedi-of-the-sea:
> > I noticed an inconsistency within the attributions plugin when a layer is selected but not visible in the current zoom level (can be tested in client "snowbox"). While selecting / deselecting layers, the layer "Ausgleichsflächen" is sometimes listed in the attributions and sometimes not. This seems to be already a problem within the `next` branch, so I suggest I ignore it in this branch and it gets a separate issue. Would you aggree?
> 
> This either seems to have already been fixed on `next` or I can not reproduce this bug at least for iceberg. Please update the branch so we may see if this is the case for this PR as well.
> 
> However, this does not seem to work at all for iceberg whereas it works for snowbox without any problems.

I can still reproduce the bug in the updated `next` branch for the snowbox client. But in iceberg, the listener for the attributions does not seem to work at all? The videos are from my tests with the `next` branch

snowbox:
https://github.com/user-attachments/assets/d3254647-924f-42d7-9d88-035fec4f5272

iceberg:
https://github.com/user-attachments/assets/0631aace-2021-4c01-9923-a8b03953da8b

EDIT: i just checked the config and iceberg has no `listenToChanges` config parameter, so this is why it won't work there. So iceberg is not suited to test the plugin watcher in the `attributions plugin`.

src/plugins/attributions/store.ts
jedi-of-the-sea: cannot work in iceberg, see https://github.com/Dataport/polar/pull/585#issuecomment-4204488484

src/plugins/attributions/store.ts
jedi-of-the-sea: cannot work in iceberg, see https://github.com/Dataport/polar/pull/585#issuecomment-4204488484

dopenguin:
> > > I noticed an inconsistency within the attributions plugin when a layer is selected but not visible in the current zoom level (can be tested in client "snowbox"). While selecting / deselecting layers, the layer "Ausgleichsflächen" is sometimes listed in the attributions and sometimes not. This seems to be already a problem within the `next` branch, so I suggest I ignore it in this branch and it gets a separate issue. Would you aggree?
> > 
> > 
> > This either seems to have already been fixed on `next` or I can not reproduce this bug at least for iceberg. Please update the branch so we may see if this is the case for this PR as well.
> > However, this does not seem to work at all for iceberg whereas it works for snowbox without any problems.
> 
> I can still reproduce the bug in the updated `next` branch for the snowbox client. But in iceberg, the listener for the attributions does not seem to work at all? The videos are from my tests with the `next` branch
> 
> snowbox: https://github.com/user-attachments/assets/d3254647-924f-42d7-9d88-035fec4f5272
> 
> iceberg: https://github.com/user-attachments/assets/0631aace-2021-4c01-9923-a8b03953da8b
> 
> EDIT: i just checked the config and iceberg has no `listenToChanges` config parameter, so this is why it won't work there. So iceberg is not suited to test the plugin watcher in the `attributions plugin`.

The config has been updated in 6cfe0dce.

I'll take a closer look regarding the inconsistencies of the attributions

src/plugins/attributions/store.ts
dopenguin: Still, prefer keeping it :D

src/plugins/reverseGeocoder/store.ts
jedi-of-the-sea: e1300dc04cda937263456ba8f49d9fe59ccac73d

src/plugins/attributions/store.ts
jedi-of-the-sea: da607e574a6676ba67753ad71d8954764d145cce

jedi-of-the-sea:
> The config has been updated in [6cfe0dc]

Ah ok. I found the same bug here now, too.

> I'll take a closer look regarding the inconsistencies of the attributions

it seems to be a bug in `getVisibleLayers` - sometimes, `layer.getVisible()` returns true when the layer is selected but invisible due to to zoom level (e.g. when starting the client), sometimes it returns false (e.g. after zooming out again after the layer was visible in a zoom level and is not anymore in the next). Don't know why

🏓 @dopenguin

dopenguin (approved):
🚀 @jedi-of-the-sea

PR #587, opened 2026-03-09
fix(core): fix width and height of map overlay

dopenguin:
## Summary

Update the width and the height of the map overlay. This was broken after #545.
Note that the overlay now is placed below the interface elements. I think this is quite nice. If you disagree, I'll change it back to the previous implementation.

`npm run snowbox` and try scrolling via mousewheel.

### DIFF
--- a/src/core/components/PolarContainer.ce.vue
+++ b/src/core/components/PolarContainer.ce.vue
@@ -102,7 +102,7 @@ function updateListeners() {
 		!hasWindowSize.value &&
 		polarMapContainer.value &&
 		polarMapContainer.value.el &&
-		mainStore.hasSmallDisplay
+		hasSmallDisplay.value
 	) {
 		hammer?.destroy()
 		hammer = new Hammer(polarMapContainer.value.el).on('pan', (e) => {
--- a/src/core/components/PolarContainer.ce.vue
+++ b/src/core/components/PolarContainer.ce.vue
@@ -5,8 +5,9 @@
 		:lang="language"
 		:data-kern-theme="mainStore.colorScheme"
 	>
+		<PolarMapOverlay ref="polar-map-overlay" />
 		<div class="polar-map-layer">
-			<PolarMap />
+			<PolarMap ref="polar-map-container" @wheel="wheelEffect" />
 		</div>
 		<div class="polar-ui-layer">
 			<div v-if="!hasWindowSize" class="polar-shadow" aria-hidden="true" />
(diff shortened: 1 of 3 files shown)


warm-coolguy (changes requested):
🏓 @dopenguin 

Due to "Press Ctrl + Scroll to zoo", I prefer the prior solution of overlapping it all.

Image displaying that the current client cuts off the scroll instructions if UI elements are above it, rendering it ineffective

dopenguin:
> 🏓 @dopenguin
> 
> Due to "Press Ctrl + Scroll to zoo", I prefer the prior solution of overlapping it all.
> Image displaying that the current client cuts off the scroll instructions if UI elements are above it, rendering it ineffective

Took some more effort, but works now 488c3cc8edbc20d7edffb32ef369add2679f0cad

warm-coolguy (changes requested):
🏓 @dopenguin While code regarding the one finger pan is present, it does work on the snowbox example in neither Android Chrome nor Safari Firefox. Please double-check.

It also occurred to me that NVDA does not read the hint. I figure that it should, in an utmost polite fashion. Do you agree? If so, I'll make it a separate issue.

dopenguin:
> 🏓 @dopenguin While code regarding the one finger pan is present, it does work on the snowbox example in neither Android Chrome nor Safari Firefox. Please double-check.
> 
> It also occurred to me that NVDA does not read the hint. I figure that it should, in an utmost polite fashion. Do you agree? If so, I'll make it a separate issue.

Good catch, works now df6130c4e0dc2dec155a4284d1cc7f921cc8267e!

Screenreaders reading the hint seems like a good idea, please add an issue!

🏓 @warm-coolguy

warm-coolguy (changes requested):
🏓 @dopenguin 

On Safari mobile, in FullScreen mode, the hint is shown, albeit the map is panning. Please also check internal channel notes on Android behaviour, where we asked the team to check it on their devices.

Edit: According to tester feedback, Firefox seems to only sometimes display the hint.

Edit: According to tester feedback, with two fingers, both the map pan and the page scroll.

dopenguin:
> 🏓 @dopenguin
> 
> On Safari mobile, in FullScreen mode, the hint is shown, albeit the map is panning. Please also check internal channel notes on Android behaviour, where we asked the team to check it on their devices.
> 
> Edit: According to tester feedback, Firefox seems to only sometimes display the hint.
> 
> Edit: According to tester feedback, with two fingers, both the map pan and the page scroll.

The overlay should now be hidden when switching to fullscreen 9caea803eed9318e334ae1eeb1903ca698654b8c.

I was not able to reproduce the behaviour of Firefox sometimes displaying the hint when it is not supposed to nor that the page scroll when panning with two fingers. Both was tested on an iPhone 15 Pro with iOS 26.4.1 and the latest versions of Firefox, Chrome and Safari.

🏓 @warm-coolguy

warm-coolguy (approved):
:shipit: @dopenguin

PR #606, opened 2026-03-23
POLAR@3: Add nominatim support

oeninghe-dataport:
## Summary

Add support for nominatim in addressSearch and reverseGeocoder

Snowbox-integrated


dopenguin:
@oeninghe-dataport Please update the PR. Will review afterwards

oeninghe-dataport:
🏓 @dopenguin  Updated

> @oeninghe-dataport Please update the PR. Will review afterwards

examples/snowbox/index.js
```diff
@@ -517,6 +521,12 @@ addPlugin(
 				type: 'mpapi',
 				url: 'https://geodienste.hamburg.de/HH_WFS_GAGES?service=WFS&request=GetFeature&version=2.0.0',
 			},
+			*/
+			{
+				type: 'nominatim',
+				url: 'https://polar.dataport.de/nominatim/search',
+				queryParameters: {},
```
dopenguin:
```suggestion
```
The parameter is optional

src/plugins/reverseGeocoder/utils/reverseGeocodeWps.ts
```diff
@@ -32,9 +32,10 @@ const parser = new Parser({
 	tagNameProcessors: [processors.stripPrefix],
 })
 
-export async function reverseGeocode(
+export async function reverseGeocodeWps(
 	url: string,
 	coordinate: [number, number],
+	_epsg: string,
```
dopenguin: I know this is done because the code in `src/plugins/reverseGeocoder/store.ts` is shorter this way, but I strongly prefer not adding a redundant parameter here.

src/lib/getFeatures/types.ts
```diff
@@ -288,3 +288,85 @@ export interface WfsParameters extends QueryParameters {
...
+	country?: string
+
+	/**
+	 * Country code (ISO 3166-1alpha2).
+	 */
+	countrycodes?: string[]
+
+	/**
+	 * County.
+	 */
```
dopenguin, line 325: The comments, beside the one on `countrycodes`, seem redundant.

src/lib/getFeatures/types.ts
```diff
@@ -288,3 +288,85 @@ export interface WfsParameters extends QueryParameters {
 	 */
 	useRightHandWildcard?: boolean
 }
+
+export interface NominatimParameters extends QueryParameters {
+	/**
+	 * Include address details in result?
```
dopenguin: Please include a description like the one for `bounded` that is not a question.

src/lib/getFeatures/types.ts
```diff
@@ -288,3 +288,85 @@ export interface WfsParameters extends QueryParameters {
...
+	 */
+	countrycodes?: string[]
+
+	/**
+	 * County.
+	 */
+	county?: string
+
+	/**
+	 * Include entrances to buildings in results?
```
dopenguin: Please include a description like the one for `bounded` that is not a question.

src/lib/getFeatures/types.ts
```diff
@@ -288,3 +288,85 @@ export interface WfsParameters extends QueryParameters {
...
+	 */
+	entrances?: boolean
+
+	/**
+	 * Add extra tags to result?
```
dopenguin: Please include a description like the one for `bounded` that is not a question.

src/lib/getFeatures/types.ts
```diff
@@ -288,3 +288,85 @@ export interface WfsParameters extends QueryParameters {
...
+	 */
+	featureType?: 'country' | 'state' | 'city' | 'settlement'
+
+	/**
+	 * Which type(s) of results to return.
+	 */
+	layer?: ('address' | 'poi' | 'railway' | 'natural' | 'manmade')[]
+
+	/**
+	 * Include more names in alternate languages?
```
dopenguin: Please include a description like the one for `bounded` that is not a question.

src/lib/getFeatures/types.ts
```diff
@@ -288,3 +288,85 @@ export interface WfsParameters extends QueryParameters {
...
+
+	/**
+	 * Postal code.
+	 */
+	postalcode?: string
+
+	/**
+	 * State.
+	 */
+	state?: string
```
dopenguin, line 361: These comments seem redundant

src/plugins/reverseGeocoder/utils/reverseGeocodeNominatim.ts
```diff
@@ -0,0 +1,53 @@
...
+		'EPSG:4326'
+	) as [number, number]
+
+	const fetchUrl = new URL(url)
+	fetchUrl.searchParams.set('lat', searchCoordinate[1].toString())
+	fetchUrl.searchParams.set('lon', searchCoordinate[0].toString())
+	fetchUrl.searchParams.set('format', 'jsonv2')
+
+	const result = await fetch(fetchUrl, { signal }).then((response) =>
+		response.json()
```
dopenguin: I'd prefer the return of `json` or the variable `result` to be typed.

src/plugins/reverseGeocoder/utils/reverseGeocodeNominatim.ts
```diff
@@ -0,0 +1,53 @@
...
+			// as returned by reverse geocoder
+			coordinates: transformCoordinate(
+				[result.lon, result.lat],
+				'EPSG:4326',
+				epsg
+			),
+			type: 'Point',
+		},
+	}
+	return resultObject
```
dopenguin:
```suggestion
	return {
		type: 'reverse_geocoded',
		title: [
			[result.address.road, result.address.house_number]
				.filter((x) => x)
				.join(' '),
			result.address.town || result.address.city || result.address.village,
		]
			.filter((x) => x)
			.join(', '),
		properties: result.properties,
		geometry: {
			// as clicked by user - usually want to keep this since user is pointing at something
			coordinates: coordinate,
			type: 'Point',
		},
		addressGeometry: {
			// as returned by reverse geocoder
			coordinates: transformCoordinate(
				[result.lon, result.lat],
				'EPSG:4326',
				epsg
			),
			type: 'Point',
		},
	}
```

src/plugins/reverseGeocoder/utils/reverseGeocodeNominatim.ts
```diff
@@ -0,0 +1,53 @@
...
+		type: 'reverse_geocoded',
+		title: [
+			[result.address.road, result.address.house_number]
+				.filter((x) => x)
+				.join(' '),
+			result.address.town || result.address.city || result.address.village,
+		]
+			.filter((x) => x)
+			.join(', '),
+		properties: result.properties,
```
dopenguin: This is an example response from the service. This does not seem usable e.g. for the addressSearch. What am I missing?

Image

src/lib/getFeatures/nominatim.ts
```diff
@@ -0,0 +1,62 @@
...
+		...Object.fromEntries(
+			Object.entries(native).map(([key, value]) => [
+				key,
+				Array.isArray(value) ? value.join(',') : value,
+			])
+		),
+		...(maxFeatures ? { limit: maxFeatures } : {}),
+		q: inputValue,
+		format: 'geojson',
+		polygon_geojson: 0, // TODO: Setting this to 1 is currently not supported well by addressSearch plugin. The plugin expects a point geometry.
```
dopenguin: If this is something we want to change, then this should rather be an issue.
However, the information seems important and should be added as a regular comment above this line.

src/lib/getFeatures/nominatim.ts
```diff
@@ -0,0 +1,62 @@
...
+		polygon_threshold: 5,
+	}).forEach(([key, value]) => {
+		fetchUrl.searchParams.set(key, value.toString())
+	})
```
dopenguin: This is overly complex and does not really safe anything compared to the straightforward implementation:

```ts
	for (const [key, value] of Object.entries(native)) {
		fetchUrl.searchParams.set(
			key,
			Array.isArray(value) ? value.join(',') : String(value)
		)
	}
	if (maxFeatures) {
		fetchUrl.searchParams.set('limit', String(maxFeatures))
	}
	fetchUrl.searchParams.set('q', inputValue)
	fetchUrl.searchParams.set('format', 'geojson')
	// Setting this to 1 is currently not supported well by addressSearch plugin. The plugin expects a point geometry.
	fetchUrl.searchParams.set('polygon_geojson', '0')
	fetchUrl.searchParams.set('polygon_threshold', '5')
```

src/lib/getFeatures/nominatim.ts
```diff
@@ -0,0 +1,62 @@
...
+
+	return await fetch(fetchUrl, {
+		signal,
+	})
+		.then((response) => response.json())
```
dopenguin: The response value of `json` should be typed IMO

src/lib/getFeatures/nominatim.ts
```diff
@@ -0,0 +1,62 @@
...
+					...feature.geometry,
+					coordinates:
+						epsg === 'EPSG:4326'
+							? feature.geometry.coordinates
+							: feature.geometry.type === 'Polygon'
+								? feature.geometry.coordinates.map((ring) =>
+										ring.map((coord) =>
+											transformCoordinates(coord, 'EPSG:4326', epsg)
+										)
+									)
```
dopenguin: As we currently do not support the service returning Polygons, this is essentially dead code walking

dopenguin (changes requested):
Can we further specify the search so the results are relevant for the client? The default seems ... not helpful :D

Image

🏓 @oeninghe-dataport

oeninghe-dataport:
> Can we further specify the search so the results are relevant for the client? The default seems ... not helpful :D
> 
> Image
> 🏓 @oeninghe-dataport

It should be noted that the POLAR nominatim service is currently (due to server constraints) limited to geocoding in Lower Saxony.

src/plugins/reverseGeocoder/utils/reverseGeocodeWps.ts
oeninghe-dataport: I think the information would actually be quite relevant here (though it's unused yet, which has to change).
IIRC, our WPS geocoding only works if the global EPSG and the WPS's EPSG match. We should transform the coordinates here from global EPSG to a configurable WPS-EPSG.

src/lib/getFeatures/types.ts
oeninghe-dataport: As this is public API, I'd prefer to have a documentation for each key, although it might be trivial.

It is also documented upstream:
https://nominatim.org/release-docs/latest/api/Search/#structured-query

src/lib/getFeatures/types.ts
oeninghe-dataport: b944581a17

src/lib/getFeatures/types.ts
oeninghe-dataport: b944581a17

src/lib/getFeatures/types.ts
oeninghe-dataport: b944581a17

src/lib/getFeatures/types.ts
oeninghe-dataport: b944581a17

src/lib/getFeatures/types.ts
oeninghe-dataport: See https://github.com/Dataport/polar/pull/606#discussion_r3304650240

src/plugins/reverseGeocoder/utils/reverseGeocodeNominatim.ts
oeninghe-dataport: I don't see your request, but maybe consider https://github.com/Dataport/polar/pull/606#issuecomment-4543144985 ?

src/lib/getFeatures/nominatim.ts
oeninghe-dataport: Changed with 6a58368bcf4bc1f7bab65c8c164cc671de4655dc and modeled as issue in https://github.com/Dataport/polar/issues/747

src/lib/getFeatures/nominatim.ts
oeninghe-dataport: Agreed. 9cae6a1731

src/lib/getFeatures/nominatim.ts
oeninghe-dataport: d19d95a16c597aa018d097332ec33a05a0f1d6ae

src/lib/getFeatures/nominatim.ts
oeninghe-dataport: I refactored this w/ 169d26efef88a1beed0b527683769e65868e9a31. Polygons are probably still dead code, however, I think it is fine to leave it there.

oeninghe-dataport:
🏓 @dopenguin

src/plugins/reverseGeocoder/utils/reverseGeocodeWps.ts
dopenguin: If it is important, that the maps projection matches the WPS's projection, then please add a separate PR that addresses this issue.

src/plugins/reverseGeocoder/utils/reverseGeocodeNominatim.ts
dopenguin: Yup, sending a reverse geocoding request for a coordinate in Hamburg yielded the above result 🎩

src/lib/transformGeometry.ts
```diff
@@ -0,0 +1,40 @@
+import type { Geometry, Point, Polygon } from 'geojson'
+
+import { transform as transformCoordinates } from 'ol/proj'
+
+/**
+ * Transforms the coordinates of a GeoJSON geometry from one EPSG projection to another.
+ *
+ * @remarks
+ * Currently supports only Point and Polygon geometries.
```
dopenguin:
```suggestion
 * Currently, supports only Point and Polygon geometries.
```

src/lib/getFeatures/wfs/parse.ts
```diff
@@ -1,9 +1,9 @@
-import { toMerged } from 'es-toolkit'
 import { GeoJSON, WFS } from 'ol/format'
-import { transform as transformCoordinates } from 'ol/proj'
 
 import type { PolarGeoJsonFeature, PolarGeoJsonFeatureCollection } from '@/core'
 
+import { transformGeometry } from '@/lib/transformGeometry'
```
dopenguin, line 5: Will this be the correct import, if #750 is merged or will there be a relative path? Especially as a relative path is used in `src/lib/getFeatures/bkg.ts`.

In this scenario, I prefer the way it is written in this file.

src/lib/getFeatures/bkg.ts
```diff
@@ -58,16 +58,11 @@ export default async function (
...
-										(feature.geometry as Point).coordinates,
-										'EPSG:4326',
-										queryParameters.epsg
-									),
-					}),
+					geometry: transformGeometry(
+						feature.geometry,
+						'EPSG:4326',
+						queryParameters.epsg
+					),
```
dopenguin, line 64: `transformCoordinates` includes the information

> If there is no available transform between the two projection, the function will throw an error.

So a call with the same projection does not yield an error, correct?

dopenguin (changes requested):
> > Can we further specify the search so the results are relevant for the client? The default seems ... not helpful :D
> > Image
> > 🏓 @oeninghe-dataport
> 
> It should be noted that the POLAR nominatim service is currently (due to server constraints) limited to geocoding in Lower Saxony.

Is it possible to update the POLAR nominatim service to include Hamburg as the snowbox typically uses a lot of Hamburg specific services?

🏓 @oeninghe-dataport

src/lib/getFeatures/wfs/parse.ts
oeninghe-dataport: This is now, that #750 is merged and also merged here, an absolute path.

src/lib/getFeatures/bkg.ts
oeninghe-dataport: Correct. The identity transformation is always registered automatically: https://github.com/openlayers/openlayers/blob/v10.9.0/src/ol/proj.js#L170-L180

src/plugins/reverseGeocoder/utils/reverseGeocodeWps.ts
oeninghe-dataport: As the change is quite trivial (although, as there are tests, the commit has a few more changes than I expected), I just added the fix here: b902a4fb1b

Otherwise, there would probably occur some merge conflicts.

src/plugins/reverseGeocoder/utils/reverseGeocodeNominatim.ts
oeninghe-dataport: ad673714b5

oeninghe-dataport:
🏓 @dopenguin

src/plugins/reverseGeocoder/utils/reverseGeocodeNominatim.ts
```diff
@@ -33,7 +74,7 @@ export async function reverseGeocodeNominatim(
 		]
 			.filter((x) => x)
 			.join(', '),
-		properties: result.properties,
+		properties: {},
```
dopenguin: Image

src/plugins/reverseGeocoder/utils/reverseGeocodeNominatim.ts
```diff
@@ -1,13 +1,54 @@
...
+		city_district?: string
+		city?: string
+		county?: string
+		state_district?: string
+		state?: string
+		// eslint-disable-next-line @typescript-eslint/naming-convention
+		'ISO3166-2-lvl4'?: string
+		postcode?: string
+		country?: string
+		country_code?: string
```
dopenguin, line 23: It feels really off that all of these are optional. Also, a link to the documentation seems relevant.

src/plugins/reverseGeocoder/utils/reverseGeocodeNominatim.ts
```diff
@@ -1,13 +1,54 @@
...
-export async function reverseGeocodeNominatim(
-	url: string,
-	coordinate: [number, number],
-	epsg: string,
+export async function reverseGeocodeNominatim({
+	url,
+	coordinate,
+	epsg,
+	signal,
+}: {
```
dopenguin, line 44: Why the change to an object parameter here and in `reverseGeocodeWps`? I doubt the API will change that much in the future.

src/plugins/reverseGeocoder/utils/reverseGeocodeNominatim.ts
```diff
@@ -0,0 +1,93 @@
...
+	) as [number, number]
+
+	const fetchUrl = new URL(url)
+	fetchUrl.searchParams.set('lat', searchCoordinate[1].toString())
+	fetchUrl.searchParams.set('lon', searchCoordinate[0].toString())
+	fetchUrl.searchParams.set('format', 'jsonv2')
+
+	const result = await fetch(fetchUrl, { signal }).then(
+		(response) => <Promise<NominatimReverseGeocodeResponse>>response.json()
+	)
```
dopenguin:
```suggestion
	const result: NominatimReverseGeocodeResponse = await fetch(fetchUrl, {
		signal,
	}).then((response) => response.json())
```
Seems more intuitive.

dopenguin (changes requested):
🏓 @oeninghe-dataport

src/plugins/reverseGeocoder/utils/reverseGeocodeNominatim.ts
oeninghe-dataport: According to [the docs](https://nominatim.org/release-docs/latest/api/Output/#json), that field does not exist. Not quite sure why I assumed that earlier.

src/plugins/reverseGeocoder/utils/reverseGeocodeNominatim.ts
oeninghe-dataport: The docs for this are [here](https://nominatim.org/release-docs/latest/api/Output/#addressdetails), will add them to the code, too. They state that "[...] the following labels **may** appear:", and in fact, it might happen for most that they are not set.

For our service, `country` and `country_code` seem to be always there de facto.

src/plugins/reverseGeocoder/utils/reverseGeocodeNominatim.ts
oeninghe-dataport: No, but as `reverseGeocodeWps` has now five parameters, I find it quite confusing to just list them. Also, we would need to have the unused parameter `serviceEpsg` here in the parameter list, to keep the API stable between the different services.

oeninghe-dataport:
🏓 @dopenguin

src/plugins/reverseGeocoder/types.ts
```diff
@@ -26,6 +31,15 @@ export interface ReverseGeocoderPluginOptions extends PluginOptions {
 	 */
 	coordinateSources?: StoreReference[]
 
+	/**
+	 * EPSG code of the coordinate system used by the WPS service.
+	 *
+	 * This is only relevant for the WPS service, since Nominatim always uses EPSG:4326.
```
dopenguin: Directly linking to the parameter seems like the way to go

src/plugins/reverseGeocoder/store.ts
```diff
@@ -78,11 +79,17 @@ export const useReverseGeocoderStore = defineStore(
...
+				}[configuration.value.type]
+				const feature = await reverseGeocodeUtil({
+					url: configuration.value.url,
 					coordinate,
-					signal
-				)
+					epsg: coreStore.configuration.epsg,
+					serviceEpsg: configuration.value.epsg || 'EPSG:25832',
+					signal,
+				})
```
dopenguin, line 95: This does not properly highlight the fact that `serviceEpsg` is required in one implementation and not an allowed parameter in the other. A verbose implementation seems favourable

src/plugins/reverseGeocoder/utils/reverseGeocodeWps.ts
```diff
@@ -30,14 +32,24 @@ function getTextContent(parent: Element, localName: string) {
...
+	epsg: string
+	serviceEpsg: string
 	signal: AbortSignal
-): Promise<ReverseGeocoderFeature> {
+}): Promise<ReverseGeocoderFeature> {
 	const response = await fetch(url, {
 		method: 'POST',
-		body: buildPostBody(coordinate),
+		body: buildPostBody(
+			transformCoordinate(coordinate, epsg, serviceEpsg) as [number, number]
```
dopenguin: `buildPostBody` is a local function. Changing the parameter type to `Coordinate` is preferable over a type-cast.

dopenguin (changes requested):
🏓 @oeninghe-dataport 

Also, do we have an update on the following from https://github.com/Dataport/polar/pull/606#pullrequestreview-4430194996?

> Is it possible to update the POLAR nominatim service to include Hamburg as the snowbox typically uses a lot of Hamburg specific services?

src/plugins/reverseGeocoder/utils/reverseGeocodeWps.ts
dopenguin: annoyed

🎩

src/plugins/reverseGeocoder/utils/reverseGeocodeNominatim.ts
dopenguin: Probably because we can [request GeoJSON](https://nominatim.org/release-docs/latest/api/Output/#geojson). Have you tried that before?

src/plugins/reverseGeocoder/utils/reverseGeocodeNominatim.ts
dopenguin: Man, I hate it, but can't be changed.

However, the response might change if https://github.com/Dataport/polar/pull/606#discussion_r3616236440 comes through.

src/plugins/reverseGeocoder/utils/reverseGeocodeNominatim.ts
dopenguin: Regarding the amount of parameter: Sure, that seems reasonable.
Regarding the parameter list: https://github.com/Dataport/polar/pull/606#discussion_r3616183259

🎩

src/plugins/reverseGeocoder/utils/reverseGeocodeNominatim.ts
oeninghe-dataport: Yes, I tried that earlier. Both work flawlessly, however, the JSONv2 structure is IMO easier and I preferred it therefore. A Nominatim-specific implementation is necessary either way.

src/plugins/reverseGeocoder/types.ts
oeninghe-dataport: Which parameter should be linked here? I'm confused

src/plugins/reverseGeocoder/store.ts
oeninghe-dataport: Your comment reads as if you would prefer sth. like
```js
let feature
switch(configuration.value.type) {
    case 'wps':
         feature = ...
    ...
}
```

I don't think this would improve readability, but in contrast, it would be harder to spot what the code is about as you have to read it n times and look for the differences.

Therefore, I propose ca44a6b1c8 as a compromise

src/plugins/reverseGeocoder/utils/reverseGeocodeWps.ts
oeninghe-dataport: f250d7f8d7

oeninghe-dataport:
> 🏓 @oeninghe-dataport
> 
> Also, do we have an update on the following from [#606 (review)](https://github.com/Dataport/polar/pull/606#pullrequestreview-4430194996)?
> 
> > Is it possible to update the POLAR nominatim service to include Hamburg as the snowbox typically uses a lot of Hamburg specific services?

🏓 @dopenguin 

There is no reply on this yet, I'll put it on the agenda of our weekly.

src/plugins/reverseGeocoder/utils/reverseGeocodeNominatim.ts
dopenguin: The response may be slightly easier to parse, but information of the feature is lost. It seems preferrable to use `'geojson'` as the response format in order to easily add the `properties` here.

src/plugins/reverseGeocoder/types.ts
dopenguin: `ReverseGeocoderPluginOptions.type`

src/plugins/reverseGeocoder/store.ts
dopenguin: Nope, I'd strive for a solution similar to how `addressSearch` handles this.

[verbose.patch](https://github.com/user-attachments/files/30540053/verbose.patch)

dopenguin:
> > 🏓 @oeninghe-dataport
> > Also, do we have an update on the following from [#606 (review)](https://github.com/Dataport/polar/pull/606#pullrequestreview-4430194996)?
> > > Is it possible to update the POLAR nominatim service to include Hamburg as the snowbox typically uses a lot of Hamburg specific services?
> 
> 🏓 @dopenguin
> 
> There is no reply on this yet, I'll put it on the agenda of our weekly.

🏓 @oeninghe-dataport

src/plugins/reverseGeocoder/utils/reverseGeocodeNominatim.ts
oeninghe-dataport: Yea, why not. ~~e62779e808~~ de4b588302

src/plugins/reverseGeocoder/types.ts
oeninghe-dataport: Makes sense. f3dc637a12

src/plugins/reverseGeocoder/store.ts
oeninghe-dataport: a5606c2b84

oeninghe-dataport:
🏓 @dopenguin

src/plugins/reverseGeocoder/utils/reverseGeocodeNominatim.ts
dopenguin: The type cast on `const properties = feature.properties as NominatimReverseGeocodeProperties` is redundant as `const result: FeatureCollection<Point, NominatimReverseGeocodeProperties>` would do the trick.
This also removes the need for the `feature.geometry.type !== 'Point'`-check.

dopenguin:
> 🏓 @dopenguin

🏓 @oeninghe-dataport

src/plugins/reverseGeocoder/utils/reverseGeocodeNominatim.ts
oeninghe-dataport: Thanks, good catch. 59fe89ac1

oeninghe-dataport:
> > 🏓 @dopenguin
> 
> 🏓 @oeninghe-dataport

🏓 @dopenguin

src/plugins/reverseGeocoder/utils/reverseGeocodeNominatim.ts
```diff
@@ -0,0 +1,101 @@
...
+		await fetch(fetchUrl, {
+			signal,
+		}).then((response) => response.json())
+
+	const feature = result.features[0]
+	if (!feature) {
+		throw new Error('No features returned from Nominatim reverse geocode')
+	}
+
+	const properties = feature.properties
```
dopenguin:
```suggestion
	const { properties } = feature
```

src/plugins/reverseGeocoder/utils/reverseGeocodeNominatim.ts
```diff
@@ -58,23 +56,33 @@ export async function reverseGeocodeNominatim({
...
-	fetchUrl.searchParams.set('format', 'jsonv2')
+	fetchUrl.searchParams.set('format', 'geojson')
 
-	const result = await fetch(fetchUrl, { signal }).then(
-		(response) => <Promise<NominatimReverseGeocodeResponse>>response.json()
-	)
+	const result: FeatureCollection<Point, NominatimReverseGeocodeProperties> =
+		await fetch(fetchUrl, {
+			signal,
+		}).then((response) => response.json())
```
dopenguin: The linter allows this to be written in one line

dopenguin:
> > > 🏓 @dopenguin
> > 
> > 
> > 🏓 @oeninghe-dataport
> 
> 🏓 @dopenguin

🏓 @oeninghe-dataport

src/plugins/reverseGeocoder/utils/reverseGeocodeNominatim.ts
oeninghe-dataport: That's true, and I like it and I changed it, although I'm surprised that prettier offers a choice here. I assumed prettier would be more opinionated.

1d1bf1ddc2

oeninghe-dataport:
> > > > 🏓 @dopenguin
> > > 
> > > 
> > > 🏓 @oeninghe-dataport
> > 
> > 
> > 🏓 @dopenguin
> 
> 🏓 @oeninghe-dataport

🏓 @dopenguin

dopenguin (approved):
🚀 @oeninghe-dataport

PR #652, opened 2026-04-14
fix: update multiple packages with vulnerable versions

dopenguin:
## Summary

Updated multiple packages that had vulnerable versions.

Installing and starting a client should work. Only `nx` had a major jump, so the other changes should be negligible.

  - [x] Chrome Lighthouse
  - [x] Firefox Accessibility


warm-coolguy (changes requested):
:shipit: @dopenguin

warm-coolguy (approved):
Oh wait, I hit the wrong button.

:shipit: @dopenguin

PR #660, opened 2026-04-17
Fix/audit fix

warm-coolguy:
## Summary

Audit fix and introducing `min-release-age`.

Only patches. Might check clients. I briefly looked and they seem to be as broken as expected.


oeninghe-dataport (changes requested):
Please update the Node version also in the corresponding GitHub workflow configurations.

oeninghe-dataport (approved):

dopenguin:
> Please update the Node version also in the corresponding GitHub workflow configurations.

See c4100697afb23c26a217d694dead799fdaad7841

🏓 @oeninghe-dataport

PR #667, opened 2026-04-20
Chore/update vite related packages

dopenguin:
## Summary

Update `vite`, `vitest` and related packages.
It does not seem to be an issue that `vite` now uses `rolldown`.

Note that there are some peer dependency issue with `vite-plugin-vue-devtools` which don't seem to be an issue for us.
Those [are being tackled](https://github.com/antfu-collective/vite-plugin-inspect/issues/173) and we should be able to update `vite-plugin-vue-devtools` later on

Closes #643, #661, #665


oeninghe-dataport (approved):
🚀 @dopenguin

PR #694, opened 2026-04-30
ci: upgrade speed

dopenguin:
## Summary

Enhances the speed of concurrent runs by caching the `node_modules` and eslint results. Pushing new commits also aborts ongoing CI runs on the branch.
Also upgrades the github actions to the latest versions.

Also, dependabot should update the github actions here as well.

## Instructions for local reproduction and review

Check the speed difference between [before](https://github.com/Dataport/polar/actions/runs/25166718715/job/73774806535) and [after](https://github.com/Dataport/polar/actions/runs/25167063434/job/73776027959).
Note that the initial commit will always take a few minutes to establish the cache.

## Relevant tickets, issues, et cetera

#695 adds the dependabot update to `main`

PR #695, opened 2026-04-30
chore: update dependabot configuration

dopenguin:
## Summary

Updates the dependabot configuration to also update github actions.

Take a look at the pipeline (which currently breaks because [diplanung-style](https://www.npmjs.com/package/diplanung-style) is gone from npm; it can somehow still be found locally)

#694 has the same changes


warm-coolguy (approved):
:shipit: @dopenguin

PR #718, opened 2026-05-12
Refactor layout

warm-coolguy:
## Summary

Removed margins from plugins to let layouting handle it.

Look at code. Look at client.


dopenguin (approved):
:shipit: @warm-coolguy 

Image

PR #733, opened 2026-05-19
refactor: update kern components to use slots in favour of props

dopenguin:
## Summary

Make the changes of the kern components feel more natural and reduce bloat by removing unnecessary variables.

As only the filter plugins currently uses these components, run `snowbox` and check whether it still looks fine.

### DIFF
--- a/src/components/kern/KernBlockButton.ce.vue
+++ b/src/components/kern/KernBlockButton.ce.vue
@@ -1,28 +1,22 @@
 <template>
 	<button
 		class="kern-btn kern-btn--block kern-btn--tertiary"
-		@click="emit('click')"
+		@click="$emit('click')"
 	>
 		<span
-			v-if="props.icon"
-			:class="{ 'kern-icon': true, [props.icon]: true }"
+			v-if="$props.icon"
+			:class="`kern-icon ${$props.icon}`"
 			aria-hidden="true"
 		/>
-		<span class="kern-label">{{ props.label }}</span>
+		<span class="kern-label"><slot /></span>
 	</button>
 </template>
 
 <script setup lang="ts">
 import type { Icon } from '@/core'
 
-const props = defineProps<{
-	icon?: Icon | false
-	label: string
-}>()
-
-const emit = defineEmits<{
-	click: []
-}>()
+defineProps<{ icon?: Icon }>()
+defineEmits<{ click: [] }>()
 </script>
 
 <style scoped>
--- a/src/components/kern/KernBlockButton.ce.vue
+++ b/src/components/kern/KernBlockButton.ce.vue
@@ -3,7 +3,7 @@
 		class="kern-btn kern-btn--block kern-btn--tertiary"
 		@click="$emit('click')"
 	>
-		<span v-if="icon" :class="`kern-icon ${icon}`" aria-hidden="true" />
+		<span v-if="icon" class="kern-icon" :class="icon" aria-hidden="true" />
 		<span class="kern-label"><slot /></span>
 	</button>
 </template>
--- a/src/components/kern/KernBlockButton.ce.vue
+++ b/src/components/kern/KernBlockButton.ce.vue
@@ -3,11 +3,7 @@
 		class="kern-btn kern-btn--block kern-btn--tertiary"
 		@click="$emit('click')"
 	>
-		<span
-			v-if="$props.icon"
-			:class="`kern-icon ${$props.icon}`"
-			aria-hidden="true"
-		/>
+		<span v-if="icon" :class="`kern-icon ${icon}`" aria-hidden="true" />
 		<span class="kern-label"><slot /></span>
 	</button>
 </template>
--- a/src/components/kern/KernBlockButtonCheckbox.ce.vue
+++ b/src/components/kern/KernBlockButtonCheckbox.ce.vue
@@ -8,11 +8,11 @@
 	/>
 	<label :for="id" class="kern-btn kern-btn--block kern-btn--tertiary">
 		<span
-			v-if="props.icon"
-			:class="{ 'kern-icon': true, [props.icon]: true }"
+			v-if="$props.icon"
+			:class="`kern-icon ${$props.icon}`"
 			aria-hidden="true"
 		/>
-		<span class="kern-label">{{ props.label }}</span>
+		<span class="kern-label"><slot /></span>
 	</label>
 </template>
 
@@ -21,10 +21,7 @@ import { useId } from 'vue'
 
 import type { Icon } from '@/core'
 
-const props = defineProps<{
-	icon?: Icon | false
-	label: string
-}>()
+defineProps<{ icon?: Icon }>()
 
 const model = defineModel<boolean>({ required: true })
 
--- a/src/components/kern/KernBlockButtonCheckbox.ce.vue
+++ b/src/components/kern/KernBlockButtonCheckbox.ce.vue
@@ -7,7 +7,7 @@
 		@focus="scrollVisible($event)"
 	/>
 	<label :for="id" class="kern-btn kern-btn--block kern-btn--tertiary">
-		<span v-if="icon" :class="`kern-icon ${icon}`" aria-hidden="true" />
+		<span v-if="icon" class="kern-icon" :class="icon" aria-hidden="true" />
 		<span class="kern-label"><slot /></span>
 	</label>
 </template>
--- a/src/components/kern/KernBlockButtonCheckbox.ce.vue
+++ b/src/components/kern/KernBlockButtonCheckbox.ce.vue
@@ -7,11 +7,7 @@
 		@focus="scrollVisible($event)"
 	/>
 	<label :for="id" class="kern-btn kern-btn--block kern-btn--tertiary">
-		<span
-			v-if="$props.icon"
-			:class="`kern-icon ${$props.icon}`"
-			aria-hidden="true"
-		/>
+		<span v-if="icon" :class="`kern-icon ${icon}`" aria-hidden="true" />
 		<span class="kern-label"><slot /></span>
 	</label>
 </template>
--- a/src/components/kern/KernBlockButtonRadioGroup.ce.vue
+++ b/src/components/kern/KernBlockButtonRadioGroup.ce.vue
@@ -19,7 +19,8 @@
 			>
 				<span
 					v-if="item.icon"
-					:class="`kern-icon ${item.icon}`"
+					class="kern-icon"
+					:class="item.icon"
 					aria-hidden="true"
 				/>
 				<span class="kern-label">{{ item.label }}</span>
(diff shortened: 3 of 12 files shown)


src/components/kern/KernBlockButton.ce.vue
```diff
@@ -1,28 +1,22 @@
 <template>
 	<button
 		class="kern-btn kern-btn--block kern-btn--tertiary"
-		@click="emit('click')"
+		@click="$emit('click')"
```
oeninghe-dataport, line 4: The usage of `$...` is Options API (see https://vuejs.org/api/component-instance.html) and therefore forbidden by our (not-yet-published) ADR to use Composition API.

src/components/kern/KernBlockButton.ce.vue
```diff
@@ -1,28 +1,22 @@
...
 	>
 		<span
-			v-if="props.icon"
-			:class="{ 'kern-icon': true, [props.icon]: true }"
+			v-if="$props.icon"
+			:class="`kern-icon ${$props.icon}`"
```
oeninghe-dataport: The usage of `$...` is Options API (see https://vuejs.org/api/component-instance.html) and therefore forbidden by our (not-yet-published) ADR to use Composition API.

src/components/kern/KernBlockButtonCheckbox.ce.vue
```diff
@@ -8,11 +8,11 @@
 	/>
 	<label :for="id" class="kern-btn kern-btn--block kern-btn--tertiary">
 		<span
-			v-if="props.icon"
-			:class="{ 'kern-icon': true, [props.icon]: true }"
+			v-if="$props.icon"
+			:class="`kern-icon ${$props.icon}`"
```
oeninghe-dataport: The usage of `$...` is Options API (see https://vuejs.org/api/component-instance.html) and therefore forbidden by our (not-yet-published) ADR to use Composition API.

src/components/kern/KernBlockButtonRadioGroup.ce.vue
```diff
@@ -1,6 +1,6 @@
 <template>
 	<div class="radiogroup" role="radiogroup">
-		<template v-for="(item, idx) of props.items" :key="idx">
+		<template v-for="(item, idx) of $props.items" :key="idx">
```
oeninghe-dataport: The usage of `$...` is Options API (see https://vuejs.org/api/component-instance.html) and therefore forbidden by our (not-yet-published) ADR to use Composition API.

src/components/kern/KernDatePicker.ce.vue
```diff
@@ -3,8 +3,8 @@
 		v-model="inputModel"
 		type="date"
 		class="kern-form-input__input"
-		:min="dateToString(props.min ?? null)"
-		:max="dateToString(props.max ?? null)"
+		:min="dateToString($props.min ?? null)"
+		:max="dateToString($props.max ?? null)"
```
oeninghe-dataport: The usage of `$...` is Options API (see https://vuejs.org/api/component-instance.html) and therefore forbidden by our (not-yet-published) ADR to use Composition API.

src/components/kern/KernDateRangePicker.ce.vue
```diff
@@ -1,15 +1,15 @@
 <template>
 	<div class="range-picker">
-		<KernDatePicker v-model="start" v-bind="props" />
+		<KernDatePicker v-model="start" v-bind="$props" />
 		–
-		<KernDatePicker v-model="end" v-bind="props" />
+		<KernDatePicker v-model="end" v-bind="$props" />
```
oeninghe-dataport: The usage of `$...` is Options API (see https://vuejs.org/api/component-instance.html) and therefore forbidden by our (not-yet-published) ADR to use Composition API.

src/components/kern/KernRadioGroup.ce.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-	<div v-for="(item, idx) of props.items" :key="idx" class="kern-form-check">
+	<div v-for="(item, idx) of $props.items" :key="idx" class="kern-form-check">
```
oeninghe-dataport: The usage of `$...` is Options API (see https://vuejs.org/api/component-instance.html) and therefore forbidden by our (not-yet-published) ADR to use Composition API.

src/components/kern/KernBlockButton.ce.vue
oeninghe-dataport:
```suggestion
			:class="['kern-icon', props.icon]"
```

I would prefer object- or array-style over string concatenation, as it is clearer and possibly safer.

src/components/kern/KernBlockButtonCheckbox.ce.vue
oeninghe-dataport:
```suggestion
			:class="['kern-icon', props.icon]"
```

src/components/kern/KernBlockButtonRadioGroup.ce.vue
```diff
@@ -19,7 +19,7 @@
 			>
 				<span
 					v-if="item.icon"
-					:class="{ 'kern-icon': true, [item.icon]: true }"
+					:class="`kern-icon ${item.icon}`"
```
oeninghe-dataport:
```suggestion
					:class="['kern-icon', item.icon]"
```

oeninghe-dataport (changes requested):
🏓 @dopenguin 

Probably we should introduce an ESLint rule to prohibit `$...`. I'll add this to the agenda of our weekly.

src/components/kern/KernBlockButton.ce.vue
dopenguin: Nope https://vuejs.org/guide/components/events.html#declaring-emitted-events

src/components/kern/KernBlockButton.ce.vue
dopenguin: Found something even easier 4e53aa866eb05e1daeea1003410e726d740ca3c6

src/components/kern/KernBlockButtonCheckbox.ce.vue
dopenguin: Found something even easier 4e53aa866eb05e1daeea1003410e726d740ca3c6

src/components/kern/KernBlockButtonRadioGroup.ce.vue
dopenguin: Found something even easier 4e53aa866eb05e1daeea1003410e726d740ca3c6

src/components/kern/KernDatePicker.ce.vue
dopenguin: Found something even easier 4e53aa866eb05e1daeea1003410e726d740ca3c6

src/components/kern/KernDateRangePicker.ce.vue
dopenguin: Found something even easier 4e53aa866eb05e1daeea1003410e726d740ca3c6

src/components/kern/KernRadioGroup.ce.vue
dopenguin: Found something even easier 4e53aa866eb05e1daeea1003410e726d740ca3c6

src/components/kern/KernBlockButton.ce.vue
dopenguin: How would you ensure this is always done?

src/components/kern/KernBlockButtonCheckbox.ce.vue
dopenguin: https://github.com/Dataport/polar/pull/733#discussion_r3273333870

src/components/kern/KernBlockButtonRadioGroup.ce.vue
dopenguin: https://github.com/Dataport/polar/pull/733#discussion_r3273333870

dopenguin:
> 🏓 @dopenguin
> 
> Probably we should introduce an ESLint rule to prohibit `$...`. I'll add this to the agenda of our weekly.

🏓 @oeninghe-dataport

src/components/kern/KernBlockButton.ce.vue
oeninghe-dataport:
```javascript
		'vue/no-restricted-syntax': [
			'error',
			{
				selector:
					'VAttribute[directive=true][key.argument.name="style"] > VExpressionContainer > TemplateLiteral',
				message:
					'Template literals in :style are not allowed. Use Array syntax (:style="[a, b]") or Object syntax (:style="{prop: value}") instead.',
			},
		],
```

src/components/kern/KernBlockButton.ce.vue
oeninghe-dataport: I don't like this change, however, it seems you are right that this is allowed.

src/components/kern/KernBlockButton.ce.vue
dopenguin: I don't think we should be tightening the rules on this. @warm-coolguy, what is your take?

warm-coolguy: 1f597a67-cab6-4aeb-b034-412c6ff28c3d_1456x1106

>possibly safer

Please elaborate.

oeninghe-dataport: 1. The object- and array-style binding are the only bindings mentioned in the [documentation](https://vuejs.org/guide/essentials/class-and-style.html#binding-html-classes). String binding is of course possible, but undocumented.
2. If a value is falsy, but not an empty string, it may be stringified to `false`, `0`, `null` or `undefined` (or possibly other options). When using the object- or array-style binding, these values are filtered out.
3. Not an argument for "safer", but it is - at least in this case, one character shorter.

warm-coolguy: 1. "[But Vue actually supports the full power of JavaScript expressions inside all data bindings](https://vuejs.org/guide/essentials/template-syntax.html#using-javascript-expressions)"
2. I play the reverse card and this is an argument for using template string syntax in class bindings to avoid filtering that out since it's easier to debug if the classname turns out to be `undefined`. Proposal for a global rule: `.false, .0, .null, .undefined { border: 20px solid red; }`
3. That almost got me.

I don't think we need to enforce this. But I do see it's advantegeous in some situations regarding readability. I'd like to keep the flexibility to do either, depending on ~~how I feel on that particular day~~ which one suits the case better.

In the upper case, I figure the array notation is more fitting. Also, I don't care.

dopenguin: As discussed, I've simplified the class usage c521dc0209e9fc30d2abc6b0ba7978a1609cddd5

src/components/kern/KernBlockButtonRadioGroup.ce.vue
dopenguin: c521dc0209e9fc30d2abc6b0ba7978a1609cddd5

dopenguin:
🏓 @oeninghe-dataport

oeninghe-dataport (approved):
🚀 @dopenguin

PR #734, opened 2026-05-21
refactor: kern btn

dopenguin:
## Summary

Replaces instances of `<button class="kern-btn">` with the reusable component.

Anything touched by this change should still behave as before.

Blocked by #733, #805


src/components/kern/KernButton.ce.vue
```diff
@@ -0,0 +1,15 @@
+<template>
+	<button class="kern-btn" @click="$emit('click')">
+		<span v-if="icon" class="kern-icon" :class="icon" aria-hidden="true" />
+		<span class="kern-label" :class="{ 'kern-sr-only': labelSrOnly }">
```
oeninghe-dataport, line 4: I'd prefer a solution where `kern-sr-only` is automatically calculated based on if the slot is empty or not.

dopenguin: The button is also used in cases where you want the label to be visible, so this wouldn't work.

oeninghe-dataport: Em, yeah, you're right, nvm

src/components/kern/KernBlockButton.ce.vue
oeninghe-dataport, line 1: I would prefer to keep `KernBlockButton` as a separate component that uses `KernButton` and holds its CSS. The component would be quite small, however, it seems more consistent to me as we also have `KernBlockButtonCheckbox` and `KernBlockButtonRadioGroup`.

src/components/kern/KernBlockButtonCheckbox.ce.vue
```diff
@@ -7,12 +7,8 @@
 		@focus="scrollVisible($event)"
 	/>
 	<label :for="id" class="kern-btn kern-btn--block kern-btn--tertiary">
-		<span
-			v-if="props.icon"
-			:class="{ 'kern-icon': true, [props.icon]: true }"
-			aria-hidden="true"
-		/>
-		<span class="kern-label">{{ props.label }}</span>
+		<span v-if="icon" :class="`kern-icon ${icon}`" aria-hidden="true" />
```
oeninghe-dataport:
```suggestion
		<span v-if="icon" class="kern-icon" :class="icon" aria-hidden="true" />
```

src/components/kern/KernBlockButtonRadioGroup.ce.vue
```diff
@@ -19,7 +19,7 @@
 			>
 				<span
 					v-if="item.icon"
-					:class="{ 'kern-icon': true, [item.icon]: true }"
+					:class="`kern-icon ${item.icon}`"
```
oeninghe-dataport:
```suggestion
					class="kern-icon"
					:class="item.icon"
```

src/components/kern/KernButton.ce.vue
```diff
@@ -0,0 +1,15 @@
...
+			<slot />
+		</span>
+	</button>
+</template>
+
+<script setup lang="ts">
+import type { Icon } from '@/core'
+
+defineProps<{ icon?: Icon; labelSrOnly?: boolean }>()
```
oeninghe-dataport, line 13: What do you think about adding `variant: 'primary' | 'secondary' | 'tertiary'` here?

src/plugins/layerChooser/components/LegendButton.ce.vue
```diff
@@ -1,21 +1,20 @@
...
+	</KernButton>
 </template>
 
 <script setup lang="ts">
+import KernButton from '@/components/kern/KernButton.ce.vue'
+
 import { PluginId } from '../types'
 
 defineProps<{
-	disabled: boolean
```
oeninghe-dataport, line 18: This is still in use, why removing it?

oeninghe-dataport (dismissed):
🏓 @dopenguin

src/components/kern/KernBlockButton.ce.vue
dopenguin: I don't really see the benefit of keeping that component as it is only being used in the Filter plugin.
I also intend to replace `KernBlockButtonCheckbox` and `KernBlockButtonRadioGroup` with `PolarInput` in similar fashion as they overlap quite a lot.

src/components/kern/KernBlockButtonCheckbox.ce.vue
dopenguin: Outdated, already updated after the merge with 0296631f28d33258bc329e9676c09dc9177d619e

src/components/kern/KernBlockButtonRadioGroup.ce.vue
dopenguin: Outdated, already updated after the merge with 0296631f28d33258bc329e9676c09dc9177d619e

src/components/kern/KernButton.ce.vue
dopenguin: The benefit seems negligible as one would basically interchange `class="kern-btn--tertiary"` with `variant="tertiary"`.
I prefer the current solution, especially if I'd be adding another class like in `src/plugins/addressSearch/components/AddressSearch.ce.vue`.

src/plugins/layerChooser/components/LegendButton.ce.vue
dopenguin: There is no need to expose the parameter as a prop as the parameter is set directly on the button element.

dopenguin:
> 🏓 @dopenguin

🏓 @oeninghe-dataport

src/plugins/layerChooser/components/LegendButton.ce.vue
oeninghe-dataport: However, naming them explicitly allows for code completion and type checking, which we both lose when removing it. Fallthrough attributes should be the exception IMO, maybe acceptable for `class` and `style` however.

src/components/kern/KernButton.ce.vue
oeninghe-dataport: I prefer the `variant` solution as it provides type-safety. Also, I like that we separate the KERN and the custom classes there.

src/components/kern/KernBlockButton.ce.vue
oeninghe-dataport: I think they behave quite differently

oeninghe-dataport:
> > 🏓 @dopenguin
> 
> 🏓 @oeninghe-dataport

🏓 @dopenguin

src/plugins/layerChooser/components/LegendButton.ce.vue
dopenguin: There are [quite a lot of attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/button#attributes) that can be set on the button that just pass through, which is a good thing!
Being able to use this component just like one would use a normal button eases the usage without cluttering the component here.
Also, fallthrough attributes are [a deliberate Vue 3 design choice](https://vuejs.org/guide/components/attrs.html).

src/components/kern/KernBlockButton.ce.vue
dopenguin: In their current implementation, yes, the behaviour of `KernBlockButtonCheckbox` and `KernBlockButtonRadioGroup` differ from `PolarInput`.

As those components have not been changed in this PR, can we move forward without further changes here?

src/components/kern/KernButton.ce.vue
dopenguin: A `variant` prop would internally just map to something like `:class="'kern-btn--' + variant"`, basically translating one string into another. That's an abstraction that doesn't carry its weight. Someone using `KernButton` already knows which class they want. The prop just adds a detour.
Additionally, with classes, a new visual treatment can be applied without ever touching `KernButton` itself. A `variant` prop would require modifying the component for every new variant (if KERN decides to add more) which goes against the Open/Closed principle. For a slim, reusable wrapper this matters.

dopenguin:
> > > 🏓 @dopenguin
> > 
> > 
> > 🏓 @oeninghe-dataport
> 
> 🏓 @dopenguin

🏓 @oeninghe-dataport

src/components/kern/KernBlockButton.ce.vue
oeninghe-dataport: No, I still don't like the idea of losing `KernBlockButton` while `KernBlockButtonCheckbox` and `KernBlockButtonRadioGroup` are generic components. We may discuss the removal if we also remove the other two components, but in the current situation, I'd strongly prefer having a wrapper here.

oeninghe-dataport:
> > > > 🏓 @dopenguin
> > > 
> > > 
> > > 🏓 @oeninghe-dataport
> > 
> > 
> > 🏓 @dopenguin
> 
> 🏓 @oeninghe-dataport

🏓 @dopenguin

src/components/kern/KernBlockButton.ce.vue
dopenguin: Looking at the current usage, `KernBlockButtonCheckbox` and `KernBlockButtonRadioGroup` are also, like `KernBlockButton` was, used exactly once, both in the`filter` plugin. None of the three are currently generically used despite living in the kern directory; they're effectively `filter`-specific components.
Consolidating them into `PolarInput` would provide a genuinely generic, reusable component with proper input semantics rather than maintaining these two single-use components that share a naming convention but not much else.
I'd prefer not to block this PR on keeping a wrapper for a naming pattern between three components that are all single-use and all candidates for replacement (see #779). Adding the replacement with `PolarInput` here would unnecessarily bloat this PR.

dopenguin:
> > > > > 🏓 @dopenguin
> > > > 
> > > > 
> > > > 🏓 @oeninghe-dataport
> > > 
> > > 
> > > 🏓 @dopenguin
> > 
> > 
> > 🏓 @oeninghe-dataport
> 
> 🏓 @dopenguin

🏓 @oeninghe-dataport

src/components/kern/KernBlockButton.ce.vue
warm-coolguy: Plus for the change underway and `KernBlockButtonCheckbox` and `KernBlockButtonRadioGroup` are functionally indistinguishable trash anyway.

dopenguin:
The [current type issue](https://github.com/Dataport/polar/actions/runs/27551027323/job/81437151028?pr=734) is a regression bug in [`@vue/language-core`](https://github.com/vuejs/language-tools/issues/6096), which is [already fixed](https://github.com/vuejs/language-tools/pull/6097) and [released](https://github.com/vuejs/language-tools/pull/6102).
We can use the new [`vue-tsc`](https://www.npmjs.com/package/vue-tsc) version tomorrow which should resolve this without further ado.

**Edit**: Fixed now by #805

oeninghe-dataport (commented):
🏓 @warm-coolguy takes review

warm-coolguy (approved):
:shipit: @dopenguin

PR #745, opened 2026-05-26
feat: add context menu

### LINKED ISSUE #403: Integration of ContextMenu as part of the core

dopenguin:
## Summary

Adds the possibility for plugins to insert themselves in a context menu by calling `addToContextMenu`. Entries can be removed by calling `removeFromContextMenu`.

- Start the `snowbox`
- On Desktop: Right click on the map canvas; on mobile: longpress the map canvas
- The context menu will be displayed with 4 entries; real entries will follow once draw or routing are implemented

## Known issues

On mobile, a pin is still being set as OpenLayers already fires a `singleclick`-event before the 1000ms of the `longPressHammer` have elapsed.

Resolves: #403
Blocked by: #733, #734, #748


warm-coolguy:
dibs

src/components/kern/KernButton.ce.vue
```diff
@@ -13,3 +13,43 @@ import type { Icon } from '@/core'
...
+		);
+	}
+	.kern-label {
+		color: var(
+			--kern-btn-text-color,
+			var(--kern-color-action-on-default-contextual)
+		);
+	}
+}
+</style>
```
warm-coolguy, line 47: I didn't figure out how this is helpful. Upon removing the section, there's a hover effect on the context menu buttons again. With this block on, any hover effect on the button itself is completely amiss.

Was that intentional? If so, I figure this'd rather belong to the ContextMenu as an override.

Anyway, I'd like to have a hover effect of some kind, as all other interactive elements do.

src/core/components/ContextMenu.ce.vue
```diff
@@ -0,0 +1,134 @@
...
+const menu = useTemplateRef<HTMLElement>('menu')
+onMounted(() => {
+	;(menu.value as HTMLElement)
+		.querySelector<HTMLElement>('[role="menuitem"]')
+		?.focus()
+})
+
+const menuItems = useTemplateRef<ComponentPublicInstance[]>('menuItems')
+
+function focusNextElement(event: KeyboardEvent, direction: -1 | 1) {
```
warm-coolguy, line 84: With this, I can go down and down forever, but only up to the first element. I find this to be confusing as there seems to be no special reason to stop going circles here. On all other context menus I tried, circles go in both directions.

Edit: There's another thread for this in a more fitting spot.

src/core/components/ContextMenu.ce.vue
```diff
@@ -0,0 +1,134 @@
...
+	.kern-card__body {
+		gap: 0;
+
+		button {
+			justify-content: start;
+
+			&:focus {
+				outline: auto;
+			}
+		}
```
warm-coolguy: We shouldn't initially display the focus on right-click, as it's confusing to some users. I've checked what Windows does and what other component libraries do, and they either have an "invisible initial" element at first that holds the focus so that pressing down goes to the first one, or have the first element selected, but not visibly so.

The latter would be the difference between `:focus` and `:focus-visible`, which I feel is preferable to the current state and easy to implement.

src/core/components/ContextMenu.ce.vue
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	.kern-card__body {
+		gap: 0;
+
+		button {
+			justify-content: start;
+
+			&:focus {
+				outline: auto;
```
warm-coolguy: The outline does not match the button style.

Image

With increased font size, you can easily see there's a small gap there in the top-left corner. It's also visible without increased font size, and the differing border-radius is also discernible anyway. Just posting this since it makes it very visible.

While we're here, the outline does use a different style than e.g. the icon menu. I'd prefer it if it used the same effects. Example for colour:

Image

src/core/components/PolarContainer.ce.vue
```diff
@@ -207,11 +256,19 @@ onMounted(() => {
 	// FIXME: Improve types for lightElement
 	// This is necessary for making `getStore` work
 	;(mainStore.lightElement as { store?: unknown }).store = useCoreStore()
+
+	mainStore.map
+		.getTargetElement()
+		.addEventListener('contextmenu', openContextMenu)
```
warm-coolguy, line 247: Here and Line `findLineWith("// Suppresses the context menu of the browser")`:

This effectively overrides the possibility to "normally" open the native context menus on a canvas, that is, however, pretty useful.

Image

I've checked and pressing Shift+RightClick on Firefox circumvents this problem: https://developer.mozilla.org/en-US/docs/Web/API/Element/contextmenu_event (Somewhere down there in the initial chapter.)

I figure that this same behaviour could be used in all browsers with little implementation effort so that users can use the native context menu anyway. A small hint at the bottom of the context menu would be helpful, too. What do you think?

src/core/components/PolarContainer.ce.vue
```diff
@@ -97,33 +104,53 @@ function wheelEffect(event: WheelEvent) {
...
-				overlay.value.show(oneFingerPan)
-			}
+	longPressHammer?.destroy()
+	longPressHammer = null
+	panHammer?.destroy()
+	panHammer = null
+
+	const container = polarMapContainer.value?.el
+	if (container && hasSmallDisplay.value) {
+		longPressHammer = new Hammer(container, { time: 1000 }).on('press', (e) => {
```
warm-coolguy, line 117: > On mobile, a pin is still being set as OpenLayers already fires a singleclick-event before the 1000ms of the longPressHammer have elapsed.

"You're going to fix that, right? ... Right?"

It's not exactly usable like this.

I understand it's a bigger issue. We need to understand whether something is a longpress in `singleclick` positions as well to prevent the `singleclick` effect to happen. Since this is relevant in both core/setupMarkers and pins, a centralized solution seems to make sense.

src/core/components/ContextMenu.ce.vue
```diff
@@ -0,0 +1,134 @@
...
+				<KernButton
+					v-for="{ id, icon, text, textNs, callback, color } in group"
+					:key="id"
+					ref="menuItems"
+					class="kern-btn--block kern-btn--tertiary"
+					:icon="icon"
+					role="menuitem"
+					tabindex="-1"
+					:style="{ '--kern-btn-text-color': color }"
+					@click="ring(callback)"
```
warm-coolguy, line 21: The context menu buttons simply don't work on mobile devices. I can open the context menu, but the button doesn't copy the coordinates, and the toast doesn't appear.

src/core/components/ContextMenu.ce.vue
```diff
@@ -0,0 +1,134 @@
...
+
+function focusNextElement(event: KeyboardEvent, direction: -1 | 1) {
+	const { target } = event
+	const items = menu.value?.querySelectorAll<HTMLElement>('[role="menuitem"]')
+	if (!items || target === null) {
+		console.warn('Could not focus any element.')
+		return
+	}
+	const index = [...items].indexOf(target as HTMLElement)
+	const nextElement = items[(index + direction) % items.length]
```
warm-coolguy: With this, I can go down and down forever, but only up to the first element due to `-1 % 3 === -1`. I find this to be confusing as there seems to be no special reason to stop going circles here. On all other context menus I tried, circles go in both directions.

src/core/components/ContextMenu.ce.vue
```diff
@@ -0,0 +1,134 @@
...
+	}
+}
+</script>
+
+<style scoped>
+.kern-card {
+	position: absolute;
+	border-width: 0;
+	min-width: 15rem;
+	z-index: 3;
```
warm-coolguy, line 108: Since we're having about 20 z-indexes strewn around by now, this becomes increasingly hard to check.

I propose we hold a central register for z-indexes in the core, where they reside as semantic variables, i.e.

```css
--z-index-ui-background: -10;
--z-index-ui-foreground: 10;
--z-index-context-menu: 20;
--z-index-toasts: 30;
--z-index-overlay: 40;
--z-index-tooltip: 50;
```

With these, I mean less the concrete plugins we have, and more the roles they fulfill. This way, if we have another toast implementation, it can just use the same z-index from core.

src/core/components/ContextMenu.ce.vue
```diff
@@ -0,0 +1,134 @@
...
+		button {
+			justify-content: start;
+
+			&:focus {
+				outline: auto;
+			}
+		}
+
+		hr {
+			height: var(--kern-metric-border-width-light, 0.0625rem);
```
warm-coolguy: IIRC we've removed default value copies previously. Proposal: We keep this removal up.

src/core/components/ContextMenu.ce.vue
```diff
@@ -0,0 +1,134 @@
...
+	pointer-events: all;
+
+	&:deep(.kern-card__container) {
+		padding: 0;
+	}
+
+	.kern-card__body {
+		gap: 0;
+
+		button {
```
warm-coolguy, line 119: After intense longpress on Safari Mobile, I commonly have the button's icon selected as text, which is rather confusing. I tried to prevent it with `user-select: none`, but I guess I put it in the wrong spot. If there's a CSS solution, it probably belongs around here.

src/core/stores/contextMenu.ts
```diff
@@ -0,0 +1,42 @@
...
+		buttons.value.set(entry.id, entry)
+	}
+
+	function removeEntry(id: string) {
+		buttons.value.delete(id)
+	}
+
+	return {
+		/** @internal */
+		buttons,
```
warm-coolguy, line 64: Since this has explicit "setter" methods (`addEntry`, `removeEntry`), it seems counter-intuitive to completely expose the `Map` as-is. Proposal: Maybe we should provide a "getter" in style `buttons.values().toArray()` (as it is consumed like this anyway).

src/core/components/PolarContainer.ce.vue
```diff
@@ -97,33 +104,53 @@ function wheelEffect(event: WheelEvent) {
...
+			contextMenuStore.show = true
+			const rect = (polarWrapper.value as Element).getBoundingClientRect()
+			const left = e.center.x - rect.left
+			const top = e.center.y - rect.top
+			contextMenuStore.clickCoordinate = mainStore.map.getCoordinateFromPixel([
```
warm-coolguy, line 122: I think this functionality (complete with registration and deregistration) belongs to the context menu store and its setup/teardown functionality. Regarding the ref issue, `map.store.map.getTarget().parentElement.parentElement` seems to return the right element, but isn't elegant and smells like problems down the line. Can we add the ref to the core store before any plugins initiate?

src/plugins/filter/components/FilterUI.ce.vue
```diff
@@ -46,3 +46,4 @@ const filterStore = useFilterStore()
 	width: 100%;
 }
 </style>
+f
```
warm-coolguy: Respects status: Paid.

src/plugins/pointerPosition/store.ts
```diff
@@ -58,25 +62,49 @@ export const usePointerPositionStore = defineStore(
...
+				transform(coordinate, mapProjection, selectedProjection.value)
+			)
+		}
+
 		const updatePointerPosition = ({ coordinate }) =>
 			(pointerPosition.value = coordinate)
 
 		function setupPlugin() {
 			coreStore.map.on('pointermove', updatePointerPosition)
+			coreStore.addToContextMenu({
```
warm-coolguy, line 90: Do we need configurable context menus? Letting each plugin just do it directly is a solution, albeit not a very dynamic one. I'm also thinking about initialization order and how this affects the view. 🤔 

However, this is a complex topic. My first idea was that core, after initialization of all plugins, calls `getMenuEntries` on all (in a new config.contextMenu) configured plugins, which is a new optional API endpoint ... but then the context menu would be a little static, since especially tools like draw will want to add and remove entries depending on their context. Bleh. Maybe simple ordering the groups in global config is enough.

What do you think about this? Is it a topic we should discuss now, or postpone to later?

src/plugins/pins/store.ts
```diff
@@ -165,6 +165,9 @@ export const usePinsStore = defineStore('plugins/pins', () => {
 	}
 
 	async function onSingleClick({ coordinate }: MapBrowserEvent) {
+		if (coreStore.showContextMenu) {
+			return
+		}
```
warm-coolguy: What is this supposed to do? I don't notice any difference.

src/core/utils/map/setupMarkers.ts
```diff
@@ -357,6 +361,9 @@ function mapClick(event: MapBrowserEvent) {
 }
 
 function mapSingleClick(event: MapBrowserEvent) {
+	if (useContextMenuStore().show) {
+		return
+	}
```
warm-coolguy: What is this supposed to do? I don't notice any difference.

src/core/utils/map/setupMarkers.ts
```diff
@@ -304,6 +305,9 @@ function mapPointerMove({ map, pixel }: MapBrowserEvent) {
 }
 
 function mapClick(event: MapBrowserEvent) {
+	if (useContextMenuStore().show) {
+		return
```
warm-coolguy: What is this supposed to do? I don't notice any difference.

warm-coolguy (changes requested):
📫 @dopenguin

src/components/kern/KernButton.ce.vue
dopenguin: Removed the background-color-override which causes the hover issue 7bdcdd1a7f9ca183141cc7c0ccb79f38f2008ab4

Generally, `--kern-btn-text-color` is to dynamically override the color of the icon and the text.

src/core/components/ContextMenu.ce.vue
dopenguin: So having the focus on the first element, but not having the focus visible here, is preferrable?

Testing the behaviour with multiple buttons can be achieved by adding the following snippet to the snowbox `index.js`

```ts
const buttons = [
	{
		id: 'color',
		icon: 'kern-icon--palette',
		text: 'Farben anpassen',
		callback: (c) => console.warn('Colour', c),
	},
	{
		id: 'path',
		icon: 'kern-icon--rebase-edit',
		text: 'Pfad anpassen',
		callback: (c) => console.warn('Path', c),
	},
	{
		id: 'duplicate',
		icon: 'kern-icon--content-copy',
		text: 'Duplizieren',
		callback: (c) => console.warn('Duplicate', c),
	},
	{
		id: 'delete',
		icon: 'kern-icon--delete',
		text: 'Löschen',
		callback: (c) => console.warn('Delete', c),
		color: 'oklch(50.78% 0.202 29.22)',
		group: 'delete',
	},
]
buttons.forEach((b) => map.store.addToContextMenu(b))
```

src/core/components/ContextMenu.ce.vue
dopenguin: The border-radius of the card now matches the buttons 7120ee56b099ef556f1c3c722c679212b7fa7e89

Bildschirmfoto 2026-07-01 um 15 47 52

The outline is the default outline of the browser. The outline of the icon buttons of the IconMenu use a completely different approach so that an active button can be highlighted as well.

src/core/components/PolarContainer.ce.vue
dopenguin: That is intended and in line with how both Google Maps and OpenStreetMap handle their contextmenu.

I don't think that implementing browser specific ways to open the default contextmenu of the browser is something we should be doing.

src/core/components/ContextMenu.ce.vue
dopenguin: Works on an iPhone 15 Pro with iOS v26.5 in Chrome

https://github.com/user-attachments/assets/cdabb707-27ed-4f72-b8e9-1470b829c946

src/core/components/ContextMenu.ce.vue
dopenguin: Removed circles altogether a030005902523546edc1d4da7ca0d28176a7e2af

src/core/components/ContextMenu.ce.vue
dopenguin: Good idea, shall I create an issue for that?

src/core/components/ContextMenu.ce.vue
dopenguin: Yup 21faad05df3760cb3f86987295fe5ad750ef6f0e

src/core/components/ContextMenu.ce.vue
dopenguin: Fixed 3bd129005a8c5280c21d63a3344c12ab3eba7fc4

src/core/stores/contextMenu.ts
dopenguin: Outside the core, that variable is not accessible, as only `addEntry` and `removeEntry` are exposed as `addToContextMenu` and `removeFromContextMenu`.
While I understand your approach here, I think this can stay as it is.

src/core/components/PolarContainer.ce.vue
dopenguin: Feels weird moving this to the store, but sure f846cba40529c04581058eb681b76af3e3fbaa1a

src/plugins/filter/components/FilterUI.ce.vue
dopenguin: o7 9d793238195918c7899d565e514eba658ddc9767

src/plugins/pointerPosition/store.ts
dopenguin: I say we postpone this. I even suggest we write an issue for this to tackle this after draw has been migrated.

src/plugins/pins/store.ts
dopenguin: This was a trial to remove the issue you mentioned in https://github.com/Dataport/polar/pull/745#discussion_r3413349101. As this clearly doesn't do the job, I removed these along with the other instances 4300635cd051c7d7a33368185835bed08bc58630

src/core/utils/map/setupMarkers.ts
dopenguin: See https://github.com/Dataport/polar/pull/745#discussion_r3506863658

src/core/utils/map/setupMarkers.ts
dopenguin: See https://github.com/Dataport/polar/pull/745#discussion_r3506863658

dopenguin:
> 📫 @dopenguin

Except https://github.com/Dataport/polar/pull/745#discussion_r3413349101, I've tackled every thread. That is something, I'll need some more time for.

🏓 @warm-coolguy until then

src/core/components/ContextMenu.ce.vue
warm-coolguy: > So having the focus on the first element, but not having the focus visible here, is preferrable?

Yeah, or, if you dislike that, there's the other option I pointed out.

However, `:focus-visible` is explicitly built for this.

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:focus-visible

It even uses a "variety of heuristics".

ChatGPT Image 2  Juli 2026, 11_17_20

src/core/components/ContextMenu.ce.vue
warm-coolguy: > The outline of the icon buttons of the IconMenu use a completely different approach so that an active button can be highlighted as well.

Impressive. Very nice. Can you change it?

Since the outline is also pretty much on the map background, and I don't fancy multiple outline styles in one product, they should look the same.

src/core/components/PolarContainer.ce.vue
warm-coolguy: Since most browser don't implement it at all and we block the browser default behaviour, I think we should.

I sent an inquiry to our user experience human.

src/core/components/ContextMenu.ce.vue
warm-coolguy: But I want to go in circles, it's fun.

src/core/components/ContextMenu.ce.vue
warm-coolguy: Yes.

src/components/kern/KernButton.ce.vue
warm-coolguy: Sure, but what is it good for? Shouldn't the buttons already be styled via kern.css?

src/core/components/ContextMenu.ce.vue
warm-coolguy: Can't view that in-browser. Downloaded the video and the Windows 11 software trying to run it offers me to buy the required codec for 99c. I don't have that kind of money.

Anyway, sent you a video via mail.

1.) Weirdly, I can press the context menu button outside of the map (where it leaks, which it is supposed to), but not inside the map.
2.) I accidentally also select the map on longpress a lot of times when opening the context menu.

src/core/stores/contextMenu.ts
warm-coolguy: I disagree and I am very mad at you. 🎩

src/components/kern/KernButton.ce.vue
warm-coolguy: Turns out I didn't get `--kern-btn-text-color` isn't a KERN variable at all, but a local override. We've discussed this and we'll choose a name making clearer that this is an optional override.

dopenguin: Renamed the variable for clarity 0d2ee9e570eaa087a7b82257b86cff2316691de3

src/plugins/pointerPosition/store.ts
warm-coolguy: Okay.

warm-coolguy:
> > 📫 @dopenguin
> 
> Except [#745 (comment)](https://github.com/Dataport/polar/pull/745#discussion_r3413349101), I've tackled every thread. That is something, I'll need some more time for.
> 
> 🏓 @warm-coolguy until then

🕺 @dopenguin 

nice_track

src/core/components/ContextMenu.ce.vue
warm-coolguy: We can also make it an issue.

src/core/components/ContextMenu.ce.vue
dopenguin: Done #844

src/core/components/ContextMenu.ce.vue
dopenguin: What style should we be using?
KERN currently works with

```css
  outline: 4px solid var(--kern-color-action-focus-default-contextual, #454B6B);
  outline-offset: 2px;
```

This approach wouldn't really work with the IconMenu-Buttons. So should we be deviating completely and just use the styling we currently see on the `PolarIconButton`?

src/core/components/PolarContainer.ce.vue
dopenguin: Any updates?

src/plugins/pointerPosition/store.ts
dopenguin: See #869

src/core/components/ContextMenu.ce.vue
dopenguin: 8fe63a3f92963738c5e0a3c7b1d74f107af1979a

i_guess_we_doin_circles_now

src/core/components/ContextMenu.ce.vue
dopenguin: The map no longer can be selected in snowbox a5b21b4073637ff339a2177a21ffb03fc0eb8b8e

dopenguin: It also worked on my phone as I used the PR preview (HTTPS) and not the local version (HTTP). `navigator.clipboard` is only available in secure contexts. This should now work as intended in both scenarios e1f85e0bc2a9a94f3b02502450ec2f01ac068fe6

src/core/components/ContextMenu.ce.vue
dopenguin: Yeah, there's no way we are able to say "Hey, the user really should see this" as https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus#focusvisible is not widely available yet.
I removed the outline 07a7c8cf290864dc714bda58f3ece9317b6a1a4b

dopenguin:
> > > 📫 @dopenguin
> > 
> > 
> > Except [#745 (comment)](https://github.com/Dataport/polar/pull/745#discussion_r3413349101), I've tackled every thread. That is something, I'll need some more time for.
> > 🏓 @warm-coolguy until then
> 
> 🕺 @dopenguin
> nice_track

~~Still on the longpress issue, have at it in the meantime~~ 🏓 @warm-coolguy

src/core/components/PolarContainer.ce.vue
dopenguin: Finally fixed this ec0adc5331e240b1609235d2535071661bbb90f7

dopenguin: … aaaaaaaaaand it behaves weirdly with `markers`. Will tackle that as well.

**Edit**: Fixed 02d8f5d7fd48936ec6e26b06e488b6c68a84f3bb

src/core/components/ContextMenu.ce.vue
warm-coolguy: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:focus-visible

It's widely available, you're on the wrong resource. Funny that they used that name twice.

src/core/components/PolarContainer.ce.vue
warm-coolguy: It's pretty much "It depends". I put it on today's weekly's agenda to get this discussed and done.

src/core/components/ContextMenu.ce.vue
warm-coolguy: We talked in person and decided that all elements directly on the map should have the POLAR Icon Button styling and all elements on a different background get the KERN default outline.

Press C for Confirm: C

src/core/components/PolarContainer.ce.vue
warm-coolguy: We decided that preventing the user from using the native context menu (except for in Firefox) is alright. 😶

src/core/components/ContextMenu.ce.vue
warm-coolguy: TODO: test locally

src/core/components/ContextMenu.ce.vue
dopenguin: C #871

warm-coolguy (approved):
> > > > 📫 @dopenguin
> > > 
> > > 
> > > Except [#745 (comment)](https://github.com/Dataport/polar/pull/745#discussion_r3413349101), I've tackled every thread. That is something, I'll need some more time for.
> > > 🏓 @warm-coolguy until then
> > 
> > 🕺 @dopenguin
> > nice_track
> 
> ~~Still on the longpress issue, have at it in the meantime~~ 🏓 @warm-coolguy 
> 
> 

:shipit: @dopenguin

PR #748, opened 2026-06-01
fix(core): remove cycles that were hidden by vite imports

dopenguin:
## Summary

Some cyclic references were previously hidden by the vite import aliases which are now resolved.

The pipelines should still be green.


oeninghe-dataport (approved):
🚀 @dopenguin

PR #749, opened 2026-06-01
test(snowbox): fix plugin detection for export plugin

oeninghe-dataport:
## Summary

The disable export plugin button in iceberg does not properly detect when the plugin is disabled.
Therefore, it cannot be enabled again.

- Open iceberg
- Disable export via button
- Enable export via button

None.


dopenguin (approved):
🚀 @oeninghe-dataport

PR #750, opened 2026-06-01
Enforce unified import syntax

oeninghe-dataport:
## Summary

We introduce two new linting rules that ensure:

1. Type imports need to be specified as `type` imports.
2. The optional `.ts` ending for imports is omitted.
3. Within the same context (that means: either core, a lib module or a plugin), we use relative imports. Within different contexts, we use `@/...`-style absolute imports.

No visible changes, just review the code.

Blocked by #748 

## Review process

This should be approved by @warm-coolguy and @dopenguin.
If necessary, we'll discuss this in the weekly.


eslintRules/import-style.ts
```diff
@@ -0,0 +1,190 @@
...
+	const rel = path.relative(srcRoot, abs).replace(/\\/g, '/')
+	return rel.startsWith('..') ? null : rel
+}
+
+/**
+ * Enforces POLAR import style conventions:
+ *
+ *  1. `.ts` extensions must be omitted in import/export sources.
+ *  2. Same-context imports must use relative paths (`./` or `../`).
+ *  3. Cross-context or context-less imports must use the `@/…` alias.
```
dopenguin, line 62: This should be split in three separate rules.

It also seems reasonable to add them to [eslint-config-geodev](https://github.com/Dataport/eslint-config-geodev) as they may seem usable across projects.

eslint.config.ts
```diff
@@ -81,6 +85,7 @@ const polarTsConfig = defineConfig({
 				groups: ['required-member', 'unknown'],
 			},
 		],
+		'@typescript-eslint/consistent-type-imports': 'error',
```
dopenguin: Please add this rule to a separate PR as this rule does not need extensive discussion compared to the import style rule IMO.

dopenguin (changes requested):
🏓 @oeninghe-dataport

eslintRules/import-style.ts
oeninghe-dataport: - This could be split in two rules at most. 2. and 3. belong to each other, as the import style is changed to the one or other.
- The definition of context (2. + 3.) is highly project-specific and is therefore not re-usable without further effort (introducing configuration could be an option here).

Therefore, IMO, only the first part could be split to a rule in eslint-config-geodev, while the second and third part would stay here. Do you agree?

eslint.config.ts
oeninghe-dataport: See #809

oeninghe-dataport:
🏓 @dopenguin

eslintRules/import-style.ts
dopenguin: Please put these rules on our agenda for our weekly so we can discuss these rule changes with everyone.

dopenguin: We can leave 1. here for now. Adding it to `eslint-config-geodev` may be a future endeavour.

Regarding the other rules:

- The same context imports of types are slightly weird because of the re-exports of them. I don't really like the `../..`.
- Why are `components` considered cross-context?

dopenguin:
> 🏓 @dopenguin

🏓 @oeninghe-dataport

eslintRules/import-style.ts
oeninghe-dataport: I also don't like the `../..` imports. I don't think we can nor should fix this automatically, but I could change them manually to go to `../../types` or similar. Would that be an acceptable solution?

oeninghe-dataport: About the `components`: I'd argue that they're similar to `lib`; each component is a usable unit on its own. If you disagree: Would you also change the behaviour for libs?

oeninghe-dataport:
🏓 @dopenguin

eslintRules/import-style.ts
dopenguin: > I also don't like the `../..` imports. I don't think we can nor should fix this automatically, but I could change them manually to go to `../../types` or similar. Would that be an acceptable solution?

Sounds reasonable, please do. Note that it may sometimes be a different file as in the case of `PolarContainer` e.g. being the `index` file

dopenguin:
> 🏓 @dopenguin

Please also rebase the PR

🏓 @oeninghe-dataport

eslintRules/import-style.ts
oeninghe-dataport: a44b496181

oeninghe-dataport: https://github.com/Dataport/polar/commit/a44b49618199d7df350104e1a09cefd2eef4b5c5#diff-128fa1a3973c05fe3d2ef09d85e7e0765738f6695c106c0ebf74377c271b301bR6 needs to stay as-is, because it is actually defined there and not just re-exported.

src/plugins/iconMenu/types.ts
```diff
@@ -80,7 +79,7 @@ export interface IconMenuPluginOptions extends PluginOptions {
 	 * @remarks
 	 * Plugins like GeoLocation can not be added here, as only plugins containing content are allowed.
 	 */
-	focusMenus?: (Menu & { icon: Icon })[]
+	focusMenus?: (Menu & { icon: string })[]
```
dopenguin: This should stay the way it was.

src/core/types/locales.ts
```diff
@@ -1,7 +1,6 @@
 import type { ResourceKey } from 'i18next'
-import type { BundledPluginId, BundledPluginLocaleResources } from '@/core'
-import type { resourcesEn as core } from '@/core/locales'
-import type { CoreId } from '@/core/types'
+import type { BundledPluginId, BundledPluginLocaleResources, CoreId } from '..'
```
dopenguin: This should be split in import of `plugin.ts` and `types.ts`

dopenguin (changes requested):
🏓 @oeninghe-dataport

src/plugins/iconMenu/types.ts
oeninghe-dataport: I have no clue how that happened. However, 6b689f130f

src/core/types/locales.ts
oeninghe-dataport: ff59a57913c37e0e351525039ecad36fe747c865. There was also a fix to the linting rule for a edge case relevant here

src/core/types/plugin.ts
```diff
@@ -47,9 +48,14 @@ import type { PluginId as ZoomPluginId } from '@/plugins/zoom'
...
 import type { NineLayoutTag } from '../utils/NineLayoutTag'
-import type { Locale } from './locales'
 import type { Icon } from './theme'
 
+/** @internal */
+export interface Locale {
+	resources: Record<string, ResourceKey>
+	type: string
+}
+
```
dopenguin, line 58: This was better placed in `src/core/types/locales.ts`, why the move?

eslintRules/import-style.ts
```diff
@@ -0,0 +1,196 @@
...
+ *
+ * A "context" is one of:
+ *   - `core`             for every file inside `src/core/`
+ *   - `plugins/<name>`   for every file inside `src/plugins/<name>/`
+ *   - `lib/<name>`       for every file inside `src/lib/<name>/`
```
dopenguin: `components` can be added here

dopenguin (changes requested):
🏓 @oeninghe-dataport

src/core/types/plugin.ts
oeninghe-dataport: This produced a cycle between `src/core/types/plugin.ts` and `src/core/types/locales.ts` that is forbidden by our architectural checks (and was, for some reason, just not detected before changing the imports).

eslintRules/import-style.ts
oeninghe-dataport: 1a266275ab

src/core/types/plugin.ts
dopenguin: I am bit baffled that the type `Locale` is only used in `plugin.ts` and not in `locales.ts`, but so be it 🎩

dopenguin (dismissed):
1/2 approves given, @warm-coolguy, you're next!

dopenguin (approved):
:shipit: @oeninghe-dataport

PR #751, opened 2026-06-01
(#618) Automatically remove previews from gh-pages

oeninghe-dataport:
## Summary

Removal of previews from GH-pages fails too often.

- Merge PR
- See the still existing preview

For obvious reasons, this PR is untested.

Hopefully reduces #618

### DIFF
--- a/.github/workflows/preview.yaml
+++ b/.github/workflows/preview.yaml
@@ -1,5 +1,7 @@
 name: Deploy PR previews
 
+concurrency: preview-${{ github.ref }}
+
 on:
   pull_request:
     types:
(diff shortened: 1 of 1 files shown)


dopenguin (commented):
If I am not mistaken, the pipeline would simply stop the next step if the event is closed, right?
But how does this help with the removal of gh-pages deployments?

oeninghe-dataport:
> If I am not mistaken, the pipeline would simply stop the next step if the event is closed, right? But how does this help with the removal of gh-pages deployments?

The pipeline skips building the project when the event is a closed (i.e., merged or declined) pull request.
This is also recommended by https://github.com/rossjrw/pr-preview-action#usage
It speeds up the pipeline runtime at merging - and, even more important - it does not fail when a PR is closed that fails, for example, at `npm ci`.

The concurrency should prohibit conflicting jobs of the same PR that run concurrently.
This is also recommended by https://github.com/rossjrw/pr-preview-action#set-a-concurrency-group

dopenguin (approved):
🚀 @oeninghe-dataport

PR #752, opened 2026-06-01
style: forbid { ns: literal } for t calls via ESLint

### LINKED ISSUE #508: POLAR@3: Using `PluginId` as the namespace in plugins
Our convention is to use `{ ns: PluginId }` in `$t` calls. There are some occurences that do not match this convention.

_Originally posted by @dopenguin in https://github.com/Dataport/polar/pull/436#discussion_r2718200250_
            
This should be changed to match. Also, there should be a linting rule that checks if the convention is met.

oeninghe-dataport:
## Summary

Forbid { ns: literal } for t calls via ESLint

Fixes #508 

Blocked by #750

### DIFF
--- a/eslintRules/no-literal-ns-in-t.ts
+++ b/eslintRules/no-literal-ns-in-t.ts
@@ -1,0 +1,63 @@
+import type { Rule } from 'eslint'
+
+/**
+ * Enforces that the `ns` option in `t`/`$t` calls uses a constant identifier
+ * (e.g. `PluginId` or `CoreId`) instead of a hard-coded string literal.
+ *
+ * Bad:  `t(($) => $.key, { ns: 'filter' })`
+ * Good: `t(($) => $.key, { ns: PluginId })`
+ */
+const noLiteralNsInT: Rule.RuleModule = {
+	meta: {
+		type: 'suggestion',
+		docs: {
+			description:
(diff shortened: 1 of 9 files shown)


eslintRules/no-literal-ns-in-t.ts
dopenguin, line 1: Can this be made auto-fixable? I'm sure @warm-coolguy would really enjoy that!

Also, if I add a literal in a template, this rule does not raise an issue.

dopenguin (changes requested):
🏓 @oeninghe-dataport

eslintRules/no-literal-ns-in-t.ts
warm-coolguy: While I would really enjoy that, I don't think it's plausible to invest that amount of time. Also confirming it doesn't work in templates and there's a lot of templates with the issue present.

Suggestion: We drop the extra rule and just search for `ns: '` and replace it in every file and call it a day.

Pro: Saves time and we have another nice thing to nag about in future PRs.
Con: There is none and it won't stop anybody from just declaring the string locally anyway. I'll just find creative ways to annoy you if you proceed.

warm-coolguy (changes requested):
I request the change "Close pull request".

dopenguin:
Superseded by #804

PR #768, opened 2026-06-03
Pins: Consider external updates and allow ignoring change sources

### LINKED ISSUE #655: AddressSearch, Pins, ReverseGeocoder: Reference hell
**Describe the bug**
The selected result of a successful search may be updated without the intent of the user.

This happens in a typical configuration that can be e.g. found in `snowbox`.
There, a selected search result sets a pin, which triggers a reverse geocoding. The result of that gets added to the addressSearch input.

This leads to two problems:
- The user sometimes sees the loader for extended periods of time due to the unwanted reverse geocoding
- The changed result may confuse users

**Expected behavior**
The selected result may only be updated if the user selects a result or sets a pin themselves.

**Steps to reproduce**

1. Start `snowbox`
3. Search for "Billstraße"
4. Select "Billstraße"
5. The result gets updated to "Billstraße 144"

**Additional context**
This problem also occurrs on `main` as far as I can tell and thus should be fixed on both branches.

It is also possible that other plugins may have such cyclic references. It may be reasonable to tackle this issue in a generalized scoped.

### LINKED ISSUE #766: Changes to pins coordinate are not reflected in the map
If you change `pinsStore.coordinate` from the outside, the effects (watchers) are executed, but the pin is not moved in the map.

oeninghe-dataport:
## Summary

Consider external updates (#766) and allow ignoring change sources (#655).

See #766 and #655 

Fixes #766 
Fixes #655


dopenguin:
The target should probably be `next` and not `main`, @oeninghe-dataport 😄

warm-coolguy:
> The target should probably be `next` and not `main`, @oeninghe-dataport 😄

 Fixed.

sailormoonmeme_6358795

src/composables/useStoreWatcher.ts
```diff
@@ -88,14 +91,26 @@ export function usePluginStoreWatcher(
...
+				`"${watcherConfig.source.plugin}" not found. Cannot watch "${watcherConfig.source.key}".`
 			)
 			return
 		}
 
 		watcherConfig.handle = watch(
 			() => store[watcherConfig.source.key],
-			watcherConfig.callback,
+			(value) => {
+				const sourceKey = watcherConfig.source.key + 'Source'
```
dopenguin: This lacks documentation. Please add information e.g. to the TSDoc of the composable that, if I watch the parameter `coordinate` I will ignore updates if I have a parameter `coordinateSource` in my store and have `ignoredSources` configured.

As this requires a lot of knowledge and extra implementation for each plugin with a possible reference loop, isn't there a possibility of a different approach without the need of further boilerplate in any plugins?

src/plugins/pins/store.ts
```diff
@@ -105,7 +108,6 @@ export const usePinsStore = defineStore('plugins/pins', () => {
 	function teardownPlugin() {
 		const { map } = coreStore
 		map.un('singleclick', onSingleClick)
-		removePin()
```
dopenguin, line 106: Is the `coordinate` reset automatically on teardown?

dopenguin (changes requested):
🏓 @oeninghe-dataport

src/plugins/pins/store.ts
oeninghe-dataport: No. The pin is removed, but the `coordinate` is not reset yet. This is fixed w/ d1295f0c6a.

src/composables/useStoreWatcher.ts
oeninghe-dataport: For the documentation: dfb50f253f

On your question:

I'm also not too happy with the current solution. However, the following aspects convinced me to go with this:

- The pins plugin seems to be the natural participant in reference loops, even in the future. It allows reading and writing of other plugins in a useful way, which is the case for few plugins; usually, our plugins communicate in one direction only. I don't expect that we will need this in another plugin in the future.
- I found no approach without further boilerplate (that is not much more "hack-y" than this one). At least this approach allows re-using it in every plugin that refers to pins.

oeninghe-dataport:
🏓 @dopenguin

src/composables/useStoreWatcher.ts
dopenguin: I've fiddled around a bit with the help of Copilot and came up with a832128672a74e0c4be1e6689b12b8d4d95ed1aa. Take a look and let me know if you find any flaws.

dopenguin:
> 🏓 @dopenguin

🏓 @oeninghe-dataport

src/composables/useStoreWatcher.ts
oeninghe-dataport: - I do not think this idea works (yet, or at all). I added a `console.log` to the `if(...)` statement in line 138 of `useStoreWatcher`, and this is not executed. I'd expect that this line is executed in the case of #655.
- I'm not quite sure I understood the code perfectly yet. If we'd want to introduce this, this needs more documentation.
- If we'd want to introduce this, IMO, we need unit tests here. I know that we decided in general to write tests in a later stage, however, this mechanism is hard to test manually and prone to errors and regressions, and I'd make an exception here.

oeninghe-dataport:
🏓 @dopenguin

src/composables/useStoreWatcher.ts
dopenguin: ## Doesn't work yet

*Oh*, it does work.

no_loops

Previously, it would further resolve to Billstraße 144. The loop is broken.

## Documentation

There are several TSDoc and normal comments. Can you elaborate on what exactly needs documenting?

## Tests

Sure, added two for the loop-breaking-mechanism e6571d96b837956ec9b11d97c1259efadcd960e6

dopenguin:
> 🏓 @dopenguin

@oeninghe-dataport 

pingpong

dopenguin (approved):

PR #447, opened 2026-01-08
feat(gfi): migrate plugin

oeninghe-dataport:
## Summary

Migrate the GFI plugin.

## Instructions for local reproduction and review

- Open snowbox.
- Click on a marker.
- See the result.

## Additional hints

- The utils `requestGfi*` were migrated as-is and do not need to be reviewed therefore.

## Relevant tickets, issues, et cetera

Closes #368

PR #769, opened 2026-06-04
chore: enforce sorted named imports

dopenguin:
## Summary

Add new new linting rule and fix the resulting issues.

Pipeline should pass.

Initially introduced in #447

### DIFF
--- a/src/composables/usePluginStoreWatcher.ts
+++ b/src/composables/usePluginStoreWatcher.ts
@@ -6,9 +6,9 @@
 
 import {
 	computed,
+	type ComputedRef,
 	onScopeDispose,
 	watch,
-	type ComputedRef,
 	type WatchOptions,
 	type WatchStopHandle,
 } from 'vue'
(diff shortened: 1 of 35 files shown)


src/composables/usePluginStoreWatcher.ts
```diff
@@ -6,9 +6,9 @@
 
 import {
 	computed,
+	type ComputedRef,
```
dopenguin, line 9: @oeninghe-dataport I'd prefer to have this grouped with the other types. Can we tackle this in this PR as well?

oeninghe-dataport (approved):

src/composables/usePluginStoreWatcher.ts
oeninghe-dataport: This is currently tackled in https://github.com/Dataport/polar/pull/750/changes#diff-3202509a51dddde1968b20a04dfcedc90615eb6681cca0cda5f43a418a516857R88 which is merge-ready after your approval

PR #780, opened 2026-06-08
fix(addressSearch): display SearchResults above icon buttons

dopenguin:
## Summary

Add z-index so the SearchResults are displayed above icon buttons (e.g. export).

Take a look at snowbox and marvel at the neatly visible SearchResults after a successful search.


warm-coolguy (approved):
:shipit: @dopenguin Tested, liked, and subscribed.

PR #786, opened 2026-06-10
Dish/dish export map/selected backgroundlayer in print

jedi-of-the-sea:
## Summary

DishExportMap should use the selected background layer as background in the print. 

It also seems like we need another urlParam for prod and qs, because the base service url for print services differ from `printHostDegree`.

- start client in mode 'INTERN'
- change background layer to basemap
- click on a monument feature in the map. You might need to zoom in to be able to click them. you need to use the split tunnel to be able to request the monument feature service
- select "Kartendruck"
- a new tab opens in DISH with a login formular (ask me for user / pw)
- login and redo the steps above. Instead of a login formular you should get a pdf with a map from the feature you clicked on.

Unfortunately, we cannot request the other services (Grundkarte Graustufen, Grundkarte Farbe, Luftbilder Farbe), so it is only possible to see a result with basemap. But you can check the URLs to see if the other service urls are used in there.

  - [x] Chrome Lighthouse
  - [x] Firefox Accessibility

### DIFF
--- a/packages/clients/dish/src/mapConfigurations/mapConfigIntern.ts
+++ b/packages/clients/dish/src/mapConfigurations/mapConfigIntern.ts
@@ -179,7 +179,7 @@ export const mapConfigIntern = (urlParams: DishUrlParams): DishMapConfig => ({
     wmsLayerUrl: `${urlParams.printServicesBaseUrl}/wms`,
     wfsLayerUrl: `${urlParams.printServicesBaseUrl}/wfs`,
     wfsLayerFeatureType: 'app:TBLGIS_ORA',
-    printImageUrlProd: `${urlParams.printImageUrlProd}/Content/MapsTmp`,
+    printImageUrlProd: urlParams.printImageUrlProd ? `${urlParams.printImageUrlProd}/Content/MapsTmp` : `${urlParams.printHostDeegree}/Content/MapsTmp` ,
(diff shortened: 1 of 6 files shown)


packages/clients/dish/src/mapConfigurations/mapConfigIntern.ts
```diff
@@ -179,11 +179,9 @@ export const mapConfigIntern = (urlParams: DishUrlParams): DishMapConfig => ({
     wmsLayerUrl: `${urlParams.printServicesBaseUrl}/wms`,
     wfsLayerUrl: `${urlParams.printServicesBaseUrl}/wfs`,
     wfsLayerFeatureType: 'app:TBLGIS_ORA',
-    printImageUrlProd: `${urlParams.printHostDeegree}/Content/MapsTmp`,
+    printImageUrlProd: urlParams.printImageBaseUrl
+      ? `${urlParams.printImageBaseUrl}/Content/MapsTmp`
+      : `${urlParams.printHostDeegree}/Content/MapsTmp`,
```
warm-coolguy:
```suggestion
    printImageUrlProd: `${
      urlParams.printImageBaseUrl ?? urlParams.printHostDeegree
    }/Content/MapsTmp`,
```

IMO ein bisschen leichter zu lesen, aber nur ein Vorschlag. 🎩

warm-coolguy (changes requested):
@jedi-of-the-sea 🏓

warm-coolguy (approved):
:shipit: @jedi-of-the-sea

PR #789, opened 2026-06-12
fix: add name prop so radio inputs are properly grouped together 

dopenguin:
## Summary

Radio inputs are only grouped together in a tab group if they share the same `name` attribute. This PR introduces that as a new prop.

`npm run snowbox` -> The radio inputs in layerChooser now should only be reachable with one tab and navigatable with arrow keys.


warm-coolguy (approved):
:shipit: @dopenguin Very nice. Impressive.

PR #790, opened 2026-06-12
fix(layerChooser): correctly interpolate legend button locale

dopenguin:
## Summary

Namespaces behave a bit differently in other locales. Also, the interpolation was completely amiss.

`npm run snowbox` -> the legend buttons should now use a correct label.


warm-coolguy (approved):
:shipit: @dopenguin Good find. Good fix.

PR #804, opened 2026-06-16
Refactor/beeline namespaces

### LINKED ISSUE #508: POLAR@3: Using `PluginId` as the namespace in plugins

warm-coolguy:
## Summary

Quick manual fix for underlying issue in https://github.com/Dataport/polar/pull/752 since we all agree it makes sense.

_By the power of RegEx! I have the power!_

### DIFF
--- a/src/core/components/PolarContainer.ce.vue
+++ b/src/core/components/PolarContainer.ce.vue
@@ -41,12 +41,15 @@ import {
 	watch,
 } from 'vue'
 
-import type { MapConfiguration, MasterportalApiServiceRegister } from '../types'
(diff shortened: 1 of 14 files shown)


dopenguin (approved):
:shipit: @warm-coolguy

PR #807, opened 2026-06-17
refactor: polar select

dopenguin:
## Summary

`PolarSelect` now uses `defineModel` in favour of the previous `emit` solution.

All adjusted plugins should still work without any issues.

### DIFF
--- a/src/components/PolarSelect.ce.vue
+++ b/src/components/PolarSelect.ce.vue
@@ -6,13 +6,12 @@
 		<div class="kern-form-input__select-wrapper" :class="{ small }">
 			<select
 				:id="resolvedId"
+				v-model="model"
 				class="kern-form-input__select"
 				:aria-label="ariaLabel"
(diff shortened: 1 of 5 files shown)


src/components/PolarSelect.ce.vue
```diff
@@ -1,29 +1,26 @@
 <template>
```
warm-coolguy, line 1: When switching to v-model, you may also consider to switch to v-model, so you can use vue shorthands while you use vue shorthands: https://github.com/Dataport/polar/pull/807/changes/95e49bec6d27a30608e7b8c68e63a0fbb859ecf8 🎩

warm-coolguy (dismissed):
🚀 @dopenguin 

I may have interfered with the implementation. You may want to check back on my fiddling before merging.

warm-coolguy (approved):
🦤🐧 @dopenguin

PR #810, opened 2026-06-18
refactor(filter): use PolarInput

### LINKED ISSUE #779: Replace KernBlockButton inputs with PolarInput and PolarInputGroup
## Description

The `filter` plugin currently uses `KernBlockButtonCheckbox` and `KernBlockButtonRadioGroup` for the `standard` layout, and `KernCheckbox` / `KernRadioGroup` for other layouts. This results in four separate components with overlapping purpose and a dynamic component switch at runtime.

These should be consolidated to use `PolarInput` and `PolarInputGroup` as the single underlying input components.

## Expected behavior

- `PolarInput` (type `checkbox` or `radio`) combined with `PolarInputGroup` is used in all cases
- The `standard` layout's block-button styling is applied via CSS (e.g. a class on the group or scoped styles)
- `KernBlockButtonCheckbox`, `KernBlockButtonRadioGroup`, `KernCheckbox`, and `KernRadioGroup` are removed

## Notes

- The optional `icon` prop currently available on `KernBlockButtonCheckbox` and radio group items can be rendered via `PolarInput`'s slot
- The block-button styling (background `#edf1fa`, checked border highlight, custom focus ring) must be preserved for `standard` layout

### Out of scope

- Changes to `PolarInput` or `PolarInputGroup` component APIs

dopenguin:
## Summary

Replaces layout-conditional checkbox/radio components (`KernBlockButtonCheckbox`, `KernCheckbox`, `KernBlockButtonRadioGroup`, `KernRadioGroup`) in `filter` with the unified `PolarInput` component (as proposed in #734).
Simplifies `category` `state` from `Record<string, boolean>` to `string[]` and introduces a reactive getter/setter-based `selection` model on categories for direct `v-model` binding and to get rid of `getCategoryStatus` and `setCategoryStatus`.

Also added some small changes to `PolarInput` and `PolarInputGroup`.

1. `npm run snowbox`
2. Open `filter`
3. Verify category checkboxes can be toggled individually and via "Alle an-/abwählen"
4. Verify the `1xx` multi-value category entry correctly selects/deselects with its individual values (`100`, `101`, `102`)
5. Verify time filter works
6. Verify the date range picker scrolls into view when "Zeitraum wählen" is selected and selection works

## Additional information

I intend to update the implementation in `src/plugins/filter/stores/category.ts` (a) and `src/plugins/filter/stores/time.ts` (#811) as well.
For (a): Currently, if a user switches between two layers, the filter state of the the previous layer is reset. This is odd from a usability perspective. If you agree, I'll change this in a separate PR so that the state stays even if the layer switches.

## Tickets

Resolves #779


oeninghe-dataport (changes requested):
Without looking into the code, it's stated in #779:

> The standard layout's block-button styling is applied via CSS (e.g. a class on the group or scoped styles)

However, the standard layout is not intact anymore but looks like the nineRegions layout.

Before:

Image

After:

Image

Please update your PR such that the styling before is still working.

dopenguin:
> Without looking into the code, it's stated in #779:
> 
> > The standard layout's block-button styling is applied via CSS (e.g. a class on the group or scoped styles)
> 
> However, the standard layout is not intact anymore but looks like the nineRegions layout.
> 
> Before:
> Image
> 
> After:
> Image
> 
> Please update your PR such that the styling before is still working.

After consulting with UX, I've adjusted the issue to scrap the block-button styling as the radios and checkboxes should not behave the same visually as one would not properly discern the possibility to select multiple elements with the checkboxes or is confused as one would only be able to select one element with the radio buttons.
We can revisit this approach later again if needed.

PR #824, opened 2026-06-26
ci: ignore @types/node major updates in dependabot to align with used node version

dopenguin:
## Summary

Dependabot is trying to update `@types/node` to the latest major, which should not be happened.

Check the validity of the yaml.

Resolves #817


oeninghe-dataport (approved):
🚀 @dopenguin

PR #839, opened 2026-06-30
refactor: use native DOMParser in favour of xml2js

dopenguin:
## Summary

By using the `DOMParser`, we are able to get rid of `xml2js`, required polyfills and transitive dependencies.
As `@masterportal/masterportalapi` also no longer uses `xml2js`, this is the perfect time to change this for ourselves as well.

- Tests should be green
- Build should be fine
- Linting should still work
- dev mode should still work


oeninghe-dataport (approved):
I'm very happy with the changes to the package.json and vite.config.ts. Finally, we get rid of this 🥳 

🚀 @dopenguin

PR #841, opened 2026-07-02
feat: add reusable toggle button component

dopenguin:
## Summary

Adds a reusable component that behaves like a radio, but looks like a toggleable button.

Can be seen in action in #842

#373, #842

### DIFF
--- a/src/components/PolarToggleButton.ce.vue
+++ b/src/components/PolarToggleButton.ce.vue
@@ -1,0 +1,126 @@
+<template>
+	<fieldset>
+		<legend :class="labelSrOnly ? 'kern-sr-only' : 'kern-label'">
+			{{ label }}
+		</legend>
+		<label v-for="option in options" :key="option.value">
+			<input
+				v-model="model"
+				type="radio"
+				:name="groupName"
+				:value="option.value"
+				:disabled="option.disabled"
+				:aria-label="unref(option.label)"
+			/>
+			<span class="kern-btn kern-btn--tertiary" aria-hidden="true">
+				<span class="kern-icon" :class="option.icon" aria-hidden="true" />
+			</span>
+		</label>
+	</fieldset>
+</template>
+
+<script setup lang="ts">
+import type { Ref } from 'vue'
+import type { Icon } from '@/core'
+
+import { unref, useId } from 'vue'
+
+interface PolarToggleOption {
+	label: string | Ref<string>
+	value: string
+	disabled?: boolean
+	icon?: Icon
+}
+
+defineProps<{
+	label: string
+	options: PolarToggleOption[]
+	labelSrOnly?: boolean
+}>()
+const model = defineModel<string>({ required: true })
+const groupName = useId()
+</script>
+
+<style scoped>
+fieldset {
+	display: flex;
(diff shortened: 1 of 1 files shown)


src/components/PolarToggleButton.ce.vue
```diff
@@ -0,0 +1,120 @@
...
+import { unref, useId } from 'vue'
+
+interface PolarToggleOption {
+	label: string | Ref<string>
+	value: string
+	disabled?: boolean
+	icon?: Icon
+}
+
+defineProps<{ label: string; options: PolarToggleOption[] }>()
```
oeninghe-dataport: Why don't just inline that interface? It is neither reused nor exported.

src/components/PolarToggleButton.ce.vue
```diff
@@ -0,0 +1,120 @@
+<template>
+	<fieldset>
+		<legend class="kern-sr-only">{{ label }}</legend>
```
oeninghe-dataport: Looking at the implementation in #842, it might be preferable in some cases to actually show the legend. Maybe it should be configurable or always shown?

src/components/PolarToggleButton.ce.vue
```diff
@@ -0,0 +1,120 @@
...
+		<label v-for="option in options" :key="option.value">
+			<input
+				v-model="model"
+				type="radio"
+				:name="groupName"
+				:value="option.value"
+				:disabled="option.disabled"
+				:aria-label="unref(option.label)"
+			/>
+			<span class="kern-btn kern-btn--tertiary" aria-hidden="true">
```
oeninghe-dataport, line 15: It would seem more natural and matching to KERN to me to make this `<span>` a `<label for="...">`, and use a `<span class="kern-label kern-sr-only">` for the `option.label` instead of the `aria-label` above.

This would match the KERN requirements for a `kern-btn` and allow to remove the `aria-hidden` here.

oeninghe-dataport (changes requested):
🏓 @dopenguin

src/components/PolarToggleButton.ce.vue
dopenguin: You currently save one line of code.
More important than that, it is more readable IMO

src/components/PolarToggleButton.ce.vue
dopenguin: Adjusted, it is now configurable

src/components/PolarToggleButton.ce.vue
dopenguin: I actively deviated from the standard KERN approach to actually make it properly accessible, as KERN does not offer a toggle button so far. These are also not buttons, but rather radio elements disguised as buttons.

I can show you in detail in a call, why I approached the component this way.

dopenguin:
> 🏓 @dopenguin

🏓 @oeninghe-dataport

src/components/PolarToggleButton.ce.vue
oeninghe-dataport: Discussed and agreed

oeninghe-dataport (approved):
🚀 @dopenguin

PR #842, opened 2026-07-02
feat(routing): migrate plugin

### LINKED ISSUE #373: Migration of Routing plugin to POLAR@3

dopenguin:
## Summary

The `routing` plugin is migrated to POLAR@3.

- Go to https://openrouteservice.org/ and either create an account or login to your HeiGIT account
- Retrieve your ORS apiKey and add it to the `examples/snowbox/index.js`
- `npm run snowbox`
- Start routing with various different options

Also, adjust the configuration to possibly display more options.

Closes #373,
Blocked by #841, #745

The plugin is later enhanced by #864, #865, #866, #867


src/plugins/routing/index.ts
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * @returns Plugin for use with {@link addPlugin}.
+ */
+export default function pluginRouting(
+	options: RoutingPluginOptions
+): PluginContainer {
+	return {
+		id: PluginId,
+		component,
+		locales,
```
oeninghe-dataport, line 34:
```suggestion
		locales,
		icon: 'kern-icon-fill--assistant-direction',
```

src/plugins/routing/index.ts
```diff
@@ -0,0 +1,40 @@
...
+
+import component from './components/RoutingWrapper.ce.vue'
+import locales from './locales'
+import { useRoutingStore } from './store'
+import { PluginId } from './types'
+
+/**
+ * Creates a plugin which offers routing functionality to the user.
+ *
+ * A user can select multiple waypoints by clicking on the map, which then are converted to an address, if the `reverseGeocoder` plugin is configured.
```
oeninghe-dataport: This should be added w/ #864 and not stated until it's true (as it might be that #864 is implemented much later or never)

examples/snowbox/index.js
```diff
@@ -488,6 +489,16 @@ addPlugin(
 						}, */
 					}),
 				},
+				{
+					plugin: pluginRouting({
+						type: 'ors',
+						url: 'https://api.openrouteservice.org/v2/directions/',
+						apiKey: '',
```
oeninghe-dataport, line 496: I understand that we cannot write a real API key here. However, I don't like the idea of having unstaged changes every time I wanna test routing. (Which would lead to routing being only tested when really necessary.)

Could we introduce a configuration / env var / etc. here for the API key (or list a proxy here that passes a valid API key)?

src/plugins/routing/index.ts
oeninghe-dataport: This would also fix the broken icon in the PR preview.

examples/snowbox/index.js
oeninghe-dataport: Until that is implemented, as a note to myself for testing:

```javascript
document.getElementById('snowbox').store.configuration.routing.apiKey = 'insert_your_apiKey_here'
```

https://account.heigit.org/manage/key

src/plugins/routing/components/RoutingWrapper.ce.vue
```diff
@@ -0,0 +1,45 @@
+<template>
+	<PolarCard>
+		<header class="kern-card__header">
+			<h2 class="kern-title">{{ $t(($) => $.title, { ns: PluginId }) }}</h2>
```
oeninghe-dataport, line 4: This header has the same formatting as the sub-headers "Start" and "Ziel". They seem to be semantically correct, however, I don't like I cannot visually differ between the levels.

src/plugins/routing/components/RoutingDetails.ce.vue
```diff
@@ -0,0 +1,193 @@
...
+	}
+	return `${distance} m`
+}
+function formatDuration(duration: number) {
+	if (duration >= 3600) {
+		return `${(duration / 3600).toFixed(2)} h`
+	} else if (duration >= 60) {
+		return `${(duration / 60).toFixed(1)} min`
+	}
+	return `${duration} sec`
```
oeninghe-dataport, line 121: This should be localized and the decimal separator should be formatted with the locale, too.

oeninghe-dataport: Maybe, using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DurationFormat/format is a good option here instead of re-implementing this ourself?

src/plugins/routing/components/RoutingDetails.ce.vue
```diff
@@ -0,0 +1,193 @@
...
+const duration = computed(() =>
+	segments.value.reduce((acc, segment) => acc + segment.duration, 0)
+)
+const steps = computed(() => segments.value.flatMap((segment) => segment.steps))
+
+function formatDistance(distance: number) {
+	if (distance >= 1000) {
+		return `${(distance / 1000).toFixed(1)} km`
+	}
+	return `${distance} m`
```
oeninghe-dataport: Although there is (sadly) no `Intl.DistanceFormat`, we should at least localize the number itself (i.e., the decimal separator), probably using `Intl.NumberFormat`.

src/plugins/routing/utils/handleErrors.ts
oeninghe-dataport, line 1: As this logic is not too complex and used only once, why not inline it?

src/plugins/routing/store.ts
```diff
@@ -0,0 +1,460 @@
...
+			if (!signal.aborted) {
+				handleErrors(error)
+			}
+		}
+	}
+
+	function initializeDraw() {
+		draw = new Draw({ stopClick: true, type: 'Point' })
+		// @ts-expect-error | internal hack to detect it in @polar/plugin-pins and @polar/plugin-gfi
+		draw._isRoutingDraw = true
```
oeninghe-dataport, line 225: If we want to introduce 36db39c59d17dc841f45ff5d5881173e77486822, this should use that pattern too. I'd split that commit off as a separate PR if you agree.

src/plugins/routing/store.ts
```diff
@@ -0,0 +1,460 @@
...
+			],
+			() => {
+				if (!routeIncomplete.value) {
+					void getRoute()
+				}
+			}
+		)
+		stopTravelModeWatch = watch(selectedTravelMode, () => {
+			selectedRouteTypesToAvoid.value = []
+		})
```
oeninghe-dataport: These watchers should be defined outside the `setupPlugin` function. Then we don't need `stopRouteWatch` and `stopTravelModeWatch`, but they're just stopped when the store is unloaded.

src/plugins/routing/store.ts
oeninghe-dataport, line 1: This would be the longest plugin store we have in this project. (Technically, `geoLocation` is longer, but it has tests; without, it's not.)

This is not a problem in and of itself, but the file is hard to follow.

Therefore, I propose splitting this into multiple files with the following ideas:

- composable for the draw interaction
- composable for the route layer
- util for `fetchRoute`

src/client.ts
```diff
@@ -39,6 +40,9 @@ function addPlugins(map: typeof PolarContainer, enabledPlugins: string[]) {
 				enabledPlugins.includes('geoLocation') && {
 					plugin: GeoLocation({ renderType: 'iconMenu' }),
 				},
+				enabledPlugins.includes('geoLocation') && {
```
oeninghe-dataport: I assume this should be `'routing'`

src/plugins/routing/store.ts
```diff
@@ -0,0 +1,460 @@
...
+	function addCoordinateToRoute(coordinate: Coordinate) {
+		route.value = route.value.toSpliced(
+			currentlyFocusedInput.value,
+			1,
+			coordinate
+		)
+	}
+
+	async function fetchRoute(signal: AbortSignal): Promise<RoutingResponseData> {
+		const response = await fetch(encodeURI(url.value), {
```
oeninghe-dataport: Why is `encodeURI` necessary here?

src/plugins/routing/components/RoutingInput.ce.vue
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * This makes sure that there are always two fillable input fields at max.
+ */
+const addWaypointButtonDisabled = computed(() => {
+	return (
+		route.value.filter((part) => Boolean(part.length)).length <
+		route.value.length - 1
+	)
+})
```
oeninghe-dataport: This can be simplified to `computed(() => route.value...)`

src/plugins/routing/components/RoutingInput.ce.vue
```diff
@@ -0,0 +1,92 @@
+<template>
+	<div class="polar-plugin-routing-route-wrapper">
+		<div class="kern-form-input">
+			<label class="kern-label" :for="`polar-plugin-routing-input-${index}`">
+				{{ $t(($) => $.label[getRouteLabel(index)], { ns: PluginId }) }}
+			</label>
+			<input
+				:id="`polar-plugin-routing-input-${index}`"
```
oeninghe-dataport, line 8: Why not `useId` here?

oeninghe-dataport (changes requested):
🏓 @dopenguin

src/plugins/routing/index.ts
dopenguin: Adjusted e0e7b67db44e7319aac10fc52fe2c9ae8a12fdf8

examples/snowbox/index.js
dopenguin: We intended to receive an organisational API key so we can proxy requests through our development server for testing. I can create an issue or add it to our weekly agenda if you want.

src/plugins/routing/components/RoutingWrapper.ce.vue
dopenguin: That is just the way KERN styles `kern-title` and `kern-label`. Thus, I wouldn't change it.

src/plugins/routing/components/RoutingDetails.ce.vue
dopenguin: Sounds like a good idea, but TypeScript tells me, that I can't do that. Changing the `ecmaVersion` in `.eslintrc.json` doesn't change that as well and `tsconfig.json` already has `esnext` as its target.

Bildschirmfoto 2026-07-10 um 16 09 23

src/plugins/routing/components/RoutingDetails.ce.vue
dopenguin: Despite the MDN documentation, it does not support strings.

Bildschirmfoto 2026-07-10 um 16 16 00

src/plugins/routing/utils/handleErrors.ts
dopenguin: Seemed like an easy thing to move to a different file without loosing readability as the function name is clear enough without the need to look at the implementation.

src/client.ts
dopenguin: Yup cf55a71d7e89baf88d7ba5f67d6533194495c343

src/plugins/routing/components/RoutingInput.ce.vue
dopenguin: Done 96d49c4f1d8ef2f213304ef68c7fc986432421ea

src/plugins/routing/components/RoutingInput.ce.vue
dopenguin: I like my ids readable, but I don't mind changing this b4985ca325afee7b60e5f4b757911455ccef59d3

src/plugins/routing/store.ts
dopenguin: 57d11d70de09d8319eee0de3977d7c2a7624187f done

src/plugins/routing/store.ts
dopenguin: It isn't 6930ad864cfa561866ecd206031d119d48d41008

src/plugins/routing/store.ts
dopenguin: The implementation seems like a clear step up from the current workarounds. Could you provide me with an example on how this would work with routing, e.g. cherry-picking that commit onto a branch based on `vue3/migrate-plugin-routing`?

src/plugins/routing/store.ts
dopenguin: It is the longest plugin store *so far* 😏

- As draw is a) dynamically added and b) is directly intertwined with the store in `drawend`, I left it as is.
- `fetchRoute` accesses 6 store values. I prefer leaving it in the store.
- Added a composable for the route layer baee986204ffbf1bf675ff9fed66c55b3fe002e6

dopenguin:
> 🏓 @dopenguin

🏐 @oeninghe-dataport 

serve

src/plugins/routing/components/RoutingDetails.ce.vue
oeninghe-dataport: It does not need to support strings. I'd call it just with `distance / 1000` (without `.toFixed(1)`) and pass `minimumFractionDigits: 1` and `maximumFractionDigits: 1` (as explained at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#digit_options) to the constructor.

src/plugins/routing/store.ts
oeninghe-dataport: Untested, but see #873

dopenguin: Yeah, it doesn't seem to work directly.

The change itself seems to be a good idea but I think it is preferable to tackle that change after routing.

src/plugins/routing/components/RoutingInput.ce.vue
dopenguin: The Ids are needed to update the focus so a draw interaction is added. Dropped the commit. Probably the reason for https://github.com/Dataport/polar/pull/842#discussion_r3569097434

src/plugins/routing/components/RoutingDetails.ce.vue
dopenguin: Adjusted d5702fd0bf210c8891ff09f98fdee9eb3e242724

src/plugins/routing/components/RoutingDetails.ce.vue
dopenguin: Simply localized the numbers for now d5702fd0bf210c8891ff09f98fdee9eb3e242724

oeninghe-dataport: According to https://www.typescriptlang.org/docs/handbook/release-notes/typescript-6-0.html#es2025-option-for-target-and-lib and https://github.com/microsoft/TypeScript/pull/63046 , this is supported in TS starting at version 6.0 which cannot be introduced currently as neostandard blocks this. Please add an issue here to change this when TS 6.0 is introduced.

Besides that, this is fine for me 🎩

src/plugins/routing/store.ts
oeninghe-dataport: According to https://github.com/Dataport/polar/pull/842#discussion_r3569198451 it could work, or not? Please elaborate if this is still up-to-date

dopenguin: Updated the PR and works like a charm now.
Merging #873 after this PR seems like the best approach.

src/plugins/routing/components/RoutingDetails.ce.vue
dopenguin: Added the information to the dependabot PR, suffices for me.

src/plugins/routing/store.ts
oeninghe-dataport: Agreed

examples/snowbox/index.js
dopenguin: We called and decided to put it on the Agenda.

oeninghe-dataport (approved):
🪨et @dopenguin

PR #843, opened 2026-07-02
refactor: use one stylelint config so IDE plugins can use it properly

dopenguin:
## Summary

As the title suggests, this fixes issues with IDE plugins and reduces overhead.

Pipeline should be green


stylelint.config.ts
```diff
@@ -1,8 +1,15 @@
...
+import path from 'node:path'
+import { fileURLToPath } from 'node:url'
+
+const repoRoot = path.dirname(fileURLToPath(import.meta.url))
+
 export default {
 	extends: ['stylelint-config-recommended', 'stylelint-config-recommended-vue'],
 	plugins: ['stylelint-value-no-unknown-custom-properties'],
+	// TODO: Remove 'vue2/**' after migration
+	ignoreFiles: ['examples/iceberg/**', 'vue2/**'],
```
oeninghe-dataport: Actually, I'd propose to not ignore `iceberg` but provide an override. `iceberg` has no custom variables, so there is no need to extract variables or similar. And, according to https://stylelint.io/user-guide/configure/#ignorefiles , the current configuration would lint the `node_modules`

package.json
```diff
@@ -54,10 +54,8 @@
 		"build:github-io": "vite build --config vite.config.github-io.ts",
 		"lint": "eslint --cache --cache-location .eslintcache --cache-strategy content .",
 		"lint:fix": "npm run lint -- --fix",
-		"lint:css": "npm run lint:css:src && npm run lint:css:github-io",
-		"lint:css:src": "stylelint \"src/**/*.vue\"",
-		"lint:css:github-io": "cd examples/github-io && stylelint \"**/*.vue\"",
-		"lint:css:fix": "npm run lint:css:src -- --fix && npm run lint:css:github-io -- --fix",
+		"lint:css": "stylelint \"**/*.vue\"",
```
oeninghe-dataport:
```suggestion
		"lint:css": "stylelint \"src/**/*.{css,vue}\" \"examples/**/*.{css,vue}\"",
```

stylelint.config.ts
oeninghe-dataport:
```suggestion
```

stylelint.config.ts
```diff
@@ -30,4 +37,24 @@ export default {
 		],
 		/* eslint-enable @typescript-eslint/naming-convention */
 	},
+	overrides: [
```
oeninghe-dataport, line 40:
```suggestion
	overrides: [
		{
			// Generic configuration for examples
			// This may be overridden, if necessary, per example.
			files: ['examples/**/*.{css,vue}'],
			rules: {
				/* eslint-disable @typescript-eslint/naming-convention */
				'csstools/value-no-unknown-custom-properties': [true],
				/* eslint-enable @typescript-eslint/naming-convention */
			},
		},
```

oeninghe-dataport (changes requested):
🏓 @dopenguin

stylelint.config.ts
dopenguin: Added `node_modules`, just skipped the second sentence 2eef0835d36c133d3f63ce023c3999c0897c09e2

I'd like to keep this pattern. `iceberg` is currently not being maintained, so we can keep it like this until that changes.

package.json
dopenguin: Added CSS files 97f13530a8df067cd37bbb7f46ae8857b1ddbb5c

dopenguin:
> 🏓 @dopenguin

Overhead is now minimal, let's keep it like this?

🏓 @oeninghe-dataport

oeninghe-dataport (approved):
I'd still prefer `{src/examples}/**/*.{css,vue}` over `**/*.{css,vue}` with an ignore list. We don't have code outside these folders, so this would be more efficient; and we would not need to remove `vue2` later. 🎩 

I'd still prefer including `iceberg`. We will probably forget to re-add this if we do later 🎩

PR #846, opened 2026-07-03
chore: ignore local built files for stylelint

oeninghe-dataport:
## Summary

When there are local built files (`**/dist/**.css`, `.dist.preview/**.css`, `docs-html/**.css` etc.), they should not be linted.
As we already have a list of these files in `.gitignore`, we'll use that as a base for stylelint, too.

Add the following `dist/my.css`:

```css
thisinvalidhtmltag {
    with-an-invalid-attribute: var(--uses-an-invalid-var);
}
```


dopenguin (changes requested):
The `**/node_modules/**` glob in `stylelint.config.ts` can be removed as well then.

🏓 @oeninghe-dataport

oeninghe-dataport:
🏓 @dopenguin

dopenguin (approved):
:shipit: @oeninghe-dataport

PR #848, opened 2026-07-03
fix(layerChooser): add support for non-positive zoomLevels

### LINKED ISSUE #673: Snowbox can zoom to layerless levels
**Describe the bug**
In the snowbox client, when zooming fully out, no visible map material is left. This is either a configuration issue or a bug. When zooming out to the last possible layers, the configures WMS switches to another view (meaning set of loaded tiles here) for a brief moment before being hidden.

**Expected behavior**
Our test client should have visible layers on all zoom levels.

oeninghe-dataport:
## Summary

Fixes #673 

- Open snowbox
- Zoom out to level 0


dopenguin (approved):
🚀 @oeninghe-dataport

PR #861, opened 2026-07-06
Allow removal and adding for plugins in iconMenu

### LINKED ISSUE #668: Plugins added by IconMenu can not be removed with "removePlugin"
**Describe the bug**
Like the title describes, plugins added by `iconMenu` can not be removed with `removePlugin`.

**Expected behavior**
I'd expect that any plugins can be removed.

**Steps to reproduce**
1. Add a button that removes e.g. `geoLocation` in `snowbox`
2. Start `snowbox`
3. Click the button
4. `geoLocation` is still part of `iconMenu` and now behaves weirdly

oeninghe-dataport:
## Summary

Plugins in iconMenu can be added and removed at runtime.

**Note:**
`coreStore.removePlugin(...)` is still not supported for plugins within iconMenu. I deem this correct. As iconMenu adds the plugin, it should also be responsible for removal of the plugin. Otherwise, we'd need to breach the separation of core and plugins.

- Open snowbox.
- Add the layer "Anliegen (MML)"
- Open filter plugin.
- Filter out some "Anliegen".
- Press the newly-introduced "Add/Remove filter plugin" button.

Fixes #668


src/plugins/filter/store.ts
```diff
@@ -52,10 +52,12 @@ export const useFilterStore = defineStore('plugins/filter', () => {
 					{ deep: true, immediate: true }
 				)
 			)
+			teardownCallbacks.push(callback)
```
dopenguin, line 53: Why is this needed?

src/plugins/iconMenu/store.ts
```diff
@@ -104,21 +112,31 @@ export const useIconMenuStore = defineStore('plugins/iconMenu', () => {
...
-			openInMoveHandle(openId)
 		}
 	}
+	watch(open, (open) => {
+		if (open) {
+			openInMoveHandle(open)
+		} else {
+			coreStore.setMoveHandle(null)
+		}
+	})
```
dopenguin: I prefer the previously implemented version in favour of this side effect solution. Same goes for `focusOpen`.

What is the practical reason why this is needed or is simply personal preference?

src/plugins/iconMenu/store.ts
```diff
@@ -64,24 +90,6 @@ export const useIconMenuStore = defineStore('plugins/iconMenu', () => {
...
-			)
-		)
-		focusMenus.value.map((menuItem) =>
-			toMerged(menuItem, {
-				plugin: {
-					component: markRaw(menuItem.plugin.component as Component),
-				},
-			})
-		)
-
```
dopenguin: The important part - making the component non-reactive with `markRaw` - seems to be lost on the new implementation.

src/plugins/iconMenu/store.ts
```diff
@@ -42,6 +41,33 @@ export const useIconMenuStore = defineStore('plugins/iconMenu', () => {
...
+					open.value = null
+				}
+				if (focusOpen.value === oldItem.plugin.id) {
+					focusOpen.value = null
+				}
+				coreStore.removePlugin(oldItem.plugin.id)
+			})
+		},
+		{ deep: true }
+	)
```
dopenguin: I agree that `iconMenu` should handle the removal itself so the separation of concerns is not broken.

However, I would've thought adding its own `removePlugin` action that handles that procedure would be the way to go here. Note that, even though they are currently described otherwise in the export, neither `menu` nor `focusMenus` are considered stable API. I've made a note for myself to go over all plugins to adjust the exports to the requirements and mark most as `@alpha`.

dopenguin (changes requested):
🏓 @oeninghe-dataport

src/plugins/iconMenu/store.ts
oeninghe-dataport: These lines have no effect anymore. Or do I miss anything?

src/plugins/iconMenu/store.ts
oeninghe-dataport: My first approach was to implement a `addPlugin` and `removePlugin` action. However, they get quite complicated, as you have to check which index and subindex to consider, remove an empty array if it is empty; and insertion similar.

If you prefer to have these actions, I can add them. I'd stick to this watcher to keep the more flexible option of direct editing, but we could e.g. mark only the actions as stable to keep `menus` as unstable API.

src/plugins/iconMenu/store.ts
oeninghe-dataport: Practical reason.

We had that logic already duplicated in `NineRegionsButton` and `StandardMenuList`, and I don't wanted to copy it a third time at https://github.com/Dataport/polar/pull/861/changes#diff-8793dc8dda9a07c08b84fcfa5edf13989ab72fde14867879719bfba5765c229aR59-R64

src/plugins/filter/store.ts
oeninghe-dataport: When the filter plugin is removed, I expect that the filtering is resetted. Therefore, this line updates the state on teardown, and state is resetted before at `teardownPlugin`.

oeninghe-dataport:
🏓 @dopenguin

src/plugins/iconMenu/store.ts
dopenguin: I prefer these actions, yeah.

src/plugins/iconMenu/store.ts
dopenguin: It could also be de-duplicated by abstracting the `toggle` functions of both components into an action.
I deem this favourable over the side-effect solution.

dopenguin:
> 🏓 @dopenguin

🏓 @oeninghe-dataport

src/plugins/iconMenu/store.ts
oeninghe-dataport: IIRC, we discussed in person appending would be enough here; therefore, b47d6cfd5a

src/plugins/iconMenu/store.ts
oeninghe-dataport: I did miss something. `markRaw` is modifying in-place. Therefore, it is used again.

src/plugins/iconMenu/store.ts
oeninghe-dataport: And removed the watchers here w/ 28a8ef6b8ffc192e47e07cf245a28607b3f6db4c

src/plugins/iconMenu/store.ts
oeninghe-dataport: I do not. However, there is an architectural decision supporting your opinion now. Therefore: 1d136e78b7

oeninghe-dataport:
🏓 @dopenguin

src/lib/watchArray.ts
dopenguin, line 1: These functions are not in use. Please remove them.

src/plugins/iconMenu/components/NineRegionsButton.ce.vue
```diff
@@ -31,11 +30,9 @@ const updateMaxWidth = inject('updateMaxWidth') as () => void
...
 function toggle() {
 	if (open.value === props.id) {
-		open.value = null
-		useCoreStore().setMoveHandle(null)
+		iconMenuStore.openMenuById(null)
 	} else {
-		open.value = props.id
-		iconMenuStore.openInMoveHandle(props.id)
+		iconMenuStore.openMenuById(props.id)
 	}
```
dopenguin: Might as well just call `iconMenuStore.openMenuById(open.value === props.id ? null : props.id)`

src/plugins/iconMenu/components/StandardMenuList.ce.vue
```diff
@@ -123,11 +123,9 @@ function updateMaxWidth() {
...
 function toggle(id: string) {
 	if (open.value === id) {
-		open.value = null
-		coreStore.setMoveHandle(null)
+		iconMenuStore.openMenuById(null)
 	} else {
-		open.value = id
-		iconMenuStore.openInMoveHandle(id)
+		iconMenuStore.openMenuById(id)
 	}
```
dopenguin: Might as well just call `iconMenuStore.openMenuById(open.value === id ? null : id)`

src/plugins/iconMenu/store.ts
```diff
@@ -172,13 +189,50 @@ export const useIconMenuStore = defineStore('plugins/iconMenu', () => {
...
 		visibleFocusMenus,
-		open,
-		focusOpen,
+
+		/**
+		 * Determines which menu is currently open.
+		 *
+		 * To change the open menu, use {@link openMenuById}.
+		 *
+		 * @readonly
```
dopenguin, line 211:
```suggestion
		 * @readonly
		 * @alpha
```

src/plugins/iconMenu/store.ts
```diff
@@ -172,13 +189,50 @@ export const useIconMenuStore = defineStore('plugins/iconMenu', () => {
...
+		 */
+		open: readonly(open),
+
+		/**
+		 * Determines which focus menu is currently open.
+		 *
+		 * To change the open focus menu, use {@link openFocusMenuById}.
+		 *
+		 * @readonly
```
dopenguin, line 221:
```suggestion
		 * @readonly
		 * @alpha
```

src/plugins/iconMenu/store.ts
```diff
@@ -172,13 +189,50 @@ export const useIconMenuStore = defineStore('plugins/iconMenu', () => {
...
+
 		buttonComponent,
-		openInMoveHandle,
 		openMenuById,
 		openFocusMenuById,
 
+		/**
+		 * Appends a group of plugins to the icon menu.
+		 *
+		 * @param menu - The menu item group to add.
```
dopenguin, line 233:
```suggestion
		 * @param menu - The menu item group to add.
		 * @alpha
```

src/plugins/iconMenu/store.ts
```diff
@@ -172,13 +189,50 @@ export const useIconMenuStore = defineStore('plugins/iconMenu', () => {
...
+		 */
+		addPlugin,
+
+		/**
+		 * Appends a plugin to the icon menu as a focus menu.
+		 *
+		 * @param menu - The focus menu item to add.
```
dopenguin, line 241:
```suggestion
		 * @param menu - The focus menu item to add.
		 * @alpha
```

src/plugins/iconMenu/store.ts
```diff
@@ -172,13 +189,50 @@ export const useIconMenuStore = defineStore('plugins/iconMenu', () => {
...
+		 */
+		addFocusPlugin,
+
+		/**
+		 * Removes a plugin from the icon menu.
+		 *
+		 * @param pluginId - The ID of the plugin to remove.
```
dopenguin, line 249:
```suggestion
		 * @param pluginId - The ID of the plugin to remove.
		 * @alpha
```

src/plugins/iconMenu/store.ts
```diff
@@ -42,6 +40,50 @@ export const useIconMenuStore = defineStore('plugins/iconMenu', () => {
...
+		const display = coreStore.configuration[pluginId]?.displayComponent
+		return typeof display === 'boolean' ? display : true
+	}
+	function addPlugin(menuGroup: Menu[]) {
+		const filteredMenuGroup = menuGroup.filter(({ plugin: { id } }) =>
+			isPluginInIconMenu(id)
+		)
+		filteredMenuGroup.forEach(({ plugin }) => {
+			if (plugin.component) {
+				markRaw(plugin.component)
```
dopenguin: It was previously more defensive with an additional `toRaw`. I feel like this should be readded here and in `addFocusPlugin`

src/plugins/iconMenu/store.ts
```diff
@@ -42,6 +40,50 @@ export const useIconMenuStore = defineStore('plugins/iconMenu', () => {
 		() => coreStore.configuration.iconMenu?.layoutTag ?? ''
 	)
 
+	function isPluginInIconMenu(pluginId: string) {
+		const display = coreStore.configuration[pluginId]?.displayComponent
+		return typeof display === 'boolean' ? display : true
+	}
```
dopenguin, line 46: This is not related to the intention of this PR thus 🎩, but I realized that this never did anything if the configuration was directly added to the plugin. If you see a quick solution, add it here, otherwise, I'll make a note for myself.

dopenguin (changes requested):
Quite a lot of actions seem to fire when removing the plugin 🤨 

Image

🏓 @oeninghe-dataport

src/lib/watchArray.ts
oeninghe-dataport: fe137607e7

src/plugins/iconMenu/components/NineRegionsButton.ce.vue
oeninghe-dataport: d2b0008192

src/plugins/iconMenu/components/StandardMenuList.ce.vue
oeninghe-dataport: d2b0008192

src/plugins/iconMenu/store.ts
oeninghe-dataport: Don't think this is necessary, however, it does not hurt. 7b57647c4cdc198911a7f74ed3592b64607aa42c

src/plugins/iconMenu/store.ts
oeninghe-dataport: I do not get why this check is necessary at all. I kept it here to prohibit regression, but in what scenario I'd add a plugin to the iconMenu (not standalone) but then do not display it?

oeninghe-dataport: So, if you agree, I'll remove this check

oeninghe-dataport:
🏓 @dopenguin

src/plugins/iconMenu/store.ts
dopenguin: I honestly thought that both `layoutTag` and `displayComponent` both don't do anything for plugins added to the `iconMenu`. Please add this to our next meeting agenda so we can think about dropping this.
Most clients already handle the use case this would tackle by simply not adding them then.

dopenguin:
> 🏓 @dopenguin

Please take a look at the amount of actions mentioned in https://github.com/Dataport/polar/pull/861#pullrequestreview-4829529204

🏓 @oeninghe-dataport

oeninghe-dataport:
> > 🏓 @dopenguin
> 
> Please take a look at the amount of actions mentioned in [#861 (review)](https://github.com/Dataport/polar/pull/861#pullrequestreview-4829529204)
> 
> 🏓 @oeninghe-dataport

The actions were there for several reasons. I reduced them a lot with 7bbdec2a71b5119aee2f322a3adf8d759e7f3572

oeninghe-dataport:
🏓 @dopenguin

src/core/stores/plugin.ts
```diff
@@ -67,20 +67,17 @@ export const usePluginStore = defineStore('plugin', () => {
...
 
-	function getPluginStore<T extends PluginId>(
+	function getPluginStore<T extends BundledPluginId>(
 		id: T
-	): ReturnType<
-		T extends BundledPluginId
-			? BundledPluginStores<typeof id>
-			: PolarPluginStore
-	> | null {
+	): ReturnType<BundledPluginStores<T>> | null
```
dopenguin, line 72: Wasn't it the intention of the function to also be able to retrieve the store of custom plugins with this function?

src/core/stores/plugin.ts
```diff
@@ -67,20 +67,17 @@ export const usePluginStore = defineStore('plugin', () => {
...
+	function getPluginStore(id: PluginId): ReturnType<PolarPluginStore> | null
+	function getPluginStore(id: PluginId): unknown {
 		const plugin = plugins.find((plugin) => plugin.id === id)
-		// @ts-expect-error | We trust that our internal IDs work.
 		return plugin?.storeModule?.(mainStore._instance) || null
 	}
 
 	return {
-		plugins,
+		plugins: readonly(plugins) as readonly PluginContainer[],
```
dopenguin, line 80: Is this the same as using a `computed`?

src/plugins/iconMenu/store.ts
```diff
@@ -70,17 +70,30 @@ export const useIconMenuStore = defineStore('plugins/iconMenu', () => {
...
 		}
-		menus.value = menus.value
-			.map((menuGroup) =>
-				menuGroup.filter(({ plugin: { id } }) => id !== pluginId)
+		const pluginIndex = menus.value.findIndex((menuGroup) =>
+			menuGroup.some(({ plugin: { id } }) => id === pluginId)
+		)
+		if (pluginIndex !== -1) {
+			// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+			const pluginMenu = menus.value[pluginIndex]!
```
dopenguin, line 78: Simply moving `pluginMenu` outside the `if`-statement and then checking whether `pluginMenu` works even better.

dopenguin (changes requested):
🏓 @oeninghe-dataport

src/core/stores/plugin.ts
oeninghe-dataport: It is...well it is mostly. `readonly` is deep. The exact equivalent to `computed(() => plugins.value)` would probably be `shallowReadonly(plugins)`

src/plugins/iconMenu/store.ts
oeninghe-dataport: As discussed in-person, I removed the check: c1552fbc5a

src/core/stores/plugin.ts
oeninghe-dataport: It is; see line 73

src/plugins/iconMenu/store.ts
oeninghe-dataport: 68725613cd

oeninghe-dataport:
🏓 @dopenguin

dopenguin:
@oeninghe-dataport I'll take a look after you updated the PR

PR #873, opened 2026-07-10
POLAR@3: Introduce interaction masking via core for routing

oeninghe-dataport:
See https://github.com/Dataport/polar/pull/842#discussion_r3559764721

Blocked by #842


src/core/stores/index.ts
```diff
@@ -120,6 +120,37 @@ export const useCoreStore = defineStore('core', () => {
 		 */
 		zoom: mainStoreRefs.zoom,
 
+		/**
+		 * Masks an interaction for a plugin.
+		 * If the interaction is already masked by another plugin, an error is thrown.
```
dopenguin, line 124: So, if I start the draw plugin and want to draw a polygon, then decide that I want to instead find a route but forget to turn of the draw plugin - `maskInteraction` fails in routing with an error but the user does not receive an information about that. That doesn't really seem the way to go.

This can currently be tested if one removes the `if`-statement in `maskInteraction` and then clicking on one of the input-fields of the routing plugin.

src/core/stores/main.ts
```diff
@@ -132,3 +156,49 @@ export const useMainStore = defineStore('main', () => {
...
+			setActivePinia(createPinia())
+			const store = useMainStore()
+			store.setup()
+			await use(store)
+			store.teardown()
+		},
+	})
+	/* eslint-enable no-empty-pattern */
+
+	test('Masking interactions works as expected', ({ store }) => {
```
dopenguin:
```suggestion
	test('interactions can be masked and unmasked', ({ store }) => {
```

src/core/stores/main.ts
```diff
@@ -132,3 +156,49 @@ export const useMainStore = defineStore('main', () => {
...
+		expect(store.isInteractionMasked(interaction)).toBe(false)
+
+		store.maskInteraction(pluginId, interaction)
+		expect(store.isInteractionMasked(interaction)).toBe(true)
+
+		store.unmaskInteraction(pluginId, interaction)
+		expect(store.isInteractionMasked(interaction)).toBe(false)
+	})
+
+	test('Masking interactions twice fails', ({ store }) => {
```
dopenguin:
```suggestion
	test('masking the same interaction twice fails', ({ store }) => {
```

dopenguin (changes requested):
🏓 @oeninghe-dataport 

I'd suggest tackling this ASAP so #447 and draw benefit from this already merged change.

src/core/stores/index.ts
oeninghe-dataport: The expected behaviour is that I can activate the next plugin s.t. the previous is deactivated then. This is implemented w/ 5fb2060b05

oeninghe-dataport:
🏓 @dopenguin

src/core/stores/main.ts
```diff
@@ -89,6 +90,31 @@ export const useMainStore = defineStore('main', () => {
 		return { ...register, ...polar } as typeof polar
 	}
 
+	const maskedInteractions = ref(
+		new globalThis.Map<string, { pluginId: PluginId; teardown: () => void }>()
```
dopenguin: I prefer changing the `ol`-import to `Map as OlMap` over `globalThis`

dopenguin:
> 🏓 @dopenguin

🏓 @oeninghe-dataport

src/core/stores/main.ts
oeninghe-dataport: 221670b5c6

oeninghe-dataport:
> > 🏓 @dopenguin
> 
> 🏓 @oeninghe-dataport

🏓 @dopenguin

dopenguin (approved):
🚀 @oeninghe-dataport

src/core/stores/main.ts
```diff
@@ -89,6 +90,31 @@ export const useMainStore = defineStore('main', () => {
 		return { ...register, ...polar } as typeof polar
 	}
 
+	const maskedInteractions = ref(
```
dopenguin, line 94: https://github.com/Dataport/polar/commit/75557977159bfb96262a8eaad2905ff01c47ecf5 is missing here.

dopenguin (changes requested):
🏓 @oeninghe-dataport

src/core/stores/main.ts
oeninghe-dataport: 6e35e46

oeninghe-dataport:
> 🏓 @oeninghe-dataport

🏓 @dopenguin

dopenguin (approved):
🚀 @oeninghe-dataport

PR #891, opened 2026-07-21
Dish/improve objektid check

jedi-of-the-sea:
## Summary

Improvement to check if there is a valid objektid as parmeter in the url. Should prevent error / warning when ObjektID is not given or empty string or if uppercase / lowercase letters are different.

- start client with mode `INTERN`
- be in VPN, you might need browser extensions to prevent CORS errors
- go to [http://localhost:1234/?ObjektID=](http://localhost:1234/?ObjektID=)
- in `main`, there will be a warning for the user, that the id has not been found, as well as an error in the browser console. This should be prevented in this branch
- go to [http://localhost:1234/?ObjektId=](http://localhost:1234/?ObjektId=12) and / or [http://localhost:1234/?objektid=](http://localhost:1234/?objektid=12)
- map should zoom to a church in Flensburg, different writing of objektid (before it was "ObjektID") should not matter anymore

This improvement was only done for internal mode. There is a URL search parameter for ObjektID in external mode (for _extendedGfi_) but the side effects of a similar adjustment there are unknown at the moment, as the backends for internal and external mode differ.

  - [x] Chrome Lighthouse
  - [x] Firefox Accessibility


packages/clients/dish/src/polar-client.ts
```diff
@@ -40,7 +40,8 @@ export default {
     const parameters = new URL(document.location as unknown as string)
       .searchParams
     // using naming from backend to avoid multiple names for same thing
-    const objektId = parameters.get('ObjektID')
+    const objektId =
+      parameters.get('ObjektID') === '' ? null : parameters.get('ObjektID')
```
dopenguin: Would `parameters.get('ObjektID') || null` be valid here as well?

jedi-of-the-sea: used your suggestion in new function `getObjektIdFromURL`

dMapybara (approved):
🕊️

dopenguin (approved):
:shipit: @jedi-of-the-sea

PR #895, opened 2026-07-24
feat(draw): migrate plugin

warm-coolguy:
## Summary

Draw is migrated.

I don't know!

Making this PR so that we have a running example on the web for bug hunt.

  - [ ] Chrome Lighthouse
  - [ ] Firefox Accessibility


src/components/PolarIconButtonSelect.ce.vue
```diff
@@ -0,0 +1,62 @@
+<!-- eslint-disable vue/enforce-style-attribute -->
+<template>
+	<div class="polar-icon-button-select">
+		<PolarIconButton
```
dopenguin, line 3: Why not use `KernButton` here? Especially as the `kern-label` basically is duplicated now

src/plugins/draw/components/DrawText.ce.vue
```diff
@@ -0,0 +1,62 @@
...
+			<label class="kern-label" for="polar-plugin-draw-textarea">
+				{{ $t(($) => $.text.content, { ns: PluginId }) }}
+			</label>
+			<textarea
+				id="polar-plugin-draw-textarea"
+				v-model="textInput"
+				class="kern-form-input__input"
+				name="polar-plugin-draw-textarea"
+			/>
+		</div>
```
dopenguin, line 23: Do you think we'll have more textareas? If so, this might be a great reusable component

src/components/PolarIconButtonSelect.ce.vue
warm-coolguy: Not fully remembering, but looking at this now, I'd argue like this:

`KernButton` has no slot where it's needed for `PolarSelect`. While I could have added another named slot, it wouldn't quite be the `KernButton` anymore after that, so I opted for this solution.

But feel free to refactor if you want to, it's not really important imo.

src/plugins/draw/components/DrawText.ce.vue
warm-coolguy: mcvwl55o8mib1

PR #906, opened 2026-07-30
chore: update search service

jedi-of-the-sea:
## Summary

The URL for the search configuration was edited.

After this is approved, the first version of this client can be released.

- start the bgw client with `npm run bgw:dev`
- type a location into the search input for "address search", e.g. "Kiel"
- there should be search results for the location
- there should be no CORS error in the console

- as an extra, you can also check if the service "Luftbilder" (background layer) is available for you. There were no changes in code for that, but the service was updated so that it is available again.

  - [x] Chrome Lighthouse
  - [x] Firefox Accessibility

### DIFF
--- a/packages/clients/bgw/CHANGELOG.md
+++ b/packages/clients/bgw/CHANGELOG.md
@@ -1,5 +1,5 @@
 # CHANGELOG
 
-## unpublished
+## 1.0.0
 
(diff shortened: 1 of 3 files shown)


dopenguin (approved):
🚀 @jedi-of-the-sea

PR #921, opened 2026-08-07
feat(core)!: rework markers for cluster usage

oeninghe-dataport:
The current implementation used `selectedBaseFeature` to keep track of which feature to select when zooming in and out.  This did not work properly, and is not intuitive either.

Therefore, we split the (meant) hovered or selected feature and its (rendered) hovered or selected cluster.  If the layer does not use clustering, cluster and feature are equivalent.

This has the additional advantage that `isSelectable` does no longer consider the cluster, but an actual feature, such that the implementation is automatically cluster-agnostical now.

This allows the GFI plugin to adapt this new API.  When `bindWithCoreHoverSelect` is used now, the synchronization with the markers is now truly bi-directional and wrapped in a single, easy-to-understand composable.  To allow the synchronization from the details view back to the marker store, the selected feature is moved to the GFI main store as a shared state between list view and details view.

One could possibly split these changes into two or three (still atomical) commits.  However, the changes only make sense in the context of this refactoring idea, and are therefore combined to one commit on purpose.

-----

This is merged onto the GFI migration, however, as the changes have a lot of changed lines and touch the core significantly, I propose to discuss this changes as a separate PR for clarity.
When this is accepted, it will (hopefully) solve a few threads in #447 

---

Blocks #447


src/core/components/PolarContainer.ce.vue
```diff
@@ -266,7 +268,7 @@ onBeforeUnmount(() => {
 	mapEl.removeEventListener('contextmenu', openContextMenu)
 	document.removeEventListener('pointerdown', contextMenuStore.dismiss)
 	if (mainStore.configuration.markers) {
-		teardownMarkers(mainStore.map)
+		markerStore.teardown()
```
dopenguin: Directly using `useMarkerStore` here seems fitting.

src/core/components/PolarMap.ce.vue
```diff
@@ -109,7 +110,7 @@ onMounted(async () => {
 		checkServiceAvailability(mainStore.configuration, mainStore.serviceRegister)
 	}
 	if (mainStore.configuration.markers) {
-		setupMarkers(mainStore.map)
+		markerStore.setup()
```
dopenguin: Directly using `useMarkerStore` here seems fitting.

dopenguin:
Note to self: https://github.com/Dataport/polar/pull/447#discussion_r3860413766 should be fixed by this.

src/core/stores/index.ts
dopenguin: Changes to this file are different to the one on `next`.
Can you highlight the difference?

src/plugins/gfi/composables/useBindWithCoreHoverSelect.ts
```diff
@@ -1,109 +1,103 @@
...
+	bindWithWatcher(coreSelectedFeature, selectedFeature, assignGfiToCoreFeature)
+	bindWithWatcher(
 		selectedFeatures,
-		(featureMap) => {
-			assignGfiToCore(coreSelectedFeature, featureMap)
-		},
-		{ immediate: true }
+		coreSelectedFeatures,
+		assignCoreToGfiCluster,
+		referenceOrder
```
dopenguin, line 95: Are these properly disposed because the composable is being called on the root level of the store?

src/plugins/gfi/composables/useBindWithCoreHoverSelect.ts
```diff
@@ -1,109 +1,103 @@
...
 }
 
-export function useBindWithCoreHoverSelect(
-	hoveredFeatures: ShallowRef<Record<string, Feature[]>>,
-	selectedFeatures: ShallowRef<Record<string, Feature[]>>,
-	coreHoveredFeature: ShallowRef<Feature | null>,
-	coreSelectedFeature: ShallowRef<Feature | null>
+function bindWithWatcher<T, S, E>(
+	target: ShallowRef<T>,
+	source: ShallowRef<S>,
```
dopenguin, line 59: Whenever I start to look at the calls of `bindWithWatcher` I stumble over the order of these.
It feels more intuitive for me that the parameters are ordered `source` then `target`.

src/plugins/gfi/composables/useBindWithCoreHoverSelect.ts
```diff
@@ -1,109 +1,103 @@
...
-
-function assignCoreToGfi(target: ShallowRef<GfiFeature>, feature: CoreFeature) {
+type CoreFeatures = Feature[]
+type GfiFeature = { layerId: string; feature: Feature } | null
+type GfiFeatures = Partial<Record<string, Feature[]>>
+
+function assignCoreToGfiCluster(
+	target: ShallowRef<GfiFeatures>,
+	features: CoreFeatures,
+	referenceOrder?: CoreFeatures
```
dopenguin, line 17: Why is the re-ordering needed?

src/plugins/gfi/composables/useBindWithCoreHoverSelect.ts
```diff
@@ -1,109 +1,103 @@
...
-	feature.set('_polarLayerId', layerId, true)
-	target.value = markRaw(
-		findLayer(coreStore.map, layerId)?.getSource() instanceof ClusterSource
-			? getCluster(coreStore.map, feature, '_polarLayerId')
-			: feature
-	)
+	target.value = markRaw({
+		layerId: feature.get('_polarLayerId'),
+		feature,
+	})
```
dopenguin:
```suggestion
	target.value = feature
		? markRaw({
				layerId: feature.get('_polarLayerId'),
				feature,
			})
		: null
```
Hope this comes through properly

src/plugins/gfi/composables/useBindWithCoreHoverSelect.ts
```diff
@@ -1,109 +1,103 @@
...
-		features.map((feature) => ({ layerId, feature }))
-	)
-
-	// The second condition is necessary for TypeScript checks.
-	if (features.length <= 0 || !features[0]) {
+	if (!feature) {
 		target.value = null
 		return
 	}
+	target.value = markRaw(feature.feature)
```
dopenguin:
```suggestion
	target.value = feature ? markRaw(feature.feature) : null
```

src/plugins/gfi/composables/useBindWithCoreHoverSelect.ts
```diff
@@ -1,109 +1,103 @@
...
-
-	if (feature === null) {
-		target.value = markRaw({})
-		return
+	const layerId = features[0]?.get('_polarLayerId')
+	const layerConfiguration = gfiMainStore.getLayerConfiguration(layerId)
+
+	if (referenceOrder) {
+		features.sort((a, b) => {
+			const getIndex = (feature) => referenceOrder.indexOf(feature)
```
dopenguin:
```suggestion
			const getIndex = (feature: Feature) => referenceOrder.indexOf(feature)
```

src/plugins/gfi/composables/useSelectedFeatures.ts
```diff
@@ -1,28 +1,115 @@
 import type { Feature as GeoJsonFeature } from 'geojson'
-import type { Feature } from 'ol'
+import type { Feature as OlFeature } from 'ol'
 
+import { isEqual, mapValues } from 'es-toolkit'
 import { GeoJSON } from 'ol/format'
-import { computed, ref, shallowRef } from 'vue'
+import { computed, shallowRef } from 'vue'
 
+interface Feature {
```
dopenguin: I certainly dislike the name as this adds a third dimension to the sometimes already confusing differentiation between a GeoJSON and an OL-Feature. I've constantly had to remember that this is a different type while reviewing this file.

src/plugins/gfi/stores/feature.ts
```diff
@@ -173,15 +183,10 @@ export const useGfiFeatureStore = defineStore('plugins/gfi/feature', () => {
 	)
```
dopenguin, line 183: As we are already updating this file:

The `Object.fromEntries(Object.entries(…))`-solution can be updated with `pickBy` from `es-toolkit`.

src/plugins/gfi/stores/feature.ts
dopenguin: As we are already updating this file:

The `Object.fromEntries(Object.entries(…))`-solution can be updated with a combination of `pickBy` and `mapValues` from `es-toolkit`.

src/plugins/gfi/composables/useSelectedFeatures.ts
```diff
@@ -1,28 +1,115 @@
...
+			features.value = mapValues(value, (list) =>
+				list.map((olFeature) => ({
+					olFeature,
+					geojsonFeature: new GeoJSON().writeFeatureObject(olFeature),
+				}))
+			)
+		},
+	})
+
+	const geojsonFeatures = computed<Record<string, GeoJsonFeature[]>>({
```
dopenguin: The naming for this variable should rather be `geoJsonFeatures`. Same goes for `geojsonFeature`

src/plugins/gfi/composables/useSelectedFeatures.ts
```diff
@@ -1,28 +1,115 @@
...
+			if (value === null) {
+				feature.value = null
+				return
+			}
+			const layerId = value.layerId
+			feature.value = {
+				layerId,
+				olFeature: value.feature,
+				geojsonFeature: new GeoJSON().writeFeatureObject(value.feature),
+			}
```
dopenguin:
```suggestion
			feature.value =
				value === null
					? null
					: {
							layerId: value.layerId,
							olFeature: value.feature,
							geojsonFeature: new GeoJSON().writeFeatureObject(value.feature),
						}
```

src/plugins/gfi/composables/useSelectedFeatures.ts
```diff
@@ -1,28 +1,115 @@
...
+			if (value === null) {
+				feature.value = null
+				return
+			}
+			const layerId = value.layerId
+			const item = features.value[layerId]?.find(({ geojsonFeature }) =>
+				isEqual(geojsonFeature, value.feature)
+			)
+			feature.value = item
+				? { ...item, layerId }
```
dopenguin, line 101: Is setting the `layerId` necessary or could this simply be `item`?

src/plugins/gfi/stores/feature.ts
```diff
@@ -55,7 +55,7 @@ export const useGfiFeatureStore = defineStore('plugins/gfi/feature', () => {
 			return
 		}
 
-		gfiMainStore.featureInformation = result
+		gfiMainStore.geojsonFeatures = result
 	}
 
 	const waitMs = computed(() => gfiMainStore.configuration.waitMs || 50)
```
dopenguin: I can't pass `0` like this

src/plugins/gfi/stores/feature.ts
dopenguin, line 86: Would it be more readable if `selection` were a `computed` with a `set` & `get` and `useMultiSelection` receives a ref and a callback so either `debouncedGetFeatureInfo` gets called or `geojsonFeatures` is reset?.

src/plugins/gfi/stores/feature.ts
```diff
@@ -134,7 +144,7 @@ export const useGfiFeatureStore = defineStore('plugins/gfi/feature', () => {
 
 	const exportProperty = computed(() =>
```
dopenguin, line 148: This is typed as `any`

src/plugins/gfi/stores/feature.ts
dopenguin, line 170: I feel like this variable does not yield any benefit and only elongates the code.

dopenguin: I think a verbose implementation seems favourable for future readability

dopenguin, line 199: Why is this only needed if the user no longer selects a feature?

Also, I'd appreaciate a comment with that information.

dopenguin: I think it would be appropriate to add a `ShowTooltip`-type

src/plugins/gfi/stores/feature.ts
```diff
@@ -208,19 +213,21 @@ export const useGfiFeatureStore = defineStore('plugins/gfi/feature', () => {
...
+					(newFeature) => {
+						if (newFeature) {
+							iconMenuStore.openMenuById(PluginId)
+						}
 					}
-				})
+				)
 			}
 		})
 	}
```
dopenguin: I feel like this rather belongs to `useSelectedFeatures`. Then, there would also be no need to watch for changes on `geojsonFeature`.

src/plugins/gfi/stores/list.ts
dopenguin, line 53: This all falls apart if `activeLayers` yields values that are not usable. The return in `string[]` gives one a false sense of safety.
Please add an additional check that the access actually yields the result one wants.

dopenguin: I think it is reasonable to merge the current content of `activeLayers` into this getter and name this one `activeLayers`

dopenguin, line 68: Why not simply use the OpenLayers type?

dopenguin, line 77: I just realized that the return type of `useOlVectorSources` is incorrectly typed and should be `ShallowRef`. The explicit return type is also redundant on that composable.

dopenguin: The filter doesn't do anything, `layer` is always defined.
This is included in the `reduce` suggestion.

dopenguin: This is

a) not needed, as `useOlVectorSources` returns a `shallowRef`
b) gives a false sense of security, as `markRaw` only marks the outer `Record` as raw, not the features.

The removal is included in the `reduce` suggestion

dopenguin:
```suggestion
			activeLayerList.value.reduce<Record<string, { feature: Feature }[]>>(
				(acc, { layerId, layerConfiguration, source }) => {
					acc[layerId] = filterSelectableFeatures(
						getSourceFeatures(
							coreStore.map,
							source,
							configuration.value?.mode || 'visible'
						).filter((feature) => isVisible(feature)),
						layerConfiguration.isSelectable
					).map((feature) => ({ feature }))
					return acc
				},
				{}
			)
```

dopenguin, line 105: The exported `listPage` is never used so this can stay a local `ref`

dopenguin, line 124: Shouldn't this be substracting `-1`?

dopenguin (commented):
I've added some comments to the store which I would've added to #447 but as some logic changed in those files it just seemed fitting to do that here.

🏓 @oeninghe-dataport 

@warm-coolguy Please take a look whether you find any issues with the implementation. It also needs to be checked if these changes properly work if one uses the markers feature without clustering.

src/plugins/gfi/composables/useBindWithCoreHoverSelect.ts
oeninghe-dataport: Yes.

src/plugins/gfi/composables/useBindWithCoreHoverSelect.ts
oeninghe-dataport: As the most known JS functions start with the target on the left side, I'd prefer to keep it in this order.

warm-coolguy: Same. Think lodash, underscore, es-toolkit, Object.assign, ...

src/plugins/gfi/composables/useBindWithCoreHoverSelect.ts
oeninghe-dataport: I expect that the features in the gfi-selected features are sorted in the same way as in the gfi-featureList. The markers store does not guarantee this order.

src/plugins/gfi/composables/useBindWithCoreHoverSelect.ts
oeninghe-dataport: No, there is a GitHub issue (in the sense of: issue of the project GitHub).

But I did it anyway: 213c094645

src/plugins/gfi/composables/useBindWithCoreHoverSelect.ts
oeninghe-dataport: 213c094645

src/plugins/gfi/composables/useSelectedFeatures.ts
oeninghe-dataport: bdcc77d79961e85e66d62276650c1b4dbdc8f87c

src/plugins/gfi/stores/feature.ts
oeninghe-dataport: 41c590ee81d83a6a90c353b1ad6a577101698f39

src/plugins/gfi/stores/feature.ts
oeninghe-dataport: b8c9bdd3c2

src/plugins/gfi/composables/useSelectedFeatures.ts
oeninghe-dataport: c7035c7363

src/plugins/gfi/composables/useSelectedFeatures.ts
oeninghe-dataport: 599ad6e678

src/plugins/gfi/composables/useSelectedFeatures.ts
oeninghe-dataport: The type of `item` is `FeaturePair` which does not have `layerId` --> It is necessary.

src/plugins/gfi/stores/feature.ts
oeninghe-dataport: But now.
393a091255

src/plugins/gfi/stores/feature.ts
oeninghe-dataport: No.

src/plugins/gfi/stores/feature.ts
oeninghe-dataport: 122a56722c

src/plugins/gfi/stores/feature.ts
oeninghe-dataport: We shorten the implementation of `selectedFeatureProperties` by using this shorter name (43 instead of 51 characters). And it is more readable.

oeninghe-dataport: @warm-coolguy What do you think?

warm-coolguy: I think it's readable enough, albeit `es-toolkit` made it a bit more difficult to me.

oeninghe-dataport: It is already documented in the explanation to `coordinateTarget`.

oeninghe-dataport: 215853ec70

src/plugins/gfi/stores/feature.ts
oeninghe-dataport: 87d9aae

src/plugins/gfi/stores/list.ts
oeninghe-dataport: 840859c3f4

oeninghe-dataport: No, IMO this would be too long; it is more readable like this. However, I renamed the getters to be more clear about their usage.

7dc54da02a

oeninghe-dataport: Bc it is not only the OL type, but also `layerId` and `layerConfiguration`.

The verbose code would be sth like:

```
{
    layerId: string;
    layerConfiguration: GfiLayerConfiguration;
    layer: Layer<Source, LayerRenderer<any>, {
        [x: string]: any;
    }>;
}
```

oeninghe-dataport: a5147027b8

oeninghe-dataport: 4f71db93e6

oeninghe-dataport: a) This combination may be relevant w/ pinia as discussed elsewhere
b) True. c100941027

oeninghe-dataport: IMO, `Object.fromEntries(X.map(...` is preferrable over `X.reduce`

oeninghe-dataport: As this is already there and well-documented, we can also keep this.

oeninghe-dataport: As `slice` (see below) and other functions use start inclusive and end exclusive, this seems the useful semantic, that I explained in the documentation of this store value now: 5b5c451d70

warm-coolguy (approved):
🦆 @oeninghe-dataport

PR #926, opened 2026-08-10
Feat/routing marker drag

JoselinGuevaraHoppe:
## Summary

Add a dragging-feature to the markers of an existing route.

- add apiKey from HeiGit to PluginRouting in example/snowbox/index.js
- select start and destination in the route plugin
- select one of the markers and drag them to a new location on the map

### DIFF
--- a/src/plugins/routing/composables/useMarkerLayer.ts
+++ b/src/plugins/routing/composables/useMarkerLayer.ts
@@ -45,13 +45,16 @@ export function useMarkerLayer(
 	})
 
 	const modify = new Modify({ source: markerSource })
-	modify.on('modifyend', () => {
-		route.value = markerSource.getFeatures().map((feature) => {
+	modify.on('modifyend', (evt) => {
+		const allFeatures = markerSource.getFeatures()
+		evt.features.forEach((feature) => {
 			const geometry = feature.getGeometry()
 			if (geometry instanceof Point) {
-				return geometry.getCoordinates()
+				const index = allFeatures.indexOf(feature)
+				if (index !== -1 && index < route.value.length) {
+					route.value[index] = geometry.getCoordinates()
+				}
 			}
-			return []
 		})
 	})
 
--- a/src/plugins/routing/composables/useMarkerLayer.ts
+++ b/src/plugins/routing/composables/useMarkerLayer.ts
@@ -55,4 +55,10 @@ export function useMarkerLayer(
 		})
 	})
 	map.addInteraction(modify)
+
+	map.on('pointermove', function (evt) {
+		const pixel = map.getEventPixel(evt.originalEvent)
+		const hit = map.hasFeatureAtPixel(pixel)
+		map.getTargetElement().style.cursor = hit ? 'pointer' : ''
+	})
 }
--- a/src/plugins/routing/composables/useMarkerLayer.ts
+++ b/src/plugins/routing/composables/useMarkerLayer.ts
@@ -55,7 +55,7 @@ export function useMarkerLayer(
 		})
 	})
 
-	map.on('pointermove', function (evt) {
+	map.on('pointermove', (evt) => {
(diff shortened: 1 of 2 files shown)


src/plugins/routing/composables/useMarkerLayer.ts
```diff
@@ -42,4 +43,22 @@ export function useMarkerLayer(
...
+
+	const modify = new Modify({ source: markerSource })
+	modify.on('modifyend', () => {
+		route.value = markerSource.getFeatures().map((feature) => {
+			const geometry = feature.getGeometry()
+			if (geometry instanceof Point) {
+				return geometry.getCoordinates()
+			}
+			return []
+		})
```
dopenguin: This has two issues
- If I only have one input filled, the second (currently empty) input is removed. Only the modified feature / coordinate should be updated
- If I enter both coordinates, then update the first feature, the coordinate in the second input gets updated as well. This should not be the case.

src/plugins/routing/composables/useMarkerLayer.ts
```diff
@@ -42,4 +43,22 @@ export function useMarkerLayer(
...
+	})
+	map.addInteraction(modify)
+
+	map.on('pointermove', function (evt) {
```
dopenguin: This should be placed above `addInteraction` just like the `modifyend`-event-handler.

dopenguin: A normal function may be usable here, but an arrow function should be favoured.

src/plugins/routing/composables/useMarkerLayer.ts
```diff
@@ -42,4 +43,22 @@ export function useMarkerLayer(
...
+		const pixel = map.getEventPixel(evt.originalEvent)
+		const hit = map.hasFeatureAtPixel(pixel)
+		map.getTargetElement().style.cursor = hit ? 'pointer' : ''
```
dopenguin: The [cursor](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/cursor#keyword) should rather be `grab` if the user is hovering a feature and `grabbing` while modifying the feature.

dopenguin (changes requested):
I should not be able to modify a feature while I am actively adding a new coordinate. It may also cause issues with other plugins. Please take a look at `coreStore.maskInteraction` and how this is handled for `Draw` in `routing` for both.

🏓 @JoselinGuevaraHoppe

src/plugins/routing/composables/useMarkerLayer.ts
JoselinGuevaraHoppe: 50c07092be176d3e8f8fceeaa069ceb7cfe9b690

src/plugins/routing/composables/useMarkerLayer.ts
JoselinGuevaraHoppe: 57a2a46446d18dff4ed504b6b1fe3e6b69e3851e

LivMoeser: 5890ef98d6032e522bb7d6f7fc830cc6d453bda5

src/plugins/routing/composables/useMarkerLayer.ts
JoselinGuevaraHoppe: b368ece4e4a1dcf66d8203cf7c94b771411ba12d

FinnDarge (commented):
- globals: true aus der Vitest-Konfiguration entfernt.
- Die Tests in useMarkerLayer erweitert, sodass auch der tatsächliche Inhalt des VectorSource geprüft wird.
- Tests für die modifyend-Logik ergänzt.
- Tests für die pointermove-Cursorzustände ergänzt (grabbing, grab, leerer Cursor).

FinnDarge:
@dopenguin 🏓

PR #928, opened 2026-08-11
Vue3/routing displays addresses with configured reversegeocoder

### LINKED ISSUE #864: Routing: Display addresses instead of coordinates if ReverseGeocoder is configured
### Description

Use the `reverseGeocoder`  plugin to retrieve addresses for the selected route coordinates and display them in the routing inputs instead of raw coordinates.

The `reverseGeocoder` plugin already provides the required building blocks: it watches configured `coordinateSources` and on change, resolves an address via `reverseGeocode`, then passes the resulting `ReverseGeocoderFeature` (which carries a human-readable title) to a configured `addressTarget` store action. This issue is about wiring the `routing` plugin into that mechanism so waypoint coordinates are shown as addresses.

### Current behavior

The connection between the `routing` and `reverseGeocoder` plugin does not exist thus the coordinates are always displayed.

- The store holds waypoints as raw coordinates in `route` (`Coordinate[]`).
- Each waypoint input in `RoutingInput.ce.vue` binds directly to `route[index]`, so the raw coordinate is what the user sees.
- There is no `coordinateSources`/`addressTarget` link exposed by `routing` for `reverseGeocoder` to consume.

### Expected behavior

Coordinates selected by a user are displayed as addresses if a client has the plugin `reverseGeocoder` configured.

- When `reverseGeocoder` is configured with the `routing` plugin as a coordinate source, selecting/updating a waypoint triggers reverse geocoding.
- The resolved address (`ReverseGeocoderFeature.title`) is stored in the `routing` state and displayed in the corresponding `RoutingInput.ce.vue` input while the coordinate is still used for the routing request.
- If `reverseGeocoder` is not configured, behavior is unchanged and raw coordinates continue to be displayed.
- Loading/failure of reverse geocoding does not block route calculation; on failure the coordinate remains displayed.

### Affected areas

| File | What needs to change |
| --- | --- |
| `src/plugins/routing/store.ts` | Expose per-waypoint coordinate(s) as a `coordinateSource` for `reverseGeocoder`, add state to hold the resolved address labels, and add an `addressTarget` action to receive `ReverseGeocoderFeature` results.
| `src/plugins/routing/components/RoutingInput.ce.vue` | Display the resolved address label instead of the raw `route[index]` value. Still use the raw coordinate as a fallback. |
| `src/plugins/routing/types.ts` | Extend `RoutingPluginOptions` and types if configuration is needed to enable the integration. |
| `src/plugins/reverseGeocoder/store.ts` & `src/plugins/reverseGeocoder/types.ts`| Verify `coordinateSources` support arrays of routing waypoints. Adjust if the current single-coordinate handling is insufficient. |

### Out of scope

A user can not search for addresses.

jedi-of-the-sea:
## Summary

The reverseGeocoder plugin can be configured for the routing plugin in such a way that the coordinates from clicking on the map are used to search an address and that address is displayed in the UI instead of the coordinates.

- start the snowbox client with the configuration from this branch
- open the routing plugin
- click in the input field for start / end and then click on the map
- instead of coordinates, the input field should show an address

The routing itself does only work with a token that was not added to the configuration.

  - [x] Chrome Lighthouse
  - [x] Firefox Accessibility

[issue#864](https://github.com/Dataport/polar/issues/864)


examples/snowbox/index.js
```diff
@@ -360,19 +360,19 @@ addPlugin(
...
 			},
 		],
 		addressTarget: {
-			plugin: 'addressSearch',
-			key: 'selectResult',
+			plugin: 'routing',
+			key: 'selectRouteResult',
 		},
 		zoomTo: 7,
 	})
```
dopenguin: This has the issue that I can not have the pin result in an address and the routing input containing an address. Is there a different approach where both is possible?

Especially as one would want to zoom in after clicking whereas with routing this feels odd

src/plugins/routing/store.ts
```diff
@@ -45,8 +46,17 @@ export const useRoutingStore = defineStore('plugins/routing', () => {
...
 	let draw: Draw | undefined
 
+	const pendingReverseGeocode = ref<{
+		coordinate: Coordinate
+		index: number
+	} | null>(null)
+
 	const _currentlyFocusedInput = ref(-1)
 	const route = ref<Coordinate[]>([[], []])
+	const routeAddressTexts = ref<(string | null)[]>([null, null])
```
dopenguin: Can't this be instantiated with the coordinates instead of `null`?

jedi-of-the-sea:
🏓 @dopenguin

src/plugins/routing/store.ts
```diff
@@ -157,12 +161,19 @@ export const useRoutingStore = defineStore('plugins/routing', () => {
...
-	function addCoordinateToRoute(coordinate: Coordinate) {
-		route.value = route.value.toSpliced(
-			currentlyFocusedInput.value,
-			1,
-			coordinate
-		)
+	async function addCoordinateToRoute(coordinate: Coordinate) {
+		const index = currentlyFocusedInput.value
+		route.value = route.value.toSpliced(index, 1, coordinate)
+		routeAddressTexts.value = routeAddressTexts.value.toSpliced(index, 1, '')
```
dopenguin: `routeAddressTexts` could simply receive `coordinate.join(', ')` so that `routeInput` in `src/plugins/routing/components/RoutingInput.ce.vue` becomes obsolete as `routeAddressTexts` always holds the correct value.

src/plugins/routing/store.ts
```diff
@@ -157,12 +161,19 @@ export const useRoutingStore = defineStore('plugins/routing', () => {
...
+		if (hasReverseGeocoder.value) {
+			const reverseGeocoderStore = coreStore.getPluginStore('reverseGeocoder')
+			const [x, y] = coordinate
+			if (x === undefined || y === undefined) {
+				return
+			}
```
dopenguin: Adding a type cast in `initializeDraw` to `(e.feature.getGeometry() as Point).getCoordinates() as [number, number]` seems to be more straighforward way to accomplish this. We can assume (an we already do) that `coordinate` is `[number, number]`.

The variables become redundant then as well.

src/plugins/routing/store.ts
```diff
@@ -157,12 +161,19 @@ export const useRoutingStore = defineStore('plugins/routing', () => {
...
+			const feature = await reverseGeocoderStore?.reverseGeocode([x, y], false)
+			routeAddressTexts.value[index] = feature?.title ?? ''
```
dopenguin: If the aforementioned change to always set `coordinate` to `routeAddressTexts`, this assignment should only be done if `title` is present on `feature`

src/plugins/routing/store.ts
```diff
@@ -327,6 +341,9 @@ export const useRoutingStore = defineStore('plugins/routing', () => {
 		route.value = remove
 			? route.value.toSpliced(index, 1)
 			: route.value.toSpliced(index, 0, [])
+		routeAddressTexts.value = remove
+			? routeAddressTexts.value.toSpliced(index, 1)
+			: routeAddressTexts.value.toSpliced(index, 0, null)
```
dopenguin: You could add an empty string here and in the instantiation if the aforementioned change is implemented

src/plugins/routing/store.ts
```diff
@@ -336,6 +353,13 @@ export const useRoutingStore = defineStore('plugins/routing', () => {
 		 */
 		route,
 
+		/**
+		 * Reverse-geocoded address labels for each waypoint in {@link route}.
+		 * `null` if no address was resolved (e.g. reverse geocoder not configured).
```
dopenguin: This comment can be updated with the `coordinate` change

src/plugins/routing/store.ts
```diff
@@ -336,6 +353,13 @@ export const useRoutingStore = defineStore('plugins/routing', () => {
...
+		 * @alpha
+		 */
+		routeAddressTexts,
```
dopenguin: Is the name `routeInput` more fitting like you added in the component?

src/plugins/routing/store.ts
```diff
@@ -55,6 +56,9 @@ export const useRoutingStore = defineStore('plugins/routing', () => {
 	const configuration = computed(
 		() => (coreStore.configuration.routing || {}) as RoutingPluginOptions
 	)
+	const hasReverseGeocoder = computed(
+		() => !!coreStore.getPluginStore('reverseGeocoder')
+	)
```
dopenguin: Changing this to `const reverseGeocoderStore = coreStore.getPluginStore('reverseGeocoder')` shaves off a few lines and yields the same result. It also reduces the need to call `getPluginStore` to only once

src/plugins/reverseGeocoder/store.ts
```diff
@@ -93,7 +96,7 @@ export const useReverseGeocoderStore = defineStore(
 					epsg: coreStore.configuration.epsg,
 					signal,
 				})
-				if (configuration.value.addressTarget) {
+				if (configuration.value.addressTarget && forwardToAddressTarget) {
 					passFeatureToTarget(configuration.value.addressTarget, feature)
 				}
 				if (configuration.value.zoomTo) {
```
dopenguin: This should be optionally disabled as well

dopenguin (changes requested):
🏓 @jedi-of-the-sea

src/plugins/routing/store.ts
jedi-of-the-sea: cd59eaeca518ce9bf7dd439ce9562690c860da49

src/plugins/routing/store.ts
jedi-of-the-sea: d826b56f82bbf44e791e1bb19555a6cc617d9bd7

src/plugins/routing/store.ts
jedi-of-the-sea: 9ecb9b9d04501cff4d68cf1d07fd321b3c6e209f

src/plugins/routing/store.ts
jedi-of-the-sea: 9381a8fbe63921a4715567434702340fa0fcf9da

src/plugins/routing/store.ts
jedi-of-the-sea: 4882f9c7b0b88f99cf10b7f6649b714a97c43729

src/plugins/routing/store.ts
jedi-of-the-sea: fcf64f3577e56fed8b0b7ca91123bdb492db4e7f

src/plugins/routing/store.ts
jedi-of-the-sea: f766c0399379e863cdcc1465b14d6896c08a1b7e

src/plugins/reverseGeocoder/store.ts
jedi-of-the-sea: 2117b46b3de595f9770311acdf23b6cf3bc5a76f

jedi-of-the-sea:
🏓 @dopenguin

warm-coolguy (approved):
🚀 @jedi-of-the-sea

PR #932, opened 2026-08-14
Refactor/display feature count on cluster markers

### LINKED ISSUE #778: Display feature count on cluster markers
## Description

Currently, clustered features (core markers) are displayed as a multi-marker; a static SVG showing three stacked pin shapes. There is no indication of **how many features** are contained in a given cluster.

The cluster marker SVG should be replaced with a new design (see Figma) that displays the number of contained features as text within the SVG.

## Current behavior

- The multi-marker is a static SVG without any count indicator (`src/core/utils/markers.ts`, `makeMultiMarker`)
- `getMarkerStyle` only receives a boolean (`multi`), not the actual feature count
- Memoization is based on `MarkerStyle` + the `multi` flag only

## Expected behavior

- A new SVG design for cluster markers that includes a `<text>` element showing the feature count
- The count is dynamically embedded into the SVG

### Affected areas

| File | What needs to change |
|------|---------------------|
| `src/core/utils/markers.ts` | Replace `makeMultiMarker` with new SVG template including count; extend `getMarkerStyle` signature to accept count; adjust memoization to account for count |
| `src/core/utils/map/setupMarkers.ts` | Pass the actual feature count (from `feature.get('features')?.length`) to `getMarkerStyle` at all call sites (layer styling, hover, selection, pointer move) |

### Out of scope

- Configurability of count text styling (color, font size, etc.)
- Changes to the single marker

dMapybara:
## Summary

_Summarization of the changes in a short and concise sentence._

The main change is the display of the number of features in a cluster. This can be configured via index.ts.

`npm i`
`npm run snowbox`

1. Display the concerns in the menu. 
2. Zoom in and out of the map. 
3. Click on individual markers. 
4. And test the hover effect.

  - [ ] Chrome Lighthouse
  - [ ] Firefox Accessibility

  - [ x] **Should not be merged before https://github.com/Dataport/polar/pull/931 ist reviewed.**


dMapybara:
🕊️ @dopenguin

src/core/stores/marker.ts
```diff
@@ -20,6 +20,10 @@ export const useMarkerStore = defineStore('marker', () => {
 		() => (configuration.value?.clusterClickZoom as boolean) || false
 	)
 
+	const displayFeatureCount = computed(
+		() => (configuration.value?.displayFeatureCount as boolean) || false
```
dopenguin:
```suggestion
		() => configuration.value?.displayFeatureCount || false
```

src/core/types/marker.ts
```diff
@@ -1,7 +1,19 @@
...
+import type { Style } from 'ol/style'
 
 export type MarkersIsSelectableFunction = (feature: Feature) => boolean
-
+export type GetMarkerFunction = (
+	style: MarkerStyle,
+	count: number,
+	displayFeatureCount: boolean
+) => Style
+export type GetSVGConfigFunction = (digits: string) => MarkerSVGConfig
```
dopenguin, line 10: The type is redundant as the information can be inferred.

src/core/types/marker.ts
```diff
@@ -1,7 +1,19 @@
...
+export type GetTextPositionFunction = (path: string) => TextPosition
+export type PinShape = 'circle' | 'pill'
```
dopenguin, line 12: As this is not used anywhere else, this can just directly be set on `MarkerSVGConfig`.

Also, do you think there will be a third option? Otherwise, using a `boolean` with a different name seems easier.

src/core/types/marker.ts
```diff
@@ -33,7 +45,7 @@ export interface MarkerStyle {
 	/**
 	 * `width` and `height` of the `<svg>`-cluster-marker.
 	 *
-	 * @defaultValue `[40, 36]`
+	 * @defaultValue `[40 * 2, 36 * 2]`
```
dopenguin, line 48: Why not 80 and 72? If changed, this should be done on both types and all other instances.

src/core/types/marker.ts
```diff
@@ -62,6 +74,72 @@ export interface MarkerStyle {
 	 * @defaultValue `'2'`
 	 */
 	strokeWidth: string | number
+
+	/**
+	 * Text to display on the marker..
+	 *
+	 * @defaultValue `''`
+	 */
+	displayedText?: string | number
```
dopenguin: Currently, it is always a number, so the type and name should reflect this.
The wrapping with `String` seems to be redundant as well.

As a completely different marker is being shown if this is not set, I suppose the `@defaultValue` can be dropped.

src/core/types/marker.ts
```diff
@@ -62,6 +74,72 @@ export interface MarkerStyle {
...
+
+	/**
+	 * The definitions for the marker (e.g. shadow patterns with image data).
+	 */
+	defs: string
+
+	/**
+	 * Calculates the x and y coordinates for the text position within the marker.
+	 * The calculation is based on the provided SVG path
+	 * @throws  Error If the provided path does not match the expected marker pattern.
```
dopenguin:
```suggestion
	 * @throws Error If the provided path does not match the expected marker pattern.
```

src/core/types/marker.ts
```diff
@@ -62,6 +74,72 @@ export interface MarkerStyle {
 	 * @defaultValue `'2'`
 	 */
 	strokeWidth: string | number
+
+	/**
+	 * Text to display on the marker..
```
dopenguin:
```suggestion
	 * Text to display on the marker.
```

src/core/types/marker.ts
```diff
@@ -62,6 +74,72 @@ export interface MarkerStyle {
...
+	/**
+	 * Calculates the x and y coordinates for the text position within the marker.
+	 * The calculation is based on the provided SVG path
+	 * @throws  Error If the provided path does not match the expected marker pattern.
+	 * @returns The x and y coordinates for the text position within the marker.
+	 */
+	getTextPosition: GetTextPositionFunction
+
+	/**
+	 * The shape of the marker
```
dopenguin:
```suggestion
	 * The shape of the marker.
```

src/core/types/marker.ts
```diff
@@ -62,6 +74,72 @@ export interface MarkerStyle {
...
+	 * The shadow path for the main/front marker layer.
+	 */
+	shadowPath: string
+
+	/** The outer shape of the marker. */
+	shapePath: string
+
+	/**
+	 * The SVG shadow paths for the stacked marker layers.
+	 */
```
dopenguin, line 122: This comment (just like later for `stackedShape1`) only belongs to `stackedShadow1`.
This could be fixed with the object signature being different e.g. with `stackedShadow: [string, string]`; the same would apply for `stackedShape` and `stackedTip`

src/core/types/marker.ts
```diff
@@ -62,6 +74,72 @@ export interface MarkerStyle {
...
+	stackedShadow1: string
+	stackedShadow2: string
+
+	/**
+	 * The SVG paths for the stacked markershape.
```
dopenguin:
```suggestion
	 * The SVG paths for the stacked shape of the marker.
```

src/core/types/marker.ts
```diff
@@ -62,6 +74,72 @@ export interface MarkerStyle {
...
+	 */
+	stackedShape1: string
+	stackedShape2: string
+	stackedTip1: string
+	stackedTip2: string
+
+	readonly textPosition: TextPosition
```
dopenguin, line 134: How come this is the only `readonly` property? Is this relevan in some way?

src/core/utils/map/setupMarkers.ts
```diff
@@ -208,20 +209,19 @@ export function setupMarkers(map: Map) {
...
+				const featureCount = feature.get('features')?.length
 				const style = getMarkerStyle(
 					getLayerConfiguration(feature.get('_polarLayerId') as string)
 						.hoverStyle,
-					isMultiFeature
+					featureCount,
+					store.displayFeatureCount
 				)
-				store.hovered.setStyle(style)
+				feature.setStyle(style)
```
dopenguin, line 234: What does this change do? I feel like I've seen this before

src/core/utils/map/setupMarkers.ts
```diff
@@ -274,31 +274,25 @@ function mapPointerMove({ map, pixel }: MapBrowserEvent) {
...
+		store.hovered = null
 		return
 	}
 	setLayerId(map, feature)
 	const layerConfiguration = getLayerConfiguration(
 		feature.get('_polarLayerId') as string
 	)
 	if (!layerConfiguration.isSelectable(feature)) {
+		store.hovered = null
 		return
```
dopenguin, line 290: Please elaborate

src/core/utils/markers.ts
```diff
@@ -23,78 +23,191 @@ const getImagePattern = (fill: MarkerStyle['fill']) =>
...
- * @masterportal/masterportalapi/public/marker.svg. */
-
-const makeMarker = ({ fill, size, stroke, strokeWidth }: MarkerStyle) =>
-	`${prefix}${encodeURIComponent(`
-<svg width="${size[0]}" height="${
-		size[1]
-	}" viewBox="0 0 30 43" xmlns="http://www.w3.org/2000/svg">
+const makeMarker = ({ fill, size, stroke }: MarkerStyle) =>
+	`${prefix}${encodeSVG(`
+<svg fill="none" width="${size[0]}" height="${size[1]}" viewBox="${circlePin.viewBox}" xmlns="http://www.w3.org/2000/svg">
```
dopenguin, line 28: Why `fill="none"`?

src/core/utils/markers.ts
```diff
@@ -23,78 +23,191 @@ const getImagePattern = (fill: MarkerStyle['fill']) =>
...
-	`${prefix}${encodeURIComponent(`
-<svg width="${clusterSize[0]}" height="${
-		clusterSize[1]
-	}" viewBox="0 0 30 43" xmlns="http://www.w3.org/2000/svg">
+const makeMultiMarker = (
+	{ fill, clusterSize, stroke, displayedText }: MarkerStyle,
+	displayFeatureCount: boolean,
+	svgConfig: MarkerSVGConfig
+) => {
+	return svgConfig.pinShape === 'circle'
```
dopenguin, line 50: If a `const` is used, then one should gently drop the `return`

src/core/utils/markers.ts
```diff
@@ -23,78 +23,191 @@ const getImagePattern = (fill: MarkerStyle['fill']) =>
...
+ * This memoization function optimises this issue by reusing styles.
+ * */
+const memoizeStyle = (getMarker: GetMarkerFunction): GetMarkerFunction => {
+	const memoizedCountStyle = memoCountStyle(getMarker)
+	const memoizedStyle = memoStyle(getMarker)
+	return (style, count, displayFeatureCount) =>
+		displayFeatureCount
+			? memoizedCountStyle(style, count, displayFeatureCount)
+			: memoizedStyle(style, count, displayFeatureCount)
+}
```
dopenguin:
```suggestion
const memoizeStyle =
	(getMarker: GetMarkerFunction): GetMarkerFunction =>
	(style, count, displayFeatureCount) =>
		displayFeatureCount
			? memoCountStyle(getMarker)(style, count, displayFeatureCount)
			: memoStyle(getMarker)(style, count, displayFeatureCount)
```

Or did you already test this and this caused issues?

src/core/utils/markers.ts
```diff
@@ -23,78 +23,191 @@ const getImagePattern = (fill: MarkerStyle['fill']) =>
...
-			)
+		warnMemoLeak(cache.size)
+		return markerStyle
+	}
+}
+
+const memoCountStyle = (getMarker: GetMarkerFunction): GetMarkerFunction => {
+	const countCache = new Map<number, Map<MarkerStyle, Style>>()
+	return (style, count, displayFeatureCount) => {
+		// vielleicht auslagern und ind warnMeoLeak einbinden?
```
dopenguin: 🤔

src/core/utils/markers.ts
```diff
@@ -23,78 +23,191 @@ const getImagePattern = (fill: MarkerStyle['fill']) =>
...
-			console.warn(
-				`1000+ styles have been created. This is possibly a memory leak. Please mind that the methods exported by this module are memoized. You *may* be calling the methods with constantly newly generated objects, or maybe there's just a lot of styles.`
-			)
+		warnMemoLeak(cache.size)
+		return markerStyle
+	}
+}
+
+const memoCountStyle = (getMarker: GetMarkerFunction): GetMarkerFunction => {
+	const countCache = new Map<number, Map<MarkerStyle, Style>>()
```
dopenguin: > Yo dawg, I heard you like `Maps`, so I put a `Map` in your `Map` which resides in your `Map`-client and was committed by d`Map`ibara

TODO

src/core/utils/markers.ts
```diff
@@ -23,78 +23,191 @@ const getImagePattern = (fill: MarkerStyle['fill']) =>
...
+	return (style, count, displayFeatureCount) => {
+		// vielleicht auslagern und ind warnMeoLeak einbinden?
+		const getTotalCachedStyles = (
+			countCache: Map<number, Map<MarkerStyle, Style>>
+		): number => {
```
dopenguin:
```suggestion
		) => {
```

dopenguin (changes requested):
Partial review, looking at `src/core/utils/markerSVG.ts` up next

🏓 @dMapybara

src/core/utils/markers.ts
dMapybara: I copied this code from the SVG that the UX Designer provided.

src/core/utils/markerSVG.ts
dopenguin, line 1: I feel like `markers` should receive its own folder like `export` with `src/core/utils/markers.ts` being the index-file.

Then, this file may be split in multiple files as well such that all the different `MarkerSVGConfig`s receive their own files.

src/core/utils/markerSVG.ts
```diff
@@ -0,0 +1,291 @@
+/**
+ * This file contains the SVG-informations
+ * the 'path' / 'viewBox' / 'position of the text element' of the markers
+ * it will be used in the marker.ts file to create the marker icons
+ * for the map.
+ */
```
dopenguin, line 5: I feel like this comment, if change slightly, rather belongs to `getSVGConfig`

src/core/utils/markerSVG.ts
```diff
@@ -0,0 +1,291 @@
...
+			return fourDigitsPin
+		case 5:
+			return fiveDigitsPin
+		default:
+			throw new Error('Unsupported number of digits')
+	}
+}
+
+/**
+ * Calculates the optically centered text position for a pill SVG path.
```
dopenguin:
```suggestion
 * Calculates the visually centered text position for a pill SVG path.
```
Albeit `optically` ain't wrong, `visually` is more common

src/core/utils/markerSVG.ts
```diff
@@ -0,0 +1,291 @@
...
+ * Expected paths in the scheme: M[x] [y] C... [xLeftCap] [yTop] h[width] C... S... [xRightCap] [yBottom] h-[width] C... Z
```
dopenguin: Please adjust this so this parth of the comment belongs to the `@param`

src/core/utils/markerSVG.ts
```diff
@@ -0,0 +1,291 @@
...
+ */
+export const getPillTextPosition: GetTextPositionFunction = (
```
dopenguin:
```suggestion
const getPillTextPosition: GetTextPositionFunction = (
```
Only used locally

src/core/utils/markerSVG.ts
```diff
@@ -0,0 +1,291 @@
...
+		x: (xLeftCap + xRightCap) / 2,
+		y: (yTop + yBottom) / 2,
+	}
+}
+
+/**
+ * Calculates the optically centered text position for a round SVG path (circle with tip).
+ * Expected paths in the scheme: M[x0] [y0]C[x1] [y1] [x2] [y2] [x3] [y3]S[x4] [y4] [x5] [y5] [x6] [y6] [x7] [y7]Z
+ */
+export const getCircleCenterPosition: GetTextPositionFunction = (
```
dopenguin:
```suggestion
const getCircleCenterPosition: GetTextPositionFunction = (
```
Only used locally

src/core/utils/markerSVG.ts
```diff
@@ -0,0 +1,291 @@
...
+
+	return {
+		x: (xLeftCap + xRightCap) / 2,
+		y: (yTop + yBottom) / 2,
+	}
+}
+
+/**
+ * Calculates the optically centered text position for a round SVG path (circle with tip).
+ * Expected paths in the scheme: M[x0] [y0]C[x1] [y1] [x2] [y2] [x3] [y3]S[x4] [y4] [x5] [y5] [x6] [y6] [x7] [y7]Z
```
dopenguin: See https://github.com/Dataport/polar/pull/932/changes#r3842218520 and https://github.com/Dataport/polar/pull/932/changes#r3842225103

src/core/utils/markerSVG.ts
```diff
@@ -0,0 +1,291 @@
...
+
+	const [, leftXStr, centerYStr] = mMatch as [string, string, string]
+	const [, rightXStr] = sMatch as [string, string]
+
+	return {
+		x: (parseFloat(leftXStr) + parseFloat(rightXStr)) / 2,
+		y: parseFloat(centerYStr),
+	}
+}
+// The paths always follow the scheme: scheme: M[x0] [y0]C[x1] [y1] [x2] [y2] [x3] [y3]S[x4] [y4] [x5] [y5] [x6] [y6] [x7] [y7]Z
```
dopenguin, line 94: I would have no idea what to do with this information tbh. If I were, I'd expect this comment to be set somewhere on `MarkerSVGConfig`. Especially as it is also added twice in this file

src/core/types/marker.ts
dMapybara: Because 40 and 36 were the smallest values that worked well, I thought it would make sense to keep them and scale them up.

src/core/utils/markerSVG.ts
dMapybara: If the path in the SVG deviates from this scheme, the marker may no longer be displayed correctly. Furthermore, the text positivity is no longer determined correctly.

It ist relevant if you want to implement a different SVG as a Marker - you would have to adjust the calculation as well.

But this comment should be above the calculating-function. 
Therefore I would suggest to move it there.

src/core/types/marker.ts
dMapybara: The readonly property allows for more relaxed access to the text position later: svgConfig.textPosition. And even if the calculation for each svgConfig is different, or even a different function is behind it, the call remains reliably the same.

src/core/utils/map/setupMarkers.ts
dMapybara: This is the cleanup part of the Watcher. If the hovered feature changes, the individual style of the previous feature is removed. Unless there is no "pre-selected" feature.

src/core/types/marker.ts
dMapybara: Typing ensures that this function always returns an MarkerSVGConfig-object. Should the function typing be completely removed? Or do you have something different in mind?

src/core/types/marker.ts
dMapybara: Yes I thought, maybe there will be more different markers, like rectangle ones or square ones.

src/core/types/marker.ts
dMapybara: clusterCount?: number
Agreed. I wil change the name and type in favor for more clarity.

src/core/types/marker.ts
dMapybara: Agreed

src/core/types/marker.ts
dMapybara: Agreed.

src/core/types/marker.ts
dMapybara: Agreed.

src/core/utils/markerSVG.ts
dMapybara: Agreed, sounds better ;)

src/core/utils/markerSVG.ts
dMapybara: Agreed.

src/core/utils/markerSVG.ts
dMapybara: Agreed.

src/core/utils/markerSVG.ts
dMapybara: done.

src/core/utils/markerSVG.ts
dMapybara: done.

src/core/utils/markers.ts
dMapybara: so you would rather implement an if-statement? Oder do you have something different in mind?
If we put the decisioin wether to display a circle or pill in the returned string, we will lose a huge amount of performance.

src/core/utils/map/setupMarkers.ts
dMapybara: This is implemented to help the performance while one is hovering over the map.

First it is checked that the selected feature keeps it stlye when hovered over.

Then it is checked whether there was a different marker hovered over which is not selected. In that case the hover-style will be removed and the marker will regain its normal style.

This should prevent, that an old hover-style can linger in the store.

src/core/types/marker.ts
dMapybara: Agreed.

src/core/types/marker.ts
dMapybara: That would be true. The implementation needs every string to be accessable at it's own. In order to build the svg in marker.ts it is neseccary to place them in the correct order. The implementation now also allows to "read" what is happening in lines 51-84.
We would lose this readability in favor to correct that comment.
And I could imagine that it could cause confusion in the future.
There would be an array and it has to have 2 members, even when there is no stacked shadow like:
stackedShadow = ['', '']
On the other hand, this would make the svgConfig a bit leaner.

src/core/utils/markers.ts
dMapybara: Why would wen want tot drop the return-type? As soon as someone would work at this function, it is clear, that a number is expected and changing the return-type could probably effect other parts of the code.

src/core/utils/markerSVG.ts
dMapybara: I adjusted the comment and moved a part of it to the svgConfig-interface and the getSVGConfig

dMapybara:
🕊️ @warm-coolguy

src/core/types/marker.ts
warm-coolguy: dopenguin likes them completely removed while they're inferrable. I personally disagree with the same argument you provided: It will, on later changes, make sure that there's no accidental type change due to an oversight.

IMO you may decide either way. If you keep it, you may just resolve this conversation. 🎩

src/core/types/marker.ts
warm-coolguy: I see no need to change this, any change would be arbitrary. 🎩

src/core/types/marker.ts
warm-coolguy: Makes sense to me. Please add that reason to the comment so that the next person reading knows; it's a nice information to have for future modification/configurability.

src/core/types/marker.ts
warm-coolguy: https://github.com/Dataport/polar/pull/932/changes/0c69b4efc79c40dddf8c834cc260367fa17fadc4

src/core/types/marker.ts
warm-coolguy: I think it's plausible enough to just copy-paste the comment to each field singularized so that tsdoc can correctly document it. Should you not like that, I furthermore have this suggestion: [proposal.patch](https://github.com/user-attachments/files/31997954/proposal.patch)

It does add a practically unneeded type, but I find synonyms to be nice for readability and deduplication.

Either solution would be fine with me. Or maybe you actually want to go `['', '']` or find yet another way to solve this. The choice is yours!

src/core/types/marker.ts
warm-coolguy: I've removed the `readonly` for testing what happens, and nothing changed. `npm run tsc` exits without errors, too. It may be that this is no longer required due to other changes; it may also be that I just don't get it! Proposal: We ☎️ and inspect this.

src/core/utils/map/setupMarkers.ts
warm-coolguy: I would like to propose this reordering:

```
	if (!feature) {
		store.hovered = null
		return
	}

	if (feature === toRaw(store.hovered)) {
		return
	}

	if (feature === toRaw(store.selected) || feature instanceof RenderFeature) {
		return
	}
```

If we're checking first whether feature is falsy, the follow-up checks can be skipped. The gains are minimal, but since the mouse is moved a lot, maybe this will make a difference in a heavy-load scenario? Might also be purely theoretical ... hence, 🎩 on this.

src/core/utils/markers.ts
warm-coolguy: For fun, and by fun I mean science since messing around and writing down the results qualifies, I set the stroke to `''`.

Screenshot 2026-09-09 110957

Then, I removed `fill="none"` on the root node.

Screenshot 2026-09-09 111108

It improved a little! ... maybe? The other one is a little futuristic, I'd say! Looks like many shapes just default to black fill when an empty string is given: https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Attribute/fill#circle

However, stroke is not intended to be optional, so I consider it fine to keep `fill="none"` if that's what we were sent. Especially since any new changes/deliveries will contain it again, anyway, and I don't think this should lead to any headache down the line.

🎩

src/core/utils/markers.ts
warm-coolguy: He probably wants this pattern

```
const f = () => a ? 'b' : 'c'
```

instead of

```
const f = () => { return a ? 'b' : 'c' }
```
.

🎩 since it doesn't really matter.

src/core/utils/markers.ts
warm-coolguy: 🏌️﹏෴﹏෴﹏෴﹏෴﹏෴﹏෴﹏⛳, but also 🎩.

src/core/utils/markers.ts
warm-coolguy: I agree @dMapybara.

However, the function no longer exists anyway. 🎩

src/core/utils/markerSVG.ts
warm-coolguy: I agree since `markers.ts` and `markerSVG.ts` belong together, i.e. `markersSVG.ts` isn't used outside of `markers.ts` and probably never will be.

I am indifferent about splitting the files further since `marker.ts` isn't *that* long and `markerSVG.ts` has a low complexity.

src/core/utils/markerSVG.ts
warm-coolguy: There's a very accepting dopenguin thumb on that comment, so I assume that's the solution agreed upon.

src/core/utils/markers.ts
warm-coolguy: The comment is no longer present. Whatever's been decided has been decided. 🎩

warm-coolguy (changes requested):
🐝 @dMapybara 

Worker bee @warm-coolguy dances the following information to worker bee @dMapybara:

* There are unresolved conflicts regarding the target branch that have to be resolved in order to progress.
* I have proceeded to close a lot of threads that seem resolved. On those with unclear status, a comment has been added regarding next plausible steps.
* I remembered that you remembered me that worker bee @dopenguin danced us a message regarding a greater cluster distance and smaller markers. I have prepared the following proposal: [proposal.patch](https://github.com/user-attachments/files/32005329/proposal.patch) In this, the snowbox has been configured to have markers slightly smaller than before, and a cluster distance way greater than before. However, these values are arbitrary. Please adjust as you deem fit. I assume the goal is to have a nice snowbox setup for the Meldemichel report clusters where they don't overlap too much.

[/dance]

src/core/types/marker.ts
dopenguin: I proclaim YAGNI

src/core/types/marker.ts
dopenguin: The type is still superfluous but may stay with the options

src/core/utils/map/setupMarkers.ts
dopenguin: I think it is a good idea to check whether this approach works, @warm-coolguy

PR #939, opened 2026-08-21
feat(core)!: rework markers for cluster usage [core-only]

oeninghe-dataport:
The current implementation used selectedBaseFeature to keep track of which feature to select when zooming in and out. This did not work properly, and is not intuitive either.

Therefore, we split the (meant) hovered or selected feature and its (rendered) hovered or selected cluster. If the layer does not use clustering, cluster and feature are equivalent.

This has the additional advantage that isSelectable does no longer consider the cluster, but an actual feature, such that the implementation is automatically cluster-agnostical now.

This allows the GFI plugin to adapt this new API. When bindWithCoreHoverSelect is used now, the synchronization with the markers is now truly bi-directional and wrapped in a single, easy-to-understand composable. To allow the synchronization from the details view back to the marker store, the selected feature is moved to the GFI main store as a shared state between list view and details view.

One could possibly split these changes into two or three (still atomical) commits. However, the changes only make sense in the context of this refactoring idea, and are therefore combined to one commit on purpose.

-----

This is merged onto the next branch.
When this is accepted, it will (hopefully) solve a few threads in https://github.com/Dataport/polar/pull/447


src/lib/findLayer.ts
dopenguin, line 1: As I anticipate that this PR will have some discussion (not as extensive as #447 I suppose), this can easily be added separately.
You can also use it then in the multiple places `map.getLayers` is used for this purpose.

src/core/utils/map/setupMarkers.ts
```diff
@@ -1,362 +0,0 @@
...
-	if (store.clusterClickZoom && isMultiFeature && isMaxZoom) {
-		resolveClusterClick(map, feature)
-		return
-	}
-
-	store.hovered?.setStyle(undefined)
-	store.hovered = null
-	updateSelection(map, feature, true)
-
-	if (store.callOnMapSelect) {
```
dopenguin, line 332: Why is `callOnMapSelect` dropped completely?

src/core/stores/marker.ts
```diff
@@ -1,41 +1,289 @@
...
+			return
+		}
+
+		event.stopPropagation()
+		lastClickEvent.value = event.originalEvent
+
+		if (
+			clusterClickZoom.value &&
+			clusterFeatures.length > 1 &&
+			!isMaxZoom.value
```
dopenguin, line 211: This was previously `store.clusterClickZoom && isMultiFeature && isMaxZoom` which is probably why it no longer recenters on a selected feature.
Beware that the behaviour should be in line on how this is handled on `main` meaning it only recenters on selection **not** on zoom

src/core/stores/marker.ts
```diff
@@ -1,41 +1,289 @@
...
-	const selected = shallowRef<Feature | null>(null)
-	const selectedCoordinates = computed(() =>
-		selected.value === null
-			? null
-			: (selected.value.getGeometry() as Point).getCoordinates()
+	const isMaxZoom = computed(
+		() => mainStore.map.getView().getMaxZoom() === mainStore.zoom
+	)
+
+	const lastClickEvent = ref<PointerEvent | KeyboardEvent | WheelEvent | null>(
```
dopenguin: There is no need for this to be a `ref` as far as I can tell

src/core/stores/marker.ts
```diff
@@ -1,41 +1,289 @@
...
+					getStyle(feature as Feature, layer.get('id'), false, false)
+				)
+				teardownCallbacks.push(() => {
+					;(layer as VectorLayer).setStyle(undefined)
+				})
+			})
+
+		listenerKeys.push(mainStore.map.on('pointermove', mapPointerMove))
+		listenerKeys.push(mainStore.map.on('click', mapClick))
+		listenerKeys.push(mainStore.map.on('singleclick', mapSingleClick))
```
dopenguin, line 292: This is now missing its comment

src/core/composables/useClusterMarker.ts
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	watch(feature, updateCluster, { immediate: true })
+
+	watch(
+		() => mainStore.zoom,
+		() => {
+			updateCluster(feature.value)
+		}
+	)
```
dopenguin, line 68: These watchers are never disposed.

src/lib/findLayer.ts
oeninghe-dataport: As requested -> https://github.com/Dataport/polar/pull/948

src/core/utils/map/setupMarkers.ts
```diff
@@ -1,356 +0,0 @@
...
-					getLayerConfiguration(feature.get('_polarLayerId') as string)
-						.hoverStyle,
-					isMultiFeature
-				)
-				store.hovered.setStyle(style)
-			}
-		}
-	)
-
-	map.on('moveend', mapMoveEnd)
```
dopenguin: Why is this dropped?

src/core/stores/index.ts
```diff
@@ -246,18 +248,79 @@ export const useCoreStore = defineStore('core', () => {
 		moveHandleTop: computed(() => moveHandleStore.top),
 
 		/**
-		 * Currently hovered marker feature or null.
+		 * Currently hovered marker feature or `null`.
+		 * You may not set this to a cluster feature.
```
dopenguin:
```suggestion
		 * You may not set this to a cluster.
```

src/core/stores/index.ts
```diff
@@ -246,18 +248,79 @@ export const useCoreStore = defineStore('core', () => {
...
+		 */
+		hoveredFeature: markerStoreRefs.hoveredFeature,
+
+		/**
+		 * Feature that is hovered by the user with a marker.
+		 * If the layer does not use clustering, this is the same as {@link hoveredFeature}.
+		 * Otherwise, this is the cluster feature that contains the {@link hoveredFeature}.
```
dopenguin:
```suggestion
		 * Otherwise, this is the cluster that contains the {@link hoveredFeature}.
```

src/core/stores/index.ts
```diff
@@ -246,18 +248,79 @@ export const useCoreStore = defineStore('core', () => {
...
+		/**
+		 * Coordinates that were hovered by the user with a marker.
+		 *
+		 * @readonly
+		 */
+		hoveredCoordinates: computed(() => markerStore.hoveredCoordinates),
+
+		/**
+		 * Currently selected marker feature or `null`.
+		 * You may not set this to a cluster feature.
```
dopenguin:
```suggestion
		 * You may not set this to a cluster.
```

src/core/stores/index.ts
```diff
@@ -246,18 +248,79 @@ export const useCoreStore = defineStore('core', () => {
...
+		 * Setting this value to a cluster feature has no effect, however, this may change in the future.
```
dopenguin:
```suggestion
		 * Setting this value to a cluster has no effect, however, this may change in the future.
```

src/core/stores/index.ts
```diff
@@ -246,18 +248,79 @@ export const useCoreStore = defineStore('core', () => {
...
+					return
+				}
+				markerStore.selectedFeature = feature
+			},
+		}),
+
+		/**
+		 * Feature that is marked as selected on the map.
+		 * If the layer does not use clustering, this is the same as {@link selectedFeature}.
+		 * Otherwise, this is the cluster feature that contains the {@link selectedFeature}.
```
dopenguin:
```suggestion
		 * Otherwise, this is the cluster that contains the {@link selectedFeature}.
```

dopenguin (changes requested):
Still looking at the refactor, but you can already give it a go

🏓 @oeninghe-dataport

src/core/stores/marker.ts
```diff
@@ -1,41 +1,289 @@
...
-		typeof configuration.value?.callOnMapSelect === 'function'
-			? (configuration.value.callOnMapSelect as CallOnMapSelect)
-			: null
+	const layers = computed(() =>
+		Object.fromEntries(
+			configuration.value?.layers
+				.map((layer) => toMerged(layerDefaults, layer))
+				.map((layer) => [layer.id, layer as MarkerLayer]) ?? []
+		)
+	)
```
dopenguin: Like this, `layers` reactively derives from `configuration` and produces new style objects on every recompute because of `toMerged`.
Since `configuration` currently never changes at runtime, the `computed` only ever runs once anyway. This means the reactivity is unnecessary to the point of being misleading. It also couples the per-object-identity memoized marker styles to a reactive construct.

Please derive the object once like before. That way the cache-miss risk doesn't exist by construction in the first place.

src/core/composables/useClusterMarker.ts
```diff
@@ -0,0 +1,71 @@
...
+
+		if (cluster.value) {
+			cluster.value.setStyle(style(true, cluster.value))
+		}
+	}
+
+	watch(feature, updateCluster, { immediate: true })
+
+	watch(
+		() => mainStore.zoom,
```
dopenguin, line 64: This is unconditionally registered for both `hovered` and `selected` and eventually triggers `getCluster` which scans the source to `find` the cluster.
As it is instantiated for both, this is run twice on every zoom change. With possibly 25k features this would mean O(n) x 2. Previously, this was effectively only the case for `selected` through a `moveend` listener and only when `zoom !== lastZoom`. 

Please restrict the zoom-driven re-resolution to `selected` or at least reintroduce a zoom guard to throttle it.

src/core/composables/useClusterMarker.ts
```diff
@@ -0,0 +1,71 @@
...
+				? null
+				: markRaw(getClusterSafe(mainStore.map, newFeature))
+
+		if (cluster.value === newCluster) {
+			return
+		}
+
+		if (cluster.value) {
+			cluster.value.setStyle(style(false, cluster.value))
+		}
```
dopenguin, line 52: This created a fixed style onto the feature and decouples it from the layer style function.

It's rather fitting to reset it like before with `setStyle(undefined)` so the layer style function takes over again.

src/core/stores/marker.ts
```diff
@@ -1,41 +1,289 @@
...
+		const clusterFeatures = feature.get('features') || [feature]
+		const selectableClusterFeatures = clusterFeatures.filter(
+			layerConfiguration.isSelectable
+		)
+		if (selectableClusterFeatures.length === 0) {
+			return
+		}
+
+		hoveredFeature.value = selectableClusterFeatures[0]
+	}
```
dopenguin, line 160: This sets `hoveredFeature` unconditionally even if the hit feature is already `selected`.
As a consequence `hoveredCluster` can equal `selectedCluster` at the same time, both composables call `setStyle` on the same feature and the visual effect is only saved by the order in `getStyle`.

Previously, there was an early-out for "already selected." Together with https://github.com/Dataport/polar/pull/939/changes#r3864901396 this yields a genuine `hover`×`select` interaction and not just cosmetics.

Please restore the early-out when the target `feature`/`cluster` is already selected.

src/core/stores/marker.ts
```diff
@@ -1,41 +1,289 @@
...
+				.map((layer) => [layer.id, layer as MarkerLayer]) ?? []
+		)
+	)
+
+	function getLayerConfiguration(id: string) {
+		return layers.value[id] as MarkerLayer
+	}
+
+	function layerFilter(layer: BaseLayer) {
+		return layer.get('id') in layers.value
```
dopenguin:
```suggestion
		return Object.hasOwn(layers.value, layer.get('id'))
```
As `id`s can be any String, they could also be `toString`

src/core/composables/useClusterMarker.ts
```diff
@@ -0,0 +1,71 @@
...
+
+import { useMainStore } from '../stores/main'
+
+function getClusterSafe(map: Map, feature: Feature) {
+	const layerId = feature.get('_polarLayerId') as string
+	// @ts-expect-error | Found layers always have a source and getDistance is defined on cluster sources.
+	return typeof findLayer(map, layerId)?.getSource().getDistance === 'function'
+		? getCluster(map, feature, '_polarLayerId')
+		: feature
+}
```
dopenguin, line 19: This along with [some other functionality](https://github.com/Dataport/polar/pull/939/changes#r3864889817) mentioned below carries `selected`-oriented weight into the `hover` path. The two instances also cross-reference each other's cluster non-reactively in the style callbacks effectively making them not completely separated.

I'd keep the deduplication as a pure shared helper for e.g. cluster resolution and coordinates but wire `hover` and `select` in separate, focused composables/utilities so each only pulls in what it needs.
That also makes the missing ["already selected"](https://github.com/Dataport/polar/pull/939/changes#r3864928320) early-out easier to reason about.

dopenguin (changes requested):
Overall good improvements with some creases to iron out!

🏓 @oeninghe-dataport

src/core/stores/index.ts
warm-coolguy: Is this really about a clustered feature or about a cluster feature? From my understanding, a cluster feature is a feature that encompasses multiple features, while a clustered feature is a feature clustered in a cluster feature. Try to say that fast ten times.

src/core/utils/map/setupMarkers.ts
oeninghe-dataport: 1) The same behaviour can be achieved by watching the GFI store, so this is superfluous.

2) `callOnMapSelect` is the successor of `dispatchOnMapSelect` from POLAR@2.
This was used for a single purpose: The iconMenu opens the GFI plugin on selection of a GFI-able feature.
This behaviour is already preserved with https://github.com/Dataport/polar/blob/75c3c6d928019483258fae78241bec8485d08cce/src/plugins/gfi/stores/feature.ts#L259-L275

src/core/composables/useClusterMarker.ts
oeninghe-dataport: Watchers are automatically disposed on scope disposal. As these watchers are at the top-level of the composable (and included at the top-level of the marker store), they get auto-disposed just as watchers at the top-level of stores.

src/core/stores/index.ts
oeninghe-dataport: image

I'd probably assume that a "clustered feature" could be a feature that is grouped with other features to a cluster (which is fine to set), while a "cluster feature" is the feature that represents the cluster itself.

Therefore, I chose the wording "cluster feature".

src/core/stores/index.ts
oeninghe-dataport: https://github.com/Dataport/polar/pull/939#discussion_r3869518209

src/core/stores/index.ts
oeninghe-dataport: https://github.com/Dataport/polar/pull/939#discussion_r3869518209

src/core/stores/index.ts
oeninghe-dataport: https://github.com/Dataport/polar/pull/939#discussion_r3869518209

src/core/stores/index.ts
oeninghe-dataport: Oh, just saw your comment too late. https://github.com/Dataport/polar/pull/939#discussion_r3869518209 was also trying to say that, @warm-coolguy

src/core/stores/marker.ts
oeninghe-dataport: - If we do not allow `configuration` changes, the risk does not exist either way.
- If we'd allow `configuration` changes, deriving this once would imply the configuration update is not considered.
  That a `configuration` update is expensive (in terms of runtime) is fine for now.

As a result, having this reactive has more advantages than disadvantages IMO.

We already use this pattern in plugins, and I still see no reason to deviate here. https://github.com/Dataport/polar/blob/75c3c6d928019483258fae78241bec8485d08cce/src/plugins/fullscreen/store.ts#L27-L29

src/core/composables/useClusterMarker.ts
oeninghe-dataport: This is required for both `hovered` and `selected`. For `selected`, it is quite obvious, but it is also necessary for `hovered`, as the attached screencast proves at it is broken in POLAR@2.

Aufzeichnung 2026-08-27 093430 (1)

The zoom guard would have no effect here. The watcher already ensures that `oldValue !== newValue` is satisfied.

src/core/stores/index.ts
dopenguin: Wouldn't ClusterFeature encompass that more precisely?

src/core/stores/marker.ts
oeninghe-dataport: Good catch

src/core/stores/index.ts
oeninghe-dataport: To me, both are equally precise. If you prefer yours, I can change it.

dopenguin: Updated the suggestions

src/core/stores/marker.ts
dopenguin: The `fullscreen` computed receives stable references so recomputing them is easy. Here, by using `toMerged`, a new object is constructed on every recompute, those objects are used as cache keys in `memoizeStyle`, so any recompute of `layers` invalidates the identity-based style cache. Not the same pattern.

Additionally, nothing is done reactively if the configuration changes; `setup` isn't called again, so `setStyle` and the layer stamping stay the same while only `layers` gets recomputed. A runtime config change wouldn't propagate consistently right now.

Deriving once, as before, removes the cache-miss risk *by construction* with no loss of a capability we actually have. If runtime reconfiguration ever becomes a feature we support this would need a deliberate re-setup, not an incidental recompute, which we can add then.

src/core/utils/map/setupMarkers.ts
dopenguin: ☎️ -> We use `callOnMapSelect` again and opens the iconMenu on default

src/core/stores/marker.ts
dopenguin: ☎️ -> @oeninghe-dataport will do that

src/core/utils/map/setupMarkers.ts
oeninghe-dataport: fb73d73591

src/core/stores/marker.ts
oeninghe-dataport: The previous `isMaxZoom` was defined as:

```javascript
const isMaxZoom = map.getView().getZoom() !== map.getView().getMaxZoom()
```

The new `isMaxZoom` is defined as `mainStore.map.getView().getMaxZoom() === mainStore.zoom` which is the opposite (however, matches the name now; previously, this was "isNotMaxZoom").

Therefore, there was no effective change here.

src/core/stores/marker.ts
oeninghe-dataport: f16ced6d38

src/core/stores/marker.ts
oeninghe-dataport: e42248f3c7

src/core/stores/marker.ts
oeninghe-dataport: ddf374c4b2

src/core/composables/useClusterMarker.ts
oeninghe-dataport: It might occur that this is not `defaultStyle` (or `unselectableStyle`) if the feature was hovered and selected and looses only one of these states.

It is also not possible to return `undefined` in the style function, as this is also used as the layer style function.

src/core/stores/marker.ts
oeninghe-dataport: However, actually, the re-centering is lost, so I re-introduced it: 5cab530e93f8d442de3ac4942b23262498bebae7

oeninghe-dataport:
🏓 @dopenguin

src/core/types/marker.ts
```diff
@@ -128,17 +128,17 @@ export interface MarkerConfiguration {
...
+	 * {
+	 * 	action: 'openMenuById',
+	 * 	payload: 'gfi',
+	 * 	pluginName: 'iconMenu'
 	 * }
 	 * ```
 	 *
 	 * @remarks
-	 * The example open the gfi window in the iconMenu, if the IconMenu exists
+	 * The default value open the gfi window in the iconMenu, if the iconMenu exists
```
dopenguin:
```suggestion
	 * The default value opens the gfi window in the iconMenu, if the iconMenu exists
```

src/core/stores/marker.ts
```diff
@@ -1,41 +1,326 @@
...
+			const { action, payload, pluginName } = callOnMapSelect.value
+			if (!pluginName) {
+				mainStore[action](payload)
+				return
+			}
+
+			const pluginListStore = usePluginStore()
+			const pluginStore = pluginListStore.getPluginStore(pluginName as PluginId)
+			if (!pluginStore) {
+				// As the default assumes GFI is available, we don't want to throw an error if the plugin is not installed.
```
dopenguin:
```suggestion
				// As the default assumes IconMenu is available, we don't want to throw an error if the plugin is not installed.
```

dopenguin (changes requested):
smol

🏓 @oeninghe-dataport

oeninghe-dataport:
🏓 @dopenguin

src/core/stores/marker.ts
oeninghe-dataport: ...and I refactored it to be watcher-less: 1718b6f5ee

dopenguin (approved):
:shipit: @oeninghe-dataport

PR #948, opened 2026-08-25
refactor: introduce shared lib findLayer that finds a layer by its ID

oeninghe-dataport:
As discussed at https://github.com/Dataport/polar/pull/939/changes#r3845697917


dopenguin (approved):
🚀 @oeninghe-dataport

PR #969, opened 2026-09-11
Fix masterportalapi defaults

### LINKED ISSUE #959: Invisible default design for point geometries
### Describe the bug

[missingDesign.patch](https://github.com/user-attachments/files/31832392/missingDesign.patch)

When removing the `markers` configuration, point geometries from WFS are invisible by default.

### Expected behavior

This differs from the behaviour of POLAR@2, where they'd be displayed in the masterportalApi default marker design. (Fancy blue pin.)

While I don't expect it to look the same, *something* should be visible.

### Steps to reproduce

1. Apply patch
2. Run snowbox
3. Make reports visible
4. They are not visible

### Screenshots

Image

### Desktop

Doesn't matter.

### Smartphone

Doesn't matter.

### Additional context

Might be a configuration slip-up or changed default behaviour. We'll discuss this later.

warm-coolguy:
## Summary

Somehow the default marker design from the masterportalAPI doesn't work at the moment. However, our docs say that we use the OL default anyway:

https://github.com/Dataport/polar/blob/22fd5540ffac9cf25b4a4506e6f8b47ab8641071/src/core/types/layer.ts#L121-L127

Hence, the OL default is now used, and markers will only appear if we configure them to.

Remove `markers` entry in snowbox. Patch is in issue if you prefer that route. Run snowbox, receive styles.

- [x] Functionality has been tested in Chrome

Resolves https://github.com/Dataport/polar/issues/959

### DIFF
--- a/src/core/utils/map/olDefaultStyle.ts
+++ b/src/core/utils/map/olDefaultStyle.ts
@@ -1,0 +1,25 @@
+import { Circle, Fill, Stroke, Style } from 'ol/style.js'
+
+const fill = new Fill({
+	color: 'rgba(255,255,255,0.4)',
+})
+const stroke = new Stroke({
+	color: '#3399CC',
+	width: 1.25,
+})
+
+/**
+ * OL default style according to their style documentation:
+ * https://openlayers.org/en/latest/apidoc/module-ol_style_Style-Style.html
+ */
+export const styles = [
+	new Style({
+		image: new Circle({
+			fill,
+			stroke,
+			radius: 5,
+		}),
+		fill,
+		stroke,
+	}),
+]
--- a/src/core/utils/map/setupStyling.ts
+++ b/src/core/utils/map/setupStyling.ts
@@ -6,15 +6,35 @@ import type {
 	MasterportalApiServiceRegister,
 } from '../../types'
 
+import { setCustomStyles } from '@masterportal/masterportalapi/src/layer/geojson'
 import createStyle from '@masterportal/masterportalapi/src/vectorStyle/createStyle'
 import styleList from '@masterportal/masterportalapi/src/vectorStyle/styleList'
 import noop from '@repositoryname/noop'
+
+import { styles } from './olDefaultStyle'
+
+const changeMasterportalApiDefaults = () => {
+	setCustomStyles({
+		// required by masterportalAPI
+		/* eslint-disable @typescript-eslint/naming-convention */
+		Point: styles,
+		LineString: styles,
+		MultiLineString: styles,
+		MultiPoint: styles,
+		MultiPolygon: styles,
+		Polygon: styles,
+		GeometryCollection: styles,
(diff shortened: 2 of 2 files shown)


src/core/utils/map/setupStyling.ts
```diff
@@ -6,15 +6,35 @@ import type {
...
 
+import { setCustomStyles } from '@masterportal/masterportalapi/src/layer/geojson'
 import createStyle from '@masterportal/masterportalapi/src/vectorStyle/createStyle'
 import styleList from '@masterportal/masterportalapi/src/vectorStyle/styleList'
 import noop from '@repositoryname/noop'
 
+import { styles } from './olDefaultStyle'
+
+const changeMasterportalApiDefaults = () => {
+	setCustomStyles({
```
dopenguin: Is there a reason for this indirection? Why is this not directly called in `setupStyling`?

src/core/utils/map/setupStyling.ts
```diff
@@ -6,15 +6,35 @@ import type {
...
+		/* eslint-enable @typescript-eslint/naming-convention */
+	})
+}
+
 export async function setupStyling(
 	map: Map,
 	configuration: MapConfiguration,
 	register: MasterportalApiServiceRegister
 ) {
+	changeMasterportalApiDefaults()
```
dopenguin: When taking a look at the default Styles in `@masterportal/masterportalapi/src/layer/geojson/style.js`, this should not be needed. The default styling also works just fine for e.g. `ausgleichsflaechen`, even if I remove the `styleId`.

Please approach this differently.

src/core/utils/map/olDefaultStyle.ts
```diff
@@ -0,0 +1,25 @@
+import { Circle, Fill, Stroke, Style } from 'ol/style.js'
```
dopenguin:
```suggestion
import { Circle, Fill, Stroke, Style } from 'ol/style'
```

dopenguin (changes requested):
🏓 @warm-coolguy

src/core/utils/map/setupStyling.ts
warm-coolguy: Small function make Grug happy. Mind at rest.

src/core/utils/map/setupStyling.ts
warm-coolguy: Why use masterportalApiDefaults when olDefaults good? masterportalApiDefaults no work for dots.

warm-coolguy:
> 🏓 @warm-coolguy

🗿 @dopenguin

src/core/utils/map/setupStyling.ts
dopenguin: I GUESS

dopenguin (approved):
:shipit: @warm-coolguy

